PR #32, opened 2021-07-11
Feature: Data part

Imipenem:
**PR Checklist**

**Description of changes**

Implementation of the data part of the ehrapy framework:

- [x] Implement autodetection for categorical columns
- [x] Init AnnData object from read dataframe
- [x] Divide reading into AnnData and encoding into separate steps
- [x] Write a high level read method
- [x] Adress the issue of reading large files?
- [x] Write a generalized encode method
- [x] Adjust encoding function to allow for multiple encodings
- [x] Fix caching via writing to an hd5a file
- [x] Reading and writing of h5ad files -> encoded like scanpy -> non-encoded by saving and restoring full string X numpy arrays
- [x] Write MIMIC 2 dataloader
- [x] generalize `vars_tree` in `__init__.py` of `util`
- [x] Add progressbars whenever applicable
- [x] Add parameter for patient_id column name
- [x] Add option when reading to NOT load specific columns into X and to only copy those into obs
- [x] Improve importing of API -> don't want to users to work with static classes
- [x] Add functions to docs 
- [x] Solve all TODOs or create issues for the ones we cannot yet solve
- [x] Add parameter to mimic 2 dataloader to return encoded version
- [x] Add function to UNDO current encoding
- [x] fix typo -> did not find
- [x] Add notebook which shows all current ehrapy features for LH to test it!
- [x] Add parameter to ann_data_tree to sort it
- [x] Rename ann_data_tree method to something better
- [x] Add all other encodings that have not been implemented yet
- [x] solve docs warnings: 
```
/home/zeth/PycharmProjects/ehrapy/docs/usage.rst:39: WARNING: duplicate object description of ehrapy.api, other instance in usage, use :noindex: for one of them
/home/zeth/PycharmProjects/ehrapy/docs/usage.rst:50: WARNING: duplicate object description of ehrapy.api, other instance in usage, use :noindex: for one of them
/home/zeth/PycharmProjects/ehrapy/docs/usage.rst:79: WARNING: duplicate object description of ehrapy.api, other instance in usage, use :noindex: for one of them
```
- [x] Determine whether it is possible to reduce the doc string redundancy of the implementation & what we expose for the API docs in the __init__.py files. Maybe we can forward the docstring?
- [x] Try to remove as many test warnings as possible (when running nox we get like 5 warnings via typeguard & tests, they are not helpful)

**Technical details**

> Implement autodetection for categorical columns

A column will be counted as categorical if any of the following conditions is met:

1. The column has string values, but it's not free text (at least two duplicate elements)
2. The column has numerical values, but at most (20%, TBD) are duplicate values
3. The column is boolean and or has only one or two unique values (unary/binary levels)


ehrapy/api/core/__init__.py
```diff
@@ -0,0 +1,5 @@
+def view_obs(ann_data):
```
Zethson, line 1: Don't think that we need this function anymore, do we?

ehrapy/api/core/_prep_data.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Args:
+        df:
+           The original dataframe
+
+    Returns:
+        A dictionary containing all categorical column names with a hint on whether they need to be encoded or not
+
+    """
+    categoricals: Mapping[str, List[Optional[str]]] = {"categorical_encode": [], "categorical_no_encode": []}
```
Zethson, line 20:
```suggestion
    categoricals: Mapping[str, List[Optional[str]]] = {"categorical_encode": [], "categorical_not_encoded": []}
```

ehrapy/api/core/_prep_data.py
```diff
@@ -0,0 +1,73 @@
...
+        # when we only have unary/binary numerical categories
+        if 1 <= len(c.categories) <= 2 or (
+            (c_dtype == "floating" or c_dtype == "integer") and 1 <= len(c.categories) <= 2
+        ):
+            if c_dtype != "floating" and c_dtype != "integer":
+                return True, "categorical_encode"
+            else:
+                return True, "categorical_no_encode"
+    # TODO: Which type of exception?
+    except Exception:
```
Zethson, line 56: A custom exception? Could create a new one.

ehrapy/api/core/_prep_data.py
```diff
@@ -0,0 +1,73 @@
...
+        print(
+            f"[bold red] Could not cast column {col.name} to Categorical type. Please file an issue"
```
Zethson, line 58: Think in this case it is fine that we point the users to file issues, but we should not always force that.

We should offer a resolution such as "ensure that col.name is of type str" or so?

ehrapy/util/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+from rich.tree import Tree
+
+
+def view_vars(ann_data):
+    tree = Tree(
+        f"[bold red]Variable names for AnnData object with {len(ann_data.raw.var_names)} variables",
+        guide_style="underline2 bright_blue",
+    )
+    branch = tree.add(Text("📄 " + "Categoricals"), style="bold green")
+    branch.add("Day_ICU_intime with 7 different categories", style="blue")
```
Zethson, line 12: I like the method. We should generalize it.

noxfile.py
```diff
@@ -15,7 +15,7 @@
     sys.exit(1)
 
 package = "ehrapy"
-python_versions = ["3.8"]
+python_versions = ["3.9"]
```
Zethson, line 18: Think 3.8 is the standard so far. We should stick to it and you might have to create a 3.8 env for it

Zethson (commented):
Some random thoughts that I just wanted to write while scrolling.

ehrapy/api/io/read.py
```diff
@@ -307,6 +332,46 @@ def _check_datafile_present_and_download(path: Union[str, Path], backup_url=None
...
+        """
+        Move the given columns from the original dataframe (and therefore X) to obs. By doing so, those values will not get lost
+        and will be stored in obs, but will not appear in X. This may be useful for textual values like free text.
+        """
+        if columns_obs_only:
+            try:
+                obs = df[columns_obs_only].copy()
+                obs = obs.set_index(df.index.map(str))
+                df = df.drop(columns_obs_only, axis=1)
+            # TODO Key error traceback still prints, have no idea why
```
Imipenem, line 364: @Zethson could need your help/input here: Despite "catching" the `KeyError` if a misspelled column is passed and raising the custom `ColumnNotFoundError`, the traceback of the `KeyError` is still printed and therefore still raised somehow. I can't see why this is the case. Maybe I'm missing something obvious here.

Zethson: https://paste.ofcode.org/P28YFe8L75vxVMryXwLanR

ehrapy/api/data/datasets.py
```diff
@@ -1,23 +1,27 @@
...
+class Datasets:
+    @staticmethod
+    def mimic_2(encode: bool = False) -> AnnData:
+        adata = DataReader.read(
+            filename="ehrapy_mimic2.csv",
+            backup_url="https://www.physionet.org/files/mimic2-iaccd/1.0/full_cohort_data.csv?download",
+            suppress_warnings=True,
+        )
+        if encode:
+            return Encoder.encode(adata, autodetect=True)
```
Zethson, line 16: Autodetect would now just label encode everything, right? That's not how it should be.

There are some columns that should def be one-hot encoded.

ehrapy/api/encode/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+    Args:
+        adata: The AnnData object
+        columns: The names of the columns to reset encoding for. Defaults to all columns.
+        from_cache_file: Whether to reset all encodings by reading from a cached .h5ad file, if available. This resets the AnnData object to its initial
+        state.
+        TODO replace this once settings.cache_dir is available
+        cache_file: The filename of the cache file to read from
+
+    Returns:
+        A (partially) encoding reset AnnData object
```
Zethson, line 55: Can we get an example call here, please?

ehrapy/api/encode/_categoricals.py
```diff
@@ -0,0 +1,89 @@
...
+    Returns:
+        A dictionary containing all categorical column names with a hint on whether they need to be encoded or not
+    """
+    categoricals: Mapping[str, List[Optional[str]]] = {
+        "categorical_encoded": [],
+        "categorical_not_encoded": [],
+        "not_categorical": [],
+    }
+    for i in range(data.shape[1]):
+        categorical_column = _is_categorical_column(data[::, i : i + 1 :].ravel(), col_names[i])
```
Zethson, line 33: This line is complex and might need an explanation?

ehrapy/api/encode/encode.py
```diff
@@ -0,0 +1,464 @@
...
+                # update current encodings in uns
+                encoded_ann_data.uns["current_encodings"] = current_encodings
+
+            # if the user did not pass every non numerical column for encoding, a Anndata object cannot be created
+            except ValueError:
+                print(
+                    "[bold red]Creation of AnnData object failed. "
+                    "Ensure that you passed all non numerical, categorical values for encoding!"
+                )
+                sys.exit(1)
```
Zethson, line 148: This might be bad when ehrapy is used as a library in notebooks as it is designed to, right?
Let's not use sys.exits because this is not just a CLI app

ehrapy/api/encode/encode.py
```diff
@@ -0,0 +1,464 @@
...
+        )
+
+        for idx, var_name in enumerate(ann_data.var_names):
+            if is_init and var_name in ann_data.uns["original_values_categoricals"]:
+                continue
+            elif var_name in categorical_names:
+                ann_data.uns["original_values_categoricals"][var_name] = ann_data.X[::, idx : idx + 1]
+
+
+class AlreadyEncodedWarning(UserWarning):
```
Zethson, line 463: Should this be:
EhrapyAlreadyEncodedWarning
?
I honestly don't know

ehrapy/api/io/read.py
```diff
@@ -1,104 +1,436 @@
...
```
Zethson, line 381: We solved this. Remove this line

ehrapy/api/io/read.py
```diff
@@ -1,104 +1,436 @@
...
+        Returns:
+            True if all elements are of the same type, False otherwise.
+        """
+        iseq = iter(sequence)
+        first_type = type(next(iseq))
+
+        return first_type if all((type(x) is first_type) for x in iseq) else False
+
+
+class IndexColumnWarning(UserWarning):
```
Zethson, line 431: Naming see above?

ehrapy/api/io/read.py
```diff
@@ -1,104 +1,436 @@
...
+    pass
+
 
+class ColumnNotFoundError(Exception):
```
Zethson, line 435: Naming see above?

ehrapy/api/io/utility_io.py
```diff
@@ -0,0 +1,90 @@
+from pathlib import Path, PurePath
+from typing import Union
+
+supported_extensions = {"csv", "tsv", "tab", "txt", "h5ad"}
+
+
+def is_valid_filename(file_path: Path, return_ext=False) -> Union[str, bool]:
```
Zethson, line 7: underscore method?

Zethson (changes requested):
Just a few minor things and we can merge this.
I will certainly find more after we're back in full force, but this 10 min review should be good enough now :)

ehrapy/api/data/datasets.py
Imipenem: Mh, I changed this because of the encoding for the caching file. 
I mean, one could just pass an _encoding_mode_ parameter, but it's not possible to apply an encoding mode to specific columns as it is the case without autodetect.

Zethson:
@Imipenem one further thing that I noticed is that the encoding API looks ugly AF:

`adata_2 = ep.encode.encode(adata_2, encodings={"one_hot_encoding": ["service_unit", "day_icu_intime"]})`

`encode.encode` is not nice

A couple of options:

1. ec.encode()
2. pp.encode() -> part of preprocessing
3. ds.encode() or dt.encode() -> part of data

PR #98, opened 2021-09-21
add access to scanpy settings

Zethson:
Utilizes the Scanpy settings for saving plots, figure settings etc


Imipenem (changes requested):
- Some tests should be added here as well

- IMO, there is also a "settings TODO" in the caching part?

Otherwise LGTM

Issue #45, opened 2021-07-20
Data imputation


Zethson:
Contrary to count data we can actually infer statistics from the data and impute it. Very important here again to be able to selectively apply these imputations to specific columns.

- [ ] Replacing with an explicit value
- [ ] KNN imputation
- [ ] Mean imputation
- [ ] z-score imputation
- [ ] fully conditional specification multiple imputation https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4945131/
- [ ] multiple imputation with chained equations using classification and regression trees (MICE)  https://towardsdatascience.com/multiple-imputation-with-random-forests-in-python-dec83c0ac55b

Imipenem:
https://scikit-learn.org/stable/auto_examples/impute/plot_missing_values.html#sphx-glr-auto-examples-impute-plot-missing-values-py

Might be a good read here.

Imipenem:
@Zethson; I can find a lot of published papers and articles online that state using mean imputation is a very bad way to impute data and should be avoided (especially because relationship between variables in a dataset might be lost). Since we have some good alternatives: What do you think of not exposing this one via our ehrapy API?

Zethson:
> @Zethson; I can find a lot of published papers and articles online that state using mean imputation is a very bad way to impute data and should be avoided (especially because relationship between variables in a dataset might be lost). Since we have some good alternatives: What do you think of not exposing this one via our ehrapy API?

I'd offer it anyways because it is trivial to implement. I guess by the time that we know which methods actually suit our data we can still trivially disable it.

Imipenem:
Fair enough. I will anyways restrict `mean` and `median` imputation to numerical data, since applying those to text data for example will just render a meaningless value that can not be mapped to any meaningful value.

PR #104, opened 2021-09-24
API restructuring, demo updated & replace_explicit draft

Zethson:
Part of #45 

1. This PR includes API shortcuts like pp for processing, ec for encode etc
2. I've updated our demo notebook to accommodate the changes in this PR
3. This PR adds a draft for explicit replacement of missing values

The replace_explicit function is not yet done nor has it tests!
However, I would like to include them in a follow up PR as soon as I find the time.

Due to the restructuring I would like to merge this one beforehand.


ehrapy/api/__init__.py
```diff
@@ -0,0 +1,5 @@
+from ehrapy.api import data as dt
```
Zethson, line 1: Does not include IO here because we are not renaming it. Should be good.

ehrapy/api/preprocessing/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Returns:
+        :class:`~anndata.AnnData` object with imputed X
+
+    Example:
+        .. code-block:: python
+
+            import ehrapy.api as ep
+            adata = ep.data.mimic_2(encode=True)
+            adata_replaced = ep.pp.replace_explicit(adata_3, replacement=0, copy=True)
```
Imipenem, line 31: does this work? what is adata_3 here?

Zethson: a copy pasta mistake from my side

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -0,0 +1,45 @@
...
+        Returns:
+            :class:`~anndata.AnnData` object with imputed X
+        """
+        adata_to_act_on = adata
+        if copy:
+            adata_copy = adata.copy()
+            adata_to_act_on = adata_copy
+
+        # scenario 1: Replace all missing values with the specified value
+        adata_to_act_on.X[np.isnan(adata_to_act_on.X)] = replacement
```
Imipenem, line 38: does `isnan()` also work on empty string for example?

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -0,0 +1,45 @@
+from typing import Dict, Optional, Tuple, Union
+
+import numpy as np
+from anndata import AnnData
+
+
+class Imputation:
```
Imipenem, line 7: In general; It would be better to no not have this in this PR, since it's not fully implemented and I really did not expected it here, since this would be something that better fits into quality control.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ scikit-learn = "^1.0"
 category_encoders = "^2.2.2"
 scanpy = "^1.8.1"
 mudata = "^0.1.0"
+leidenalg = "^0.8.7"
```
Imipenem, line 36: why is this added?

ehrapy/api/preprocessing/_data_imputation.py
Zethson: I needed it for the notebook update. But I can add it in another PR that you can then use as your base.

pyproject.toml
Zethson: Leidenalg is an optional dependency in Scanpy. However, we need it for clustering as is now included in our example notebook. It should imo always be installed.

ehrapy/api/preprocessing/_data_imputation.py
Imipenem: IMO, for now its fine, just wanted to state it.

Imipenem:
Before merging this, tests should be added here

Issue #94, opened 2021-09-17
Translations


Zethson:
Add Deepl integration to translate columns from language A to language B. Usually X to English.
https://github.com/DeepLcom/deepl-python

It should
1. Handle authentication
2. Allow for the replacement of an obs column or whether it should just be added

Zethson:
https://github.com/nidhaloff/deep-translator

exists. But does not support all features of e.g. DeepL (yet). I shall think how we should proceed.

Zethson:
Might be good if I simply rewrap deep-translator as stupid as it sounds because the API is also not consistent....

PR #108, opened 2021-09-27
add DeepL translation

Zethson:
#94 

This PR adds

1. Integration with DeepL
2. Tests for DeepL which can only be run with a token

We still need methods to easily update AnnData objects with these translations. Might be this PR, might also get a follow up PR.

### DIFF
--- a/ehrapy/api/tools/nlp/_translators.py
+++ b/ehrapy/api/tools/nlp/_translators.py
@@ -147,7 +147,7 @@ class DeepL:
     def translate_obs_column(
         self,
         adata: AnnData,
-        target_language: str,
+        target_language: str = "EN-US",
         columns=Union[str, List],
         translate_column_name: bool = False,
         inplace: bool = False,
@@ -156,7 +156,7 @@ class DeepL:
 
         Args:
             adata: :class:`~anndata.AnnData` object containing the obs column to translate
-            target_language: The target language to translate into, e.g. EN-US
+            target_language: The target language to translate into (default: EN-US)
             columns: The columns to translate. Can be either a single column (str) or a list of columns
             translate_column_name: Whether to translate the column name itself
             inplace: Whether to replace the obs values or add a new obs column
@@ -172,7 +172,7 @@ class DeepL:
             if translate_column_name:
                 target_column = self.translator.translate_text(column, target_lang=target_language).text
             if not inplace:
-                f"{target_column}_{target_language}"
+                target_column = f"{target_column}_{target_language}"
 
             adata.obs[target_column] = adata.obs[column].apply(
                 lambda text: self.translator.translate_text(text, target_lang=target_language).text
@@ -181,7 +181,7 @@ class DeepL:
     def translate_var_column(
         self,
         adata: AnnData,
-        target_language: str,
+        target_language: str = "EN-US",
         columns=Union[str, List],
         translate_column_name: bool = False,
         inplace: bool = False,
@@ -190,7 +190,7 @@ class DeepL:
 
         Args:
             adata: :class:`~anndata.AnnData` object containing the obs column to translate
-            target_language: The target language to translate into, e.g. EN-US
+            target_language: The target language to translate into (default: EN-US)
             columns: The columns to translate. Can be either a single column (str) or a list of columns
             translate_column_name: Whether to translate the column name itself
             inplace: Whether to replace the obs values or add a new obs column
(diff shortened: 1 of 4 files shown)


Zethson:
I should ensure that we have 3 ways to translate

1. translate obs (done)
2. translate var (missing!)
3. translate X (WIP)

ehrapy/api/tools/nlp/_translators.py
```diff
@@ -0,0 +1,252 @@
...
+        Example: "DE (German)"
+        """
+        for language in self.translator.get_source_languages():
+            print(f"{language.code} ({language.name})")
+
+    def print_target_languages(self) -> None:  # pragma: no cover
+        """Prints all possible target languages to translate to"""
+        for language in self.translator.get_target_languages():
+            if language.supports_formality:
+                print(f"{language.code} ({language.name}) supports formality")
```
Imipenem, line 80: What does "support formality" mean?

ehrapy/api/tools/nlp/_translators.py
```diff
@@ -0,0 +1,252 @@
...
+            target_language: The target language to translate the Text into, e.g. EN-GB
+
+        Returns:
+            A :class:`~deepl.TextResult` object
+        """
+        return self.translator.translate_text(text, target_lang=target_language)
+
+    @_check_usage  # type: ignore # pragma: no cover
+    def translate_document(
+        self, input_file_path: str, output_path: str, target_language: str, formality: str = Formality.DEFAULT
```
Imipenem, line 99: General question concerning "target_language": Shouldn't this (at least default to) always be English? Since (at least for now I'd say) our standard language is English.

ehrapy/api/tools/nlp/_translators.py
```diff
@@ -0,0 +1,252 @@
...
+
+        for column in columns:
+            # as of Pandas 1.1.0 the default for new string column is still 'object'
+            if adata.var[column].dtype != str and adata.var[column].dtype != object:
+                raise ValueError("Attempted to translate column {column} which does not contain only strings.")
+            target_column = column
+            if translate_column_name:
+                target_column = self.translator.translate_text(column, target_lang=target_language).text
+            if not inplace:
+                f"{target_column}_{target_language}"
```
Imipenem, line 209: see `obs` comment on same line

ehrapy/api/tools/nlp/_translators.py
```diff
@@ -0,0 +1,252 @@
...
+            target_language: The target language to translate into, e.g. EN-US
+            columns: The columns to translate. Can be either a single column (str) or a list of columns
+            translate_column_name: Whether to translate the column name itself (only translates var_names, not var)
+        """
+        if isinstance(columns, str):
+            columns = [columns]
+
+        for column in columns:
+            index: int = get_column_indices(adata, column)[0]
+            column_values = get_column_values(adata, index)
```
Imipenem, line 237: In `get_column_indices` and `get_column_values` you already loop over (possibly) all var names. It would be more performant to actually move (at least the index function) outside of this loop and passing the `columns` parameter as list (I think it can handle it, what I remember). So this loop only runs once like O(n) and not (worst case, O(n^2)).

tests/_test_util.py
```diff
@@ -0,0 +1,118 @@
...
+    obsm_types
+        What kinds of containers should be in `.obsm`?
+    varm_types
+        What kinds of containers should be in `.varm`?
+    layers_types
+        What kinds of containers should be in `.layers`?
+    """
+    M, N = shape
+    obs_names = pd.Index(f"cell{i}" for i in range(shape[0]))
+    var_names = pd.Index(f"gene{i}" for i in range(shape[1]))
```
Imipenem, line 43: maybe we should rename this one and `cell` above ;)

tests/_test_util.py
```diff
@@ -0,0 +1,118 @@
+from string import ascii_letters
+from typing import Collection, Tuple
+
+import numpy as np
+import pandas as pd
+from anndata import AnnData
+from scipy import sparse
+
+
+def gen_adata(
```
Imipenem, line 10: In general: I like this function a lot, but I think we have to alter it a bit (does not have to be this PR), since it does not displays our data well (like multiple events of one patient for example).

Zethson: Perfectly agree. It was just copied from AnnData :)

ehrapy/api/tools/nlp/_translators.py
```diff
@@ -0,0 +1,252 @@
...
+            if adata.obs[column].dtype != str and adata.obs[column].dtype != object:
+                raise ValueError("Attempted to translate column {column} which does not contain only strings.")
+            target_column = column
+            if translate_column_name:
+                target_column = self.translator.translate_text(column, target_lang=target_language).text
+            if not inplace:
+                f"{target_column}_{target_language}"
+
+            adata.obs[target_column] = adata.obs[column].apply(
+                lambda text: self.translator.translate_text(text, target_lang=target_language).text
```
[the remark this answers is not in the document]
Zethson: >               raise ValueError("text must not be empty")
E               ValueError: text must not be empty

PR #147, opened 2021-11-03
add draft notebooks CI wf

### LINKED ISSUE #138: Add a CI job to run all notebooks.
Ideally they should run in parallel. We need to ensure that they do not take too long.

Zethson:

Fixes #138


Zethson:
I can't find a reasonable way to make it work. The action expects params, but we don't want to parameterize our notebooks.

Zethson:
Maybe we have to write custom code to convert it and then run it

https://stackoverflow.com/questions/60078735/start-notebook-from-other-notebook/60081404#60081404

Issue #52, opened 2021-07-21
Normalization


Zethson:
Our **non-categorical**  and **non-NLP** data needs to be normalized to facilitate downstream analysis (not DE) and especially visualization. The input fluids plot is a great example of when a single feature clearly dominates the variance in the UMAP.

![image](https://user-images.githubusercontent.com/21954664/141275997-c7cc8888-07d8-4af8-957f-0bdd4de89a72.png)

We want to implement at least the following normalization strategies:

- [ ] min max scaling
- [ ] mean normalization
- [ ] z-score normalization
- [ ] constant factor

We should look at https://scikit-learn.org/stable/modules/preprocessing.html to get ideas for more normalization strategies.

When implementing the normalizing algorithms it is absolutely vital to ensure that we only transform the real numerical values and not any encoded values (e.g. one-hot encoded). This requires selecting the appropriate columns in X and then applying the normalization selectively. The in place updates need to ensure that they do not overwrite any encoded columns.

Proposed milestones:

- [x] Look for normalization strategies and algorithms. Any new ones that we can make use of?
- [x] Look at how scanpy implemented its normalization
- [ ] Remove the normalization that we are currently wrapping from the Scanpy API. It does not support the selection of columns that we require.
- [x] Add a new class for normalization in the preprocessing module in its own file & write stub functions
- [x] Add a new corresponding test class & write stub test functions
- [x] Load the MIMIC-II dataset with encoded columns
- [x] Try to apply a transformation to X to only the numerical values ensuring that categoricals are untouched
- [ ] Implement all normalization strategies. 
- [ ] Write tests
- [ ] Submit PR
- [ ] Celebrate

PR #162, opened 2021-11-29
Add normalization

lazappi:
**PR Checklist**

**Description of changes**

This PR will implement normalisation for numeric variables as suggested in #52 

**Technical details**

Things that have been implemented so far:

- [x] Skeleton for the `normalization` submodule
- [x] Skeleton for testing the `normalization` submodule
- [x] Skeletons for getter and setter functions for numeric variables  
- [x] Functional numeric getter and setter functions


lazappi:
@Zethson It would great if you could take a quick look at the skeletons here and let me know if there is anything I should add/change before I start on making things actually functional

ehrapy/api/_anndata_util.py
```diff
@@ -36,3 +37,41 @@ def get_column_values(adata: AnnData, indices: Union[int, List[int]]) -> np.ndar
         :class:`~numpy.ndarray` object containing the column values
     """
     return np.take(adata.X, indices, axis=1)
+
+
+def get_numeric_vars(adata: AnnData) -> List[str]:
+    """Fetches the column names for numeric variables in X
```
Zethson, line 43:
```suggestion
    """Fetches the column names for numeric variables in X.
```

It is convention to always end the header sentence with a `.` :) I disabled the linting checks for this because I don't want it to be pedantic, but just mentioning it once.

ehrapy/api/_anndata_util.py
```diff
@@ -36,3 +37,41 @@ def get_column_values(adata: AnnData, indices: Union[int, List[int]]) -> np.ndar
...
+
```
Zethson, line 44: We could add a description here saying that this method filters data by removing all categorical values.

ehrapy/api/_anndata_util.py
```diff
@@ -36,3 +37,41 @@ def get_column_values(adata: AnnData, indices: Union[int, List[int]]) -> np.ndar
...
+    numeric_vars: List[str] = []
+
+    return numeric_vars
+
+
+def set_numeric_vars(
+    adata: AnnData, values: Union[np.ndarray, spmatrix], vars: Optional[List[str]] = None, copy: bool = False
+) -> Optional[AnnData]:
+    """Fetches the column names for numeric variables in X
```
Zethson, line 60: Copy paste mistake. This description needs to be amended :)

ehrapy/api/_anndata_util.py
```diff
@@ -36,3 +37,41 @@ def get_column_values(adata: AnnData, indices: Union[int, List[int]]) -> np.ndar
...
+        adata: :class:`~anndata.AnnData` object
+        values: Matrix containing the replacement values
+        vars: List of names of the numeric variables to replace. If `None` they will be detected using ~ehrapy.pp.get_numeric_vars.
+        copy: Whether to return a copy with the normalized data.
+
+    Returns:
+        :class:`~anndata.AnnData` object with updated X
+    """
+
+    adata_to_set = adata
```
Zethson, line 72: This was my ugly way of handling the copy. Think I already simplified this somewhere

```
if copy:
    adata = adata.copy()
```

should suffice I guess.

tests/api/preprocessing/test_normalization.py
```diff
@@ -0,0 +1,23 @@
...
+from anndata import AnnData
+
+from ehrapy.api.preprocessing import Normalization
+
+CURRENT_DIR = Path(__file__).parent
+_TEST_PATH = f"{CURRENT_DIR}/test_preprocessing"
+
+
+class TestNormalization:
+    def test_identity(self):
```
Zethson, line 13: I would suggest to add a setup method which creates a couple of AnnData objects to use for testing purposes.
We should test 1 AnnData object which only has numericals and one which has a mix of numericals and categoricals which got encoded earlier. You could also think about a case where X still contains strings which did not get encoded yet. These should not be used for the normalization. Also, the information in `uns` about which columns are categoricals may not yet exist. You could attempt to figure out which columns are Strings and therefore categoricals and proceed as usual or throw an exception telling people to encode their data first. Happy with either solution.

See: https://github.com/theislab/ehrapy/blob/development/tests/api/tools/nlp/test_translator.py#L23

Zethson (commented):
Looks good! I have the impression that you understood the concepts well and we'll have no trouble getting your eventual implementation into ehrapy in a quick manner.

As soon as the methods have been implemented we can take a look at how to get your methods into the documentation as I've shown you already.

So I guess the next step would be that you implement the `set_numeric_vars` and `get_numeric_vars` helper methods and a single normalization method and I take a look at it again?

Then finally the other normalization methods afterwards.

lazappi:
Thanks for the comments, they will be super helpful! I will update this when I have implemented some things.

lazappi:
I have tidied up some of the things you mentioned and filled in the getter/setter functions. You might want to check how I have handled raising errors because I wasn't really sure the best way to do that. I also wasn't sure where to put tests for these? I had a look around but I couldn't find an obvious place.

If this looks good I will move onto the first normalisation function which should be pretty quick.

ehrapy/api/_anndata_util.py
```diff
@@ -36,3 +36,66 @@ def get_column_values(adata: AnnData, indices: Union[int, List[int]]) -> np.ndar
...
+        copy: Whether to return a copy with the normalized data.
+
+    Returns:
+        :class:`~anndata.AnnData` object with updated X
+    """
+
+    if vars is None:
+        vars = get_numeric_vars(adata)
+
+    if not np.issubdtype(values.dtype, np.number):
```
Zethson, line 81: Is it np.number or np.float32? The former would have been my intuition.

ehrapy/api/_anndata_util.py
```diff
@@ -36,3 +36,66 @@ def get_column_values(adata: AnnData, indices: Union[int, List[int]]) -> np.ndar
...
+
+    if copy:
+        adata = adata.copy()
+
+    vars_idx = get_column_indices(adata, vars)
+
+    for i in range(n_values):
+        adata.X[:, vars_idx[i]] = values[:, i]
+
+    return adata
```
Zethson, line 97: I am not sure whether you want to return the whole AnnData object here.

What I would have expected is something like a dictionary that looks like 

```
{ 
    numerical_feature_1_var_name: numpy column with values,
    numerical_feature_2_var_name: numpy column with values,
    ...
}
```

And then later you replace the respective columns in the actual AnnData object when normalizing. My idea might however be slow and space wasteful. Not sure what your plan is.

This is just one of many ideas to approach this. Feel completely free to follow your own ideas here!

Zethson (commented):
Looks great. Exceptions also good. Looking forward to your further implementations here.

ehrapy/api/_anndata_util.py
lazappi: StackOverflow told me to use `np.number` to check for any kind of numeric matrix but happy to change it.

Zethson: @Imipenem got an idea here? Is `np.number` just a superset of `np.float32`? Any issue for us?

ehrapy/api/_anndata_util.py
lazappi: I was going to calculate the normalised values and then use this function to set them. I'll probably implement it that way and then we can discuss if something else would be better.

Zethson: `I was going to calculate the normalised values and then use this function to set them. I'll probably implement it that way and then we can discuss if something else would be better.`

Just need to ensure that you still have all previous non-numerical values and that you don't change the order of the columns. But if it works out totally fine.

lazappi: Ah, this is something I haven't thought of before. Should the normalized values be added as new columns of `X` or should they replace the existing columns and store the old values somewhere (I was going to do this)?

Zethson: Maybe we can store the unnormalized data in a new layer -> layer "raw"?
Then you can either replace the columns one by one in the current X with the normalized values or you replace the complete X. In both cases we need to ensure that the order still matches.

`X` should always have the same dimension. Normalized and unnormalized values should never be in the same matrix.

ehrapy/api/_anndata_util.py
Imipenem: @Zethson, I do not think we will get into any trouble here. 
In fact, since line 78 enforces the AnnData object to be already encoded and our default AnnData type (for X of course) is _float32_, this should be fine. 
IIRC, we don't even have an option to change X to be anything else than _float32_, when encoding columns (might need to implement this to give user more control over their data).

ehrapy/api/_anndata_util.py
Imipenem: Can't we just use our existing layer "original"? It should contain all original numerical values (and those, that are encoded are already stored in _.uns_).

Zethson: Guess this could also work. Wasn't quite sure whether this is issue because of quality control etc. But QC should affect the layer as well right?

Imipenem: Per default yes, IMO. (But we can also allow passing a parameter like `copy_layer=True` or similar, allowing API users decide on whether changes made on `X` should be reflected in the layer or not)

Zethson: I think that it would be easiest for @lazappi if we he simply adds a `raw` layer as discussed and we, @Imipenem will later discuss how we can optimize the required memory of the layers.

But I'd [KISS ](https://en.wikipedia.org/wiki/KISS_principle) it for now.

ehrapy/api/_anndata_util.py
```diff
@@ -36,3 +36,73 @@ def get_column_values(adata: AnnData, indices: Union[int, List[int]]) -> np.ndar
...
+    Args:
+        adata: :class:`~anndata.AnnData` object
+
+    Returns:
+        Set of column numeric column names
+    """
+
+    assert_encoded(adata)
+
+    return adata.uns["categoricals"]["not_categorical"]
```
Imipenem, line 60: @Zethson , @lazappi  ; Didn't test it myself but I guess we may run into issues here, considering the following scenario:

Encoding happens via `autodetect`. When this is the case we have three "types" of categories: The `non_categoricals` (used here), the `categoricals` that needs to be encoded and a third group that is stored under `categoricals`, but does not need any encodings since it's already numerical. 

I don't remember why we made a difference in the very beginning between those two types of categoricals (do you @Zethson ?). Maybe its completely unnecessary and we can remove it? Because those are the numerical values, that won't be returned here.

Zethson: "a third group that is stored under categoricals, but does not need any encodings since it's already numerical."

Maybe this was for columns like e.g sex? sex in the MIMIC2-IAC dataset is 0 or 1. Categorical but numerical.

Imipenem: Yes, but where are we actually using this? Do we need this?

Zethson: No not really. Should not be an issue with coloring. Thing is, I had the impression that we had this well mapped out so there must be some use-case for this that I cannot remember.
I'll need a fresh brain for this... 
If we don't come up with better ideas we can remove this ofc.

But atm not an issue for @lazappi right? He can just make the assumption that it does not matter for now.

Imipenem: I mean, if those "numerical" categoricals should not be normalized in any way, then it's no issue.

Zethson:
Ah of course they should not. This was exactly the use case haha. Don't normalize sex. Now I remember. So all good Luke.  Thursday, 16 December 2021, 06:48PM +01:00 from Philipp Ehmele  ***@***.*** :

***@***.*** commented on this pull request.

ehrapy/api/_anndata_util.py
lazappi: Yeah I noticed that an integer column I had was put into the "categorical but not encoded" category. I meant to ask about that but just assumed you chose to have it that way for some reason.

Zethson: Aye. See my comment in the whole PR. Everything is fine :)

lazappi:
I have implemented a couple of methods now. I changed the API around a fair bit from what I had in the template so you might want to check that you are ok with how it works now (basically I created one function that lets the user specify how each numeric variable is normalised rather than having a separate for each method).

If that looks good it's super quick to add more methods.

ehrapy/api/_anndata_util.py
```diff
@@ -36,3 +36,73 @@ def get_column_values(adata: AnnData, indices: Union[int, List[int]]) -> np.ndar
         :class:`~numpy.ndarray` object containing the column values
     """
     return np.take(adata.X, indices, axis=1)
+
+
+def assert_encoded(adata: AnnData):
```
Zethson, line 41: This method will eventually overlap with https://github.com/theislab/ehrapy/issues/182

But let's keep it for now!

ehrapy/api/preprocessing/normalization.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+
+        assert_encoded(adata)
+
+        num_vars = get_numeric_vars(adata)
+        if isinstance(methods, str):
+            methods = dict.fromkeys(num_vars, methods)
+        else:
+            if not set(methods.keys()) <= set(num_vars):
+                raise ValueError("Some keys of methods are not numeric variables")
```
Zethson, line 47: Will eventually also be solved with https://github.com/theislab/ehrapy/issues/182
But again, let 's keep it here.

Zethson (commented):
Excellent work @lazappi ! This is pretty much what I had in mind. Also very happy to see the tests + exceptions to test the preconditions of the functions.

Great!

tests/api/preprocessing/test_normalization.py
```diff
@@ -0,0 +1,103 @@
...
+        adata_norm = Normalization._normalize(self.adata, methods="identity", copy=True)
+
+        assert np.allclose(adata_norm.X, self.adata.X, equal_nan=True)
+        assert np.allclose(adata_norm.layers["raw"], self.adata.X, equal_nan=True)
+
+    def test_norm_mixed(self):
+        """Test for normalization with mixed methods."""
+
+        adata_norm = Normalization._normalize(
+            self.adata, methods={"Numeric1": "minmax", "Numeric2": "scale"}, copy=True
```
Zethson, line 92: Oh, @lazappi it might be better to actually implement this the other way around like we did it for encoding.

```
adata_norm = Normalization._normalize(
        self.adata, methods={"minmax": "numeric1", "scale": "numeric2"}, copy=True
)
```

The reason for this is that you will likely for larger datasets repeat the methods unnecessarily often. Easier to group the columns for specific normalization methods.

Zethson:
@lazappi do I see this right that you implemented the normalization methods column specific for now? So if I normalize a matrix of e.g. 5 numerical columns each column will be normalized in isolation?

I am asking because one use-case is the normalization across columns. The MIMIC-II dataset suffers from extremely large values for the `input_fluids` which then dominate the UMAP. Normalization across columns might mitigate this.

I would suggest to implement both possibilities . What do you think?

Zethson:
I will fix up this branch. 

@lazappi we decided to remove most of the classes to reduce code duplication. Initially we thought that they would help us with the tests to employ mocking more easily, but it turned out to not be necessary.

lazappi:
Ok, I have added the other methods from **sklearn**. We now have:

- Standard scaling
- MinMax
- MaxAbs
- Robust scaling
- Quantile normalisation (uniform and normal)
- Power transformation (Yeo-Johnson and Box-Cox)

I think it would also be nice to have clipping and winsorization but there aren't handy functions for these so they will be a bit more work and require some additional arguments. Logging is probably also something that we need.

Few questions/things I have thought of:

- There is a note in the original issue about removing existing wrappers of Scanpy functions. I think I know which these are so happy to do that unless you want to handle it (or let me know exactly which ones).
- It would probably be good to let users pass arguments onto the functions we are wrapping but not sure the best way to handle that, it's a bit tricky when multiple methods might be used at the same time.
- There isn't really any logging of what normalisation has been done at the moment. Do we want to record that somehow (store something in `obs`)?

From my point of view I think there is a decent amount of functionality now, just depends how many of these extra things you want to have before you merge this or are happy to add later.

Zethson:
Great! Thank you.

> Few questions/things I have thought of:
> 
>     * There is a note in the original issue about removing existing wrappers of Scanpy functions. I think I know which these are so happy to do that unless you want to handle it (or let me know exactly which ones).

Think it's only this one? https://github.com/theislab/ehrapy/blob/13f5e699cb00ef81fa00619942f942d7a2b8218f/ehrapy/api/preprocessing/_scanpy_pp_api.py#L116



>     * It would probably be good to let users pass arguments onto the functions we are wrapping but not sure the best way to handle that, it's a bit tricky when multiple methods might be used at the same time.

I think that you are looking for `**kwargs`. https://book.pythontips.com/en/latest/args_and_kwargs.html You can pass them to every normalization function.

>     * There isn't really any logging of what normalisation has been done at the moment. Do we want to record that somehow (store something in `obs`)?

Mhm. Think it's not stored in Scanpy as well, right? But Scanpy's case is simpler because you usually normalize everything the same way. I think a quick solution would be to just add a dictionary in `.uns` which records all applied normalizations.

> I think it would also be nice to have clipping and winsorization but there aren't handy functions for these so they will be a bit more work and require some additional arguments. 

If you have the time it would of course be cool if you could implement them. If not feel free to open an issue and we can merge this PR without these. You could tackle them later if you want or not (also totally fine).

> From my point of view I think there is a decent amount of functionality now, just depends how many of these extra things you want to have before you merge this or are happy to add later.

Not too much more :) The things you brought up and I think that we're good to go?
Happy to do another reviews after you think it's ready, but I doubt that I will have to ask for more stuff.

Zethson:
Oh, one more thing that would be cool:

There's https://github.com/theislab/ehrapy-tutorials which is used as a submodule for the documentation. Basically, in the end all notebooks will be rendered here: https://ehrapy.readthedocs.io/en/development/tutorials/index.html -> if you click on the MIMIC-II notebook you get the version that is in the ehrapy-tutorials repository. Can explain more if you want to.

Steps would be (order can be switched):
1. We merge this PR
2. You add a normalization method for things that you deem appropriate (I already point something out in the MIMIC-II notebook) and create a PR against the main branch of the `ehrapy-tutorials` repository.
3. We merge the PR.
4. We pull the submodule on the development branch of ehrapy.

This is simple and would complete the whole story of your contribution :)
I can also do it, but if you want to learn what submodules are this could be interesting.

lazappi:
> Think it's only this one?
> 
> https://github.com/theislab/ehrapy/blob/13f5e699cb00ef81fa00619942f942d7a2b8218f/ehrapy/api/preprocessing/_scanpy_pp_api.py#L116

There's also `log1p()` and `scale()` but maybe there are other uses for those?

> I think that you are looking for `**kwargs`. [book.pythontips.com/en/latest/args_and_kwargs.html](https://book.pythontips.com/en/latest/args_and_kwargs.html) You can pass them to every normalization function.

How would you handle a situation where you are normalising with multiple wrapped methods and some of those methods have parameters with the same name and you want to set them to different values? I couldn't find an example at the moment but could definitely be possible at some stage if more methods were wrapped.

Zethson:
> > Think it's only this one?
> > https://github.com/theislab/ehrapy/blob/13f5e699cb00ef81fa00619942f942d7a2b8218f/ehrapy/api/preprocessing/_scanpy_pp_api.py#L116
> 
> There's also `log1p()` and `scale()` but maybe there are other uses for those?

Oh, true. I forgot those. I think that log1p should stay, but scale is some kind of normalization for sure, right? What do you think? Should we rewrap these (or just scale) or just remove them?

> > I think that you are looking for `**kwargs`. [book.pythontips.com/en/latest/args_and_kwargs.html](https://book.pythontips.com/en/latest/args_and_kwargs.html) You can pass them to every normalization function.
> 
> How would you handle a situation where you are normalising with multiple wrapped methods and some of those methods have parameters with the same name and you want to set them to different values? I couldn't find an example at the moment but could definitely be possible at some stage if more methods were wrapped.

Now that's tricky. Brainstorming:
One solution could be that we always require a prefix for keyword arguments that overlap. 

Example:
There are two functions (a and b) which both have the same keyword x.

We have our normalize function -> def normalize(adata: AnnData, methods: dict[str, list[str]], **kwargs).
Now this normalize function can call function `a` and `b` depending on the key of `methods`. The trick is that we require the `**kwargs` to prefix their arguments with the method if a parameter overlaps for methods.

1. non-overlapping case -> normalize(adata, methods_dict, y=6, z=20)
2. overlapping case -> normalize(adata, methods_dict, max_abs_x=5, robust_scaling_x=10)

Case 1 is trivial, we just forward the kwargs.
Case 2 is where we need to pass the correct kwargs to the appropriate function

_max_abs(adata, my_columns, x=maxabs_x, **kwargs)

_robust_scaling(adata, my_columns, x=robust_scaling_x, **kwargs)

Not pretty but if documented it should be fine, I guess? What do you think?

Zethson:
An alternative could be that we throw away the "master normalize" method and only provide access to the underlying single normalization methods. Then there would not be any kwargs ambiguity.

lazappi:
> Oh, true. I forgot those. I think that log1p should stay, but scale is some kind of normalization for sure, right? What do you think? Should we rewrap these (or just scale) or just remove them?

Scaling we pretty much have implemented already (but a bit less flexible because of the parameters thing) so that can probably be removed. It is only applied to numeric variable though. `log1p` we haven't done so I guess the question is whether that should also only be applied to numeric variables (probably yes?). If we do reimplement it should it go into the current framework or as a separate function?

Zethson:
> > Oh, true. I forgot those. I think that log1p should stay, but scale is some kind of normalization for sure, right? What do you think? Should we rewrap these (or just scale) or just remove them?
> 
> Scaling we pretty much have implemented already (but a bit less flexible because of the parameters thing) so that can probably be removed. It is only applied to numeric variable though. `log1p` we haven't done so I guess the question is whether that should also only be applied to numeric variables (probably yes?). If we do reimplement it should it go into the current framework or as a separate function?

Let's remove scaling. log1p only for numerical + current framework then, please.

lazappi:
> An alternative could be that we throw away the "master normalize" method and only provide access to the underlying single normalization methods. Then there would not be any kwargs ambiguity.

Yeah, that would be a design choice about what is easier for users/developers I guess. Could also keep the master method but remove the option for applying multiple methods at once. Maybe a compromise that means the user still only sees one function but we avoid this arguments issue.

Another option would be to add arguments for things we think people would want to change but that could get messy in the long run.

Zethson:
> Yeah, that would be a design choice about what is easier for users/developers I guess. Could also keep the master method but remove the option for applying multiple methods at once. Maybe a compromise that means the user still only sees one function but we avoid this arguments issue.

Think that this would be inconsistent with the design of the rest though. The encoding also has such a master method whereas the imputation specifically does not have a master method and we offer every method one by one.

I am fine with both options, but not a "hybrid" that you described.

>  Another option would be to add arguments for things we think people would want to change but that could get messy in the long run.

Yeah, this would be bad. Difficult to maintain and to document.

lazappi:
Ok I have added a basic record of the applied normalisations. It's a bit flakey in that if somebody manually replaces `X` then it won't be correct anymore. Also added `log1p` and `sqrt` methods and removed the Scanpy `normalize_total`, `log1p` and `scale` functions.

ehrapy/api/preprocessing/_normalization.py
```diff
@@ -219,6 +226,25 @@ def _norm_power_box_cox(values: np.ndarray) -> np.ndarray:
         return power_transform(values, method="box-cox")
 
 
+def _norm_log1p(values: np.ndarray, base: int | float | None) -> np.ndarray:
```
Zethson, line 233: What do you think about having a default value for this normalization? If we set one we should also document it.

lazappi: For the base? The default will be natural log if `base = None` which I think I did in the master function but not here I guess. Could add it here as well.

Zethson: Yes, for the base. I couldn't see a default in the master function. Maybe I am blind?

lazappi: Here (L25) after all the `|` (which I think is the right syntax) 

```python
def normalize(
    adata: AnnData, methods: dict[str, list[str]] | str, base: int | float | None = None, copy: bool = False
)
```

Zethson: Ah yeah, think we misunderstood each other. Yes, internally the default will be set to the natural log, but this might not be visible or known to the user. If we make it explicit here it is more clear. We can also just document this behavior.

Not a big issue.

PR #171, opened 2021-12-05
Refactoring: IO and encode part

### LINKED ISSUE #177: Encoding fails if there is no "original" layer
**Describe the bug**

If the AnnData object is created by `ehrapy.io.read` a layer called "original" is added. The encoding function assumes this is present and fails with an unclear error if it isn't. 

**To Reproduce**

```python
from anndata import AnnData
import ehrapy

X = np.array([[1, 3.4, "A string", "A different string"], [2, 5.4, "Silly string", "A different string"], [2, 5.7, "A string", "What string?"]], dtype=np.dtype(object))
adata = AnnData(X=X_strings, dtype=np.dtype(object))

# Encoding fails
ehrapy.preprocessing.encode(adata, autodetect=True)
KeyError: 'original'

# If we add "original" first it works
adata.layers["original"] = adata.X.copy()
ehrapy.preprocessing.encode(adata, autodetect=True)
AnnData object with n_obs × n_vars = 3 × 4
    obs: '2', '3'
    uns: 'categoricals', 'original_values_categoricals', 'current_encodings'
    layers: 'original'
```

**Expected behavior**

Either the "original" layer is created automatically or a clearer error saying that it is missing and how to add it

**System [please complete the following information]:**

-   OS: MacOS 10.15
-   Language Version: Python 3.8
-   Virtual environment: Conda

Imipenem:
**PR Checklist**

**Description of changes**

__IO__

- `read()` has now a `dataset_path` arg, which can be used to name a downloaded directory/file

- renamed some args to make clear that it can either read a single file or from a directory

- removed unused parts of `read()` and unsupported file format (.txt)

- added more archive formats (only .zip before) --> now supporting zip, tar, gztar, bztar, xztar

__Caching__

- added caching for multiple AnnData objects

- fixed an issue that causes obs to be deleted when readings from cache file

- fixed an issue that caused columns_obs_only to be deleted when using caching/reading from cache

- fixed an issue that caused index column to be not set when using caching/reading from cache

- index column now only allows passing a string (for a single object) or a dict (for multiple files, but only with one string value per object indicating the index column name)

- fixed type annotation for col_obs_only and index column

- added an error message indicating that caching is currently not supported for MuData objects

__Encoding__

- fixes #177

- fixed a bug that caused ehrapy to crash when autodetect was used, but no encodings found

__General__

- changed default _write/read/cache directory_ in ehrapy settings to be "ehrapy_unique"

- refactored some docstrings and naming in general to reflect the changes made in this PR

- removed multipledispatch dependency, as it's not needed currently

**Technical details**


ehrapy/api/data/__init__.py
```diff
@@ -2,7 +2,7 @@
 from ehrapy.api.data.datasets import Datasets
 
 
-def mimic_2(encode: bool = False):
+def mimic_2_demo(encode: bool = False):
```
Zethson, line 5: That's not quite correct.

We are using the full mimic_2_IAC dataset. We should actually name this mimic_2_iac or MIMIC_II_IAC or something along those lines.

The MIMIC-III dataset that we are using is explicitly only a demo dataset. It is not the full one.

ehrapy/api/tools/__init__.py
```diff
@@ -1,4 +1,6 @@
 from ehrapy.api.tools._hpo import HPO
-from ehrapy.api.tools._scanpy_tl_api import *  # noqa: E402,F403
 from ehrapy.api.tools.nlp._medcat import MedCAT
 from ehrapy.api.tools.nlp._translators import DeepL
+
+# TODO: uncomment when fa2 issue got fixed (see scanpy)
```
Zethson, line 5: No. This cannot be commented! We need this import. We need to restrict users to Python 3.8 and that's it for the moment.

pyproject.toml
```diff
@@ -34,13 +34,13 @@ mudata = "^0.1.0"
 leidenalg = "^0.8.7"
 deepl = "^1.2.0"
 pynndescent = "^0.5.4"
-fa2 = "^0.3.5"
+# TODO Uncomment when fa2 issue got fixed (see scanpy)
```
Zethson, line 37: No. This is a required dependency.

Zethson (commented):
Couple of major comments.

Zethson:
Would also be good if you could look at:

1. https://github.com/theislab/ehrapy/blob/97a62f6a1e001fea29dee8722087c0edc86bedf1/ehrapy/api/preprocessing/encoding/_categoricals.py#L81 <- guess an artificial dataset can help for this.
2. https://github.com/theislab/ehrapy/blob/97a62f6a1e001fea29dee8722087c0edc86bedf1/ehrapy/api/preprocessing/encoding/_encode.py#L469
3. https://github.com/theislab/ehrapy/blob/97a62f6a1e001fea29dee8722087c0edc86bedf1/ehrapy/api/io/read.py#L67

Imipenem:
Mostly done; just to want to glimpse over it tomorrow to get it ready for review. 
More fixes and work than expected ;).

ehrapy/api/io/read.py
```diff
@@ -31,62 +25,69 @@ class DataReader:
...
     @staticmethod
     def read(
-        file_path: Union[Path, str],
-        download_file_name: Optional[str] = None,
+        dataset_path: Union[Path, str],
         extension: Optional[str] = None,
         delimiter: Optional[str] = None,
-        index_column: Union[str, Optional[int]] = None,
-        columns_obs_only: Optional[List[Union[str]]] = None,
+        index_column: Optional[Union[Dict[str, Union[str, int]], Union[str, int]]] = None,
```
Zethson, line 31: Stuff like this really shows that we need Python 3.10 haha

ehrapy/api/io/read.py
```diff
@@ -31,62 +25,69 @@ class DataReader:
...
+            dataset_path: Path to the file or directory to read.
             extension: File extension. Required to select the appropriate file reader.
             delimiter: File delimiter. Required for e.g. csv vs tsv files.
-            index_column: The index column of the tables.
+            index_column: The index column of obs.
             columns_obs_only: Which columns to only add to obs and not X.
             return_mudata: Whether to create and return a MuData object.
             cache: Whether to use the cache when reading.
-            backup_url: URL to download the data file from if not yet existing.
+            download_dataset_name: Name of the file or directory, in case the dataset is downloaded
```
Zethson, line 49:
```suggestion
            download_dataset_name: Name of the file or directory in case the dataset is downloaded
```

ehrapy/api/io/read.py
```diff
@@ -31,62 +25,69 @@ class DataReader:
...
         if not file.exists():
+            if backup_url is not None:
+                download_default_name = backup_url.split("/")[-1]
+                download_dataset_name = download_dataset_name or download_default_name
+                # currently supports zip, tar, gztar, bztar, xztar
+                archive_formats, _ = zip(*shutil.get_archive_formats())
+                is_archived = download_default_name[-3:] in archive_formats
+
+            else:
+                raise BackupURLNotProvidedError(
```
Zethson, line 71: I like this

ehrapy/api/io/read.py
```diff
@@ -104,96 +105,111 @@ def _read(
...
+            that are obs only for each object
         """
+        obs_only_all = {}
         if not return_mudata_object:
             anndata_dict = {}
         else:
             mudata = None
         for file in filename.iterdir():
-            if file.is_file() and file.suffix in [".csv", ".tsv"]:
+            if file.is_file() and file.suffix in {".csv", ".tsv"}:
```
Zethson, line 203: Why the change to curly brackets?

ehrapy/api/preprocessing/encoding/_categoricals.py
```diff
@@ -78,7 +78,7 @@ def _is_categorical_column(col: np.ndarray, col_name: str) -> CategoricalColumnT
         )
         sys.exit(1)
     if c_dtype == "string":
-        # TODO Consider this when we are dealing with freetext
+        # As discussed: will currently leave it as it is; freetext -> medcat first!
```
Zethson, line 81: Maybe open an issue for the decorator idea?

Zethson (approved):
Nice PR! Did not spot anything that really stuck into my eye. But I guess I'll need to play around with things later.

Can we add 1-2 tests for the caching?

Feel free to merge when you addressed my comments.

ehrapy/api/io/read.py
Imipenem: ;D and we need multi dispatch as well

Zethson:
@Imipenem oh and fix the merge conflicts please

ehrapy/api/io/read.py
Imipenem: Definitely not needed; i was just going over all lines , where we lookup anything in a collection and was refactoring it to set, if possible (for the faster lookup). But that's obviously not needed here, just to keep it consistent.

ehrapy/api/preprocessing/encoding/_categoricals.py
Imipenem: Ah, yes will do asap

ehrapy/api/io/read.py
Zethson: I don't mind it :)

ehrapy/api/preprocessing/encoding/_categoricals.py
Imipenem: See #182

PR #203, opened 2021-12-29
Add new Sphinx documentation structure with Sphinx gallery

Zethson:
Adds a whole new documentation structure based on the pydata_sphinx_theme instead of sphinx_rtd. This means that we also removed the sphinx-rtd-dark-theme. It was always just working close to 95% and not in all cases. A darkmode for pydata_sphinx_theme is in the works.

Looks like: 

image
image
image
image

TODO in a follow up PR:
- [ ] Move tutorials to their own repository
- [ ] Trigger tutorial installation workflows via https://github.com/convictional/trigger-workflow-and-wait

Alternative could be to simply checkout the complete tree with the submodule and then CD and run all notebooks. Likely easier.


Zethson:
Autosummary doesn't seem to find stuff. Not sure why yet.

Zethson:
```
nox > Command python -m pip install --constraint=.nox/mypy-3-8/tmp/requirements.txt mypy pytest types-pkg-resources types-requests types-attrs failed with exit code 1:
ERROR: Cannot install mypy because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested mypy
    The user requested (constraint) mypy==0.930

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies

```

I don't get it.

PR #208, opened 2022-01-02
Add data imputation

Imipenem:
**PR Checklist**

**Description of changes**

- implemented replace_explicit Imputation

- implemented KNN Imputation

- implemented MissForest Imputation

- implemented Mean/median and most_frequent Imputation

- added a check when reading an AnnData object for X's dtype to be float from the start if all data is numerical (except for missing data)


Imipenem:
So here's the thing with KNN Imputation: KNN Imputation only works on numerical data. So to impute non-numerical data, this data needs to be encoded. Ehrapy can, of course, encode the data, but, in contrast to working on a `pandas Dataframe`, information on WHERE the missing values are is lost (at least in X, our main data). This is different to a pandas dataframe, since pandas stores a mask in the background where missing values actually are and imputes on this mask.

So we can call KNN imputation on our encoded `X` but it will NOT have ANY EFFECT. This is because the missing values will be encoded too (for example with `one_hot_encoding` or `label_encoding`), so on an abstract level (for the imputer that will be called later on), they're technically not missing (altough they are in reality).

So TL;DR: I think we have to do the following when using KNN Imputation: 

1. Get our `AnnData` object passed.
2. Transform `AnnData object` (**unencoded**) to a `pandas Dataframe`
3. Encode the Dataframe using any Encoder like `OneHot` or `Label` (does not really matter though, I guess)
4. Call  KNN impute` on the **Encoded** `Dataframe`
5. Transform this imputed` Dataframe` back to an `AnnData `object (while ensuring it matches "ehrapy" requirements); if possible, decode here as well.

Any Comments? @Zethson

Zethson:
Good finding and explanation.
Your suggestion totally makes sense and is certainly a workaround that will work. We could implement this as an *option*.

Have you seen https://stackoverflow.com/questions/64900801/implementing-knn-imputation-on-categorical-variables-in-an-sklearn-pipeline ? Think that they also have some nice suggestions for dealing with this.

Imipenem:
> Have you seen https://stackoverflow.com/questions/64900801/implementing-knn-imputation-on-categorical-variables-in-an-sklearn-pipeline ? Think that they also have some nice suggestions for dealing with this.

Nice. But I guess, we should then also be able to somehow "decode" the imputed data (minus the imputed values, which can be stored as a mask). 
So the second suggestion could work with my addition (cause an Imputer should not return encoded data IMO, thats not his responsibility and I guess no user would expect this). 

Could also implement third and first suggestion as another option.

Zethson:
> So the second suggestion could work with my addition (cause an Imputer should not return encoded data IMO, thats not his responsibility and I guess no user would expect this).
> 
> Could also implement third and first suggestion as another option.

Yup, fully agree on the first part. Would be an unintended side effect. I'd also suggest to implement the other options. Should not be too difficult (I hope).

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
 import numpy as np
 import pandas as pd
 from anndata import AnnData
+from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+from sklearn.experimental import enable_iterative_imputer  # noqa: F401
```
Imipenem, line 7: needed: see https://stackoverflow.com/questions/55846680/can-not-import-iterativeimputer-from-sklearn-impute

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+        _knn_impute(adata, **kwargs)
+        # decode ordinal encoding to obtain imputed original data
+        adata.X = enc.inverse_transform(adata.X)
+
+    return adata
+
+
+def _knn_impute(adata: AnnData, **kwargs) -> None:
+    """Utility function to impute data using KNN-Imputer"""
+    n_neighbors = kwargs.get("n_neighbors")
```
Imipenem, line 215: @Zethson ; might be better to remove this one. Especially for imputing non-numerical data using KNN its important to only use the next nearest neighbor, instead of rendering a meaningless mean of some values, that could not be decoded afterwards. Thoughts?

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    Args:
+        adata: The AnnData object to use MissForest Imputation on
+        **kwargs: Keyword args
+                  replacement: A list of strings indicating which columns to use MissForest imputation on (if None -> all columns)
+
+    Returns:
+            The imputed (but unencoded) AnnData object
+    """
+    # TODO: passe replacement kwarg to indicate which column is numeric and which is mixed data (for those that should be imputed using MissForest strategy)
+    non_num_indices = _get_non_numerical_column_indices(adata.X)
```
Imipenem, line 243: Any ideas for a more efficient implementation here @Zethson ?

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+
+def _is_float(val):
+    try:
+        float(val)
+    except ValueError:
+        if val is np.nan:
+            return True
+        return False
+    else:
+        # required to keep original booleans, otherwise those would be casted into ints; Keep or document?
```
Imipenem, line 282: Thoughts @Zethson ?

ehrapy/api/io/_read.py
```diff
@@ -497,14 +498,19 @@ def df_to_anndata(df: pd.DataFrame, columns_obs_only: list[str] | None, index_co
         df = df.set_index(index_column)
     # move columns from the input dataframe to later obs
     dataframes = _move_columns_to_obs(df, columns_obs_only)
+    all_num = False
+    # if data is numerical only, short-circuit AnnData creation to have float dtype instead of object
+    if all(np.issubdtype(column_dtype, np.number) for column_dtype in dataframes.df.dtypes):
```
Zethson, line 503: Can't this be shortened to `all_num = all(np.issubdtype(column_dtype, np.number) for column_dtype in dataframes.df.dtypes)`?

ehrapy/api/preprocessing/__init__.py
```diff
@@ -1,3 +1,43 @@
...
+    _median,
+    _miss_forest,
+    _most_frequent,
+    _replace_explicit,
+)
 from ehrapy.api.preprocessing._scanpy_pp_api import *  # noqa: E402,F403
-from ehrapy.api.preprocessing.encoding import encode, type_overview, undo_encoding
+
+
+def impute(adata: AnnData, mode: str, copy: bool = False, **kwargs) -> AnnData:
```
Zethson, line 14: Should we really do this? Maybe we should just provide the imputation functions explicitly and get rid of this high level function. I think that it this is easier to document.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
-    3. Replace all missing values in a subset of columns with a specified value per column. ( str ,(str, int) )
+    2. Replace all missing values in a subset of columns with a specified value per column. ( str ,(str, int) )
 
     Args:
         adata: :class:`~anndata.AnnData` object containing X to impute values in
-        replacement: Value to use as replacement and optionally keys to indicate which columns to replace.
-        See scenarios above
-        impute_empty_strings: Whether to also impute empty strings
         copy: Whether to return a copy with the imputed data.
+        **kwargs: replacement: Value to use as replacement and optionally keys to indicate which columns to replace.
```
Zethson, line 24: This really should not be a kwargs in my opinion. This is a key argument. This function is useless without an explicit `replacement`.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+                  impute_empty_strings: Whether to also impute empty strings
```
Zethson, line 25: No need for a kwarg imo.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
 
     Returns:
         :class:`~anndata.AnnData` object with imputed X
+    """
+    # ensure replacement parameter has been passed when using explicit impute mode
+    try:
+        replacement = kwargs["replacement"]
+    except KeyError:
```
Zethson, line 33: No need when not using a kwarg. Will throw a missing argument error.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    """Extract the replacement value for a given column in the :class:`~anndata.AnnData` object
+
+    Returns: The value to replace missing values
+
+    """
+    # try to get a value for the specific column
+    imputation_value = replacement.get(column_name)
+    if imputation_value:
+        return imputation_value
+    # search for a default value in case no value was specified for that column
```
Zethson, line 75: Add an explicit else clause here for readability?

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+            f"no default was found!"
+        )
+
+
+# ===================== Mean Imputation =========================
+
+
+def _mean(adata: AnnData, **kwargs) -> AnnData:
+    """
+    Impute AnnData object using mean imputation. This works only for numerical data.
```
Zethson, line 91:
```suggestion
    Impute AnnData object using mean imputation. This only works for numerical data.
```

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+
+    Args:
+        adata: The AnnData object to use mean Imputation on
+        **kwargs: Keyword args
+                  replacement: A list of strings indicating which columns to use mean imputation on (if None -> all columns)
```
Zethson, line 96: No. This is a key argument.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    except ValueError:
+        raise ImputeStrategyNotAvailableError(
+            "Can only impute numerical data using mean strategy. Try to restrict imputation"
+            "to certain columns using replacement parameter or use a different mode."
+        )
+
+    return adata
+
+
+def _median(adata: AnnData, **kwargs) -> AnnData:
```
Zethson, line 122: Again, no kwarg, please.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    except ValueError:
+        raise ImputeStrategyNotAvailableError(
+            "Can only impute numerical data using median strategy. Try to restrict imputation"
+            "to certain columns using replacement parameter or use a different mode."
+        )
+
+    return adata
+
+
+def _most_frequent(adata: AnnData, **kwargs) -> AnnData:
```
Zethson, line 155: kwarg

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    return adata
+
+
+# ===================== KNN Imputation =========================
+
+
+def _knn(adata: AnnData, **kwargs) -> AnnData:
+    """Impute data using the KNN-Imputer.
+    When using KNN Imputation with mixed data (non-numerical and numerical), encoding using ordinal encoding is required
+    since KNN Imputation can only work on numerical data. The encoding itself is just a utility and will be redone once
```
Zethson, line 187:
```suggestion
    since KNN Imputation can only work on numerical data. The encoding itself is just a utility and will be undone once
```

undone and not redone, right?

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+        adata.X = enc.fit_transform(adata.X)
+        # impute the data using KNN imputation
+        _knn_impute(adata, **kwargs)
+        # decode ordinal encoding to obtain imputed original data
+        adata.X = enc.inverse_transform(adata.X)
+
+    return adata
+
+
+def _knn_impute(adata: AnnData, **kwargs) -> None:
```
Zethson, line 213: kwarg

ehrapy/api/preprocessing/_data_imputation.py
Zethson: Let's discuss today.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+        adata.X = imputer.fit_transform(adata.X)
 
-    # scenario 2: Replace all missing values in a subset of columns with the specified value
-    # TODO
 
-    # scenario 3: Replace all missing values in a subset of columns with a specified value per column
-    # TODO
+def _miss_forest(adata: AnnData, **kwargs) -> AnnData:
+    """Impute data using the MissForest stratgey. See https://academic.oup.com/bioinformatics/article/28/1/112/219101.
+    This requires the computation of which columns in X contain numerical only (including NaNs) and which mixed data, which could be an
```
Zethson, line 229: Do you think that we should support mixed data in the first place? Let's discuss.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    expensive operation on a larger X.
+    Speeding this up could be achieved by passing the "replacement" kwarg and indicate, which column is
+    numerical only and which mixed data.
+
+    Args:
+        adata: The AnnData object to use MissForest Imputation on
+        **kwargs: Keyword args
```
Zethson, line 236: Please no ^^

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    is_numeric_numpy = np.vectorize(_is_float, otypes=[bool])
+    mask = np.apply_along_axis(is_numeric_numpy, 0, X)
+
+    _, column_indices = np.where(~mask)
+    non_num_indices = set(column_indices)
+
+    return non_num_indices
+
+
+def _is_float(val):
```
Zethson, line 274: I have the impression that I've seen this function somewhere already. Maybe in `encode` or something?

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    try:
+        float(val)
+    except ValueError:
+        if val is np.nan:
+            return True
```
Zethson, line 279: Then the function should be renamed to _is_float_or_nan

ehrapy/api/preprocessing/_data_imputation.py
Zethson: Fine with me if the function is named accordingly and the behavior + reasoning is well documented.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+        return False
+    else:
+        # required to keep original booleans, otherwise those would be casted into ints; Keep or document?
+        if val is not False and val is not True:
```
Zethson, line 283: Possibly easier to check if val is instance of bool?

ehrapy/api/preprocessing/_data_imputation.py
Zethson, line 215: I'd also make this a proper parameter. Will be a better interface if the high level impute method is removed as mentioned above.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,300 @@
...
+    return adata
+
+
+def _knn_impute(adata: AnnData, **kwargs) -> None:
+    """Utility function to impute data using KNN-Imputer"""
+    n_neighbors = kwargs.get("n_neighbors")
+    replacement = kwargs.get("replacement")
+    imputer = KNNImputer(n_neighbors=n_neighbors if n_neighbors else 1)
+
+    if isinstance(replacement, list):
```
Zethson, line 219: Replacement is not the proper argument name here. I'd remove the kwarg as mentioned many times and add a new parameter "columns: str | List[str]" which defaults to "all" or something.

Zethson (changes requested):
Really not a fan of your usage of kwargs. `replacement` is an absolute core argument that cannot be missing.

Imipenem:
@Zethson; just finishing this PR and I noticed we need to support imputation for multiple AnnData objects as well: As I'm writing this currently I wonder if we should provide two methods per imputation mode, one for single AnnData and the other for multiple (only at an API level, under the surface they're using the same functions). This would ease a lot of the "not understandable" type hints and a few lines of type checking for each mode. Thoughts?

Zethson:
> @Zethson; just finishing this PR and I noticed we need to support imputation for multiple AnnData objects as well.

We do? Why would we?

Imipenem:
> > @Zethson; just finishing this PR and I noticed we need to support imputation for multiple AnnData objects as well.
> 
> We do? Why would we?

I mean we should (we do not currently). Because if not, users have to write their own script in order to impute multiple Anndata objects (consider our MIMIC III test dataset for example).

Zethson:
> > > @Zethson; just finishing this PR and I noticed we need to support imputation for multiple AnnData objects as well.
> > 
> > 
> > We do? Why would we?
> 
> I mean we should (we do not currently). Because if not, users have to write their own script in order to impute multiple Anndata objects (consider our MIMIC III test dataset for example).

That's fine. One has to examine the specific AnnData objects before imputing anyways. It is also extremely unlikely that one ever uses 26 AnnData objects for a single analysis anyways. Even 2-3 are already a lot.

Don't worry about such use-cases for now. Not an issue.

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,236 @@
...
+
+# ===================== KNN Imputation =========================
+
+
+def _knn(adata: AnnData, var_names: list[str] | None) -> AnnData:
+    # numerical only data needs no encoding since KNN Imputation can be applied directly
+    if np.issubdtype(adata.X.dtype, np.number):
+        _knn_impute(adata, var_names)
+    else:
+        # ordinal encoding is used since non-numerical data can not be imputed using KNN Imputation
```
Zethson, line 130: Should this be wrapped with a try except where we can revert the changes if something goes wrong?

Zethson (changes requested):
I'd always remove the empty new line after the `Returns:` section.

I actually wouldn't write an interface in the `__init__.py`. Let's just import these functions. Much easier to maintain and with less code duplication.

We now have 3 very very comparable implementations for mean, median & most frequent. I think that it would be fine to merge them under "simple_imputation" or something and parameterize the strategy. My gut feeling tells me that "explicit" should also be a part of simple then (although it is implemented differently). What do you think?

The `var_names: list[str]` parameter of mean, median etc should be Collection[str] with the corresponding isinstance test below. We should always try to program against the interfaces and not the explicit types whenever possible.

Same for `var_names: dict[str, list[str]]`. If these can be Mappings - great.

Further, you are raising a couple of errors from None. Not really a fan of this. Yes, the traceback might be shorter, but it is masking the origin of the error to some extent. Ideally, our errors are 100% descriptive and reliable, but well - we can't guarantee that, right? Thoughts?

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,236 @@
...
+        adata.X[::, num_indices] = imp_num.fit_transform(adata.X[::, num_indices])
+    if non_num_indices:
+        adata.X[::, non_num_indices] = imp_cat.fit_transform(adata.X[::, non_num_indices])
+        # decode ordinal encoding to obtain imputed original data
+        adata.X[::, non_num_indices] = enc.inverse_transform(adata.X[::, non_num_indices])
 
     return adata
+
+
+def _get_non_numerical_column_indices(X: np.ndarray) -> set:
```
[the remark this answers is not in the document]
Imipenem: We can move it, but makes no sense to dispatch it, since this would only work with a numpy array (so `X` in our case).

ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -3,51 +3,236 @@
...
 
-    # scenario 3: Replace all missing values in a subset of columns with a specified value per column
-    # TODO
+    # infer non numerical and numerical indices automatically
+    else:
+        non_num_indices_set = _get_non_numerical_column_indices(adata.X)
+        num_indices = [ind for ind in range(adata.X.shape[1]) if ind not in non_num_indices_set]
+        non_num_indices = list(non_num_indices_set)
+
+    imp_num = IterativeImputer(estimator=ExtraTreesRegressor(), initial_strategy="mean", max_iter=10, random_state=0)
```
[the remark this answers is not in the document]
Imipenem: Should not parametrize the initial strategy here, since (from what I've read) its the common way for MissForest to use `mean`. 
We can parametrize `max_iter` of course, but one must be aware that, the larger this one got (and the more numerical columns `X` has, that will be imputed) this will result in an extrem long runtime.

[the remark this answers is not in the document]
Zethson: I would parameterize all of them with our defaults? Think that's reasonable.

ehrapy/api/preprocessing/_data_imputation.py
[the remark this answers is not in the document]
Zethson: Yeah, one could select X from an AnnData object of course, but yeah, let's not make it complicated.

We can also keep it here, idc.

Imipenem:
> 
> 
> I'd always remove the empty new line after the Returns: section.

Agreed. Thought this is standard default.

>   I actually wouldn't write an interface in the __init__.py. Let's just import these functions. Much easier to maintain and with less code duplication.

Agree. Was also my feeling while implementing this.

>   We now have 3 very very comparable implementations for mean, median & most frequent. I think that it would be fine to merge them under "simple_imputation" or something and parameterize the strategy. My gut feeling tells me that "explicit" should also be a part of simple then (although it is implemented differently). What do you think?

That was one thing, I also had in mind. I initially decided against it to be consistent from a user perspective (one function per impute mode), but makes more sense from code perspective and is not that hard to remember (especially since it uses the same Imputer from `scikit learn` anyways).

>   The var_names: list[str] parameter of mean, median etc should be Collection[str] with the corresponding isinstance test below. We should always try to program against the interfaces and not the explicit types whenever possible.
>   Same for var_names: dict[str, list[str]]. If these can be Mappings - great.

I can't see any advantage here using a more generic type. `var_names` should always be a List and nothing else. I might miss your point here?
  
>   Further, you are raising a couple of errors from None. Not really a fan of this. Yes, the traceback might be shorter, but it is masking the origin of the error to some extent. Ideally, our errors are 100% descriptive and reliable, but well - we can't guarantee that, right? 

It depends I'd say. Here I'd agree and say it would be better to take the longer traceback in exchange to not masking the source of the error (altough we can be very confident here (but not 100% as you stated) that this is indeed the source of the error).

Zethson:
Happy to see that we agree on pretty much everything.

> I can't see any advantage here using a more generic type. var_names should always be a List and nothing else. I might miss your point here?

All right, if it needs to be exactly that then it's ok. I can't find the post anymore where someone specifically stated to always program against the interfaces. I'll try to dig it up...

Let's keep it as you have it.

Issue #137, opened 2021-10-25
Add simple AnnData -> Pandas DF and Pandas DF -> AnnData


Zethson:
When wrangling with data it might be easier to first work with a Pandas object and to then transform it into an AnnData object. 
A part of the reasoning for this is that it is very difficult to get an overview of the values stored in X together with their headers. We could either improve this or try to support the easy transformation between the two formats.

Imipenem:
> https://anndata.readthedocs.io/en/latest/anndata.AnnData.to_df.html
> 
> This actually exists!

I think we have to adapt this for our data though (for example datetime data is obs only, what to do with encoded data etc.)

Zethson:
> encoded data

Think we can just keep that. Would not add any special cases for this.

> datetime data is obs only

Good point. Don't think that we should automatically append all `obs` columns to X and then transform to a DF. I'd suggest that we simply parameterize this initially for both directions. Basically a parameter `obs_columns` for AnnData -> DataFrame and the usual `obs_only` for DataFrame -> AnnData

PR #244, opened 2022-01-11
[Feature]#137: Add pandas DF -> AnnData and AnnData -> pandas DF

Imipenem:
**PR Checklist**

**Description of changes**

- implemented a function to convert a given Anndata object to a pandas dataframe

- moved the existing df_to_anndata function from IO part to util part and expose it via the API

### DIFF
--- a/ehrapy/api/_anndata_util.py
+++ b/ehrapy/api/_anndata_util.py
@@ -1,7 +1,98 @@
 from __future__ import annotations
 
+from typing import NamedTuple
+
 import numpy as np
+import pandas as pd
 from anndata import AnnData
+
+
+class BaseDataframes(NamedTuple):
+    obs: pd.DataFrame
+    df: pd.DataFrame
+
+
+def df_to_anndata(
+    df: pd.DataFrame, columns_obs_only: list[str] | None = None, index_column: str | None = None
+) -> AnnData:
+    """Transform a given pandas dataframe into an AnnData object.
+
+    Args:
+        df: The pandas dataframe to be transformed
+        columns_obs_only: An optional list of column names that should belong to obs only and not X
+        index_column: The (optional) index column of obs
+
+    Returns:
+        An AnnData object created from the given pandas dataframe
+    """
+    if index_column:
+        df = df.set_index(index_column)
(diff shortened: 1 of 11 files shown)


Imipenem:
@Zethson not quite happy with location of the two functions. Any suggestions?

ehrapy/api/_anndata_util.py
```diff
@@ -1,9 +1,100 @@
...
+                "Make sure you spelled the column names correctly."
+            )
+    else:
+        obs = pd.DataFrame(index=df.index.map(str))
+
+    return BaseDataframes(obs, df)
+
+
+def anndata_to_df(adata: AnnData, add_from_obs: list[str] | str | None = None) -> pd.DataFrame:
+    """Transform an AnnData object to a pandas Dataframe
```
Zethson, line 75:
```suggestion
    """Transform an AnnData object to a pandas Dataframe.

```

Zethson (changes requested):
LGTM

Three things:
1. Please add a few tests. Think these methods can be well tested.
2. Please add these functions to the documentation
3. Let's move these functions to somewhere else as discussed. Where? TBD.

Zethson:
I think that your submodule is not up to date. You can set git pull to also always pull the submodule. Google how to.

Zethson (commented):
I'd suggest that you use a setup_method in the tests to create these objects? Think you have some things that are repeated.

Imipenem:
> I'd suggest that you use a setup_method in the tests to create these objects? Think you have some things that are repeated.


Hm I see your point but unfortunately, this would require a lot of customization. Each test function requires different columns and test values, not sure if its worth the effort.

Zethson:
If it allows you to more quickly move to the other issues I am fine with keeping them ugly for now. Although it hurts my soul ^_^

Imipenem:
> If it allows you to more quickly move to the other issues I am fine with keeping them ugly for now. Although it hurts my soul ^_^

Wrote before

> If it allows you to more quickly move to the other issues I am fine with keeping them ugly for now. Although it hurts my soul ^_^

Think first, write then -> applies to my comment from above^^

Issue #151, opened 2021-11-15
Add function to move data from X to obs and vice versa


Zethson:
Would be very useful to have a method to move data from X to obs and the other way around. I am sure that when loading complex data one sometimes forgets to add a column to "obs_only" and instead of loading the complete (large) file again it would help to just move columns.

ivirshup:
Could you propose an API here? Also, I imagine `obsm`, `var`, and `varm` would be good destinations too?

Zethson:
> Could you propose an API here? Also, I imagine `obsm`, `var`, and `varm` would be good destinations too?

Yes, to all. We'll discuss.

ivirshup:
<details>
<summary> Drafts in case you want to play around with the API: </summary>

```python
from typing import Union

import anndata as ad, numpy as np, pandas as pd

def split_out(adata: ad.AnnData, idx: "np.ndarray[1, bool]", *, axis=1):
    idxs = [slice(None), slice(None)]
    idxs[axis] = idx
    idxs = tuple(idxs)
    df = adata[idxs].to_df()
    if axis == 1:
        adata._inplace_subset_var(~idx)
        adata.obs = adata.obs.join(df)
    elif axis == 0:
        adata._inplace_subset_obs(~idx)
        adata.var = adata.var.join(df)


def splice_in(
        adata: ad.AnnData,
        *,
        obs: Union[str, list[str]]=None,
        var: Union[str, list[str]]=None,
    ) -> ad.AnnData:
    assert (obs is None) + (var is None) == 1
    if obs is not None:
        if isinstance(obs, str): obs = [obs]
        res = ad.concat([adata, ad.AnnData(adata.var[obs])], axis=0)
        res.var.drop(columns=obs, inplace=True)
        return res
    elif var is not None:
        if isinstance(var, str): var = [var]
        res = ad.concat([adata, ad.AnnData(adata.obs[var])], axis=1)
        res.obs.drop(columns=var, inplace=True)
        return res


from anndata.tests.helpers import gen_adata

a = gen_adata((20, 10))
b = gen_adata((20, 5))
c = ad.concat({"a": a, "b": b}, axis=1, index_unique="-", label="vartype")
d = c.copy()
d
```

```
AnnData object with n_obs × n_vars = 20 × 15
    var: 'var_cat', 'cat_ordered', 'int64', 'float64', 'uint8', 'vartype'
    varm: 'array', 'sparse', 'df'
    layers: 'array', 'sparse'
```

```python
[...]

PR #250, opened 2022-01-17
[Feature] API for moving columns from X to obs

Imipenem:
Adresses Issue #151.

**PR Checklist**

**Description of changes**

- added a function to move columns from X to obs

- X -> obs as well

- numerical/non-numerical columns will now be updated when encoding/or moved from/to X


ehrapy/api/anndata_ext.py
```diff
@@ -102,6 +102,50 @@ def anndata_to_df(adata: AnnData, add_from_obs: list[str] | str | None = None) -
...
+    Args:
+        adata: The AnnData object
+        to_x: The columns to move to X
+
+    Returns:
+        A new AnnData object with moved columns from obs to X. This should not be used for datetime columns currently.
+    """
+    if isinstance(to_x, str):
+        to_x = [to_x]
+    new_adata = concat([adata, AnnData(adata.obs[to_x], dtype="object")], axis=1)
```
Imipenem, line 139: @Zethson : I noticed the following issue: This will create an `obs` where all columns have `object` dtype regardless whether they are numerical or not. Do you see a way to get this info once its "lost" (best case not by trying to type cast). Or do we need to store this explicitly? Any ideas?

ehrapy/api/anndata_ext.py
```diff
@@ -102,6 +102,50 @@ def anndata_to_df(adata: AnnData, add_from_obs: list[str] | str | None = None) -
...
+        to_obs: The columns to move to obs
+
+    Returns:
+        The original AnnData object with moved columns from X to obs
+    """
+    if isinstance(to_obs, str):
+        to_obs = [to_obs]
+    # don't allow moving encoded columns as this could lead to inconsistent data in X and obs
+    if any(column.startswith("ehrapycat") for column in to_obs):
+        raise ObsMoveError("Cannot move encoded columns from X to obs. Either undo encoding or remove them from the list!")
```
Zethson, line 118: Ain't no undo encoding yet in ehrapy :)
But guess the instructions are still fine.

ehrapy/api/anndata_ext.py
Zethson: I know that this: https://github.com/theislab/anndata/blob/master/anndata/_core/anndata.py#L1188 exists. But this does not really solve the problem here, does it?

Zethson (commented):
Think that these methods should also have a `copy` parameter.

ehrapy/api/anndata_ext.py
Imipenem: We do ;). It just resets thew whole anndata object though.

Imipenem:
Agree, `move_to_obs` will return a copy by default though.

ehrapy/api/anndata_ext.py
Imipenem: Yeah not really. We would need like the other way round. I mean would be easy to code ourself, but it would include some sort of type casting, which is what I want to avoid if possible (cause I guess this would be a costly operation).

ehrapy/api/anndata_ext.py
```diff
@@ -102,6 +102,62 @@ def anndata_to_df(adata: AnnData, add_from_obs: list[str] | str | None = None) -
     return df
 
 
+def move_to_obs(adata: AnnData, to_obs: list[str] | str, copy: bool = False) -> AnnData:
+    """Move some columns from X to obs inplace
```
Zethson, line 106:
```suggestion
    """Move features from X to obs inplace.
```

ehrapy/api/anndata_ext.py
```diff
@@ -102,6 +102,62 @@ def anndata_to_df(adata: AnnData, add_from_obs: list[str] | str | None = None) -
...
+    updated_num_uns, updated_non_num_uns, num_var = _update_uns(adata, to_obs)
+    adata.obs[num_var] = adata.obs[num_var].apply(pd.to_numeric, errors="ignore", downcast="float")
+    adata.uns["numerical_columns"] = updated_num_uns
+    adata.uns["non_numerical_columns"] = updated_non_num_uns
+
+    return adata
+
+
+def move_to_x(adata: AnnData, to_x: list[str] | str) -> AnnData:
+    """Move some columns from obs to X inplace
```
Zethson, line 139:
```suggestion
    """Move features from obs to X inplace.
```

Zethson (approved):
Not the biggest fan of this return `tuple[list[str], list[str], list[str] | None]` for the `_update_uns` function, but LGTM

PR #269, opened 2022-01-21
[Fix]  Get numerical vars for normalization

### LINKED ISSUE #262: Categoricals only stored if `autodetect=True` when encoding
**Describe the bug**

The `adata.uns["categoricals"]` field is only set when `autodetect=True` during encoding. This is important because we use this field to check an object has been encoded and identify numeric variables for normalization in #162.

**To Reproduce**

Example with MIMIC-II

```python
adata = ep.dt.mimic_2(encoded=False)
adata = ep.pp.encode(
    adata, encodings={"one_hot_encoding": ["service_unit", "day_icu_intime"]}
)
"categoricals" in adata.uns_keys()
```

**Expected behavior**

Either need to make sure this is always set or have another system for identifying numeric variables.

**System [please complete the following information]:**

-   OS: MacOS
-   Language Version: Python 3.8
-   Virtual environment: conda

Imipenem:
**PR Checklist**

**Description of changes**

- fixes #262
 
- numerical vars are now stored in uns when reading

- updated caching to keep those numerical values in uns

- updated tests to reflect new changes

**Technical details**

This is a, I'd call it, potentially breaking change, cause this will detect (in some cases) more numerical columns, since they are determined when reading. @lazappi , @Zethson: Mind to take a quick look at this and identify potential issues (if any)?


ehrapy/api/anndata_ext.py
```diff
@@ -32,13 +33,18 @@ def df_to_anndata(
     # if data is numerical only, short-circuit AnnData creation to have float dtype instead of object
     all_num = all(np.issubdtype(column_dtype, np.number) for column_dtype in dataframes.df.dtypes)
     X = dataframes.df.to_numpy(copy=True)
-
+    # initializing an OrderedDict with a non-empty dict might not be intended,
+    # see: https://stackoverflow.com/questions/25480089/right-way-to-initialize-an-ordereddict-using-its-constructor-such-that-it-retain/25480206
```
Zethson, line 37: Should we investigate this at some point and add a TODO here or is this fine?

ehrapy/api/anndata_ext.py
```diff
@@ -32,13 +33,18 @@ def df_to_anndata(
...
+    uns = OrderedDict()
+    # store all numerical columns (used for normalization)
+    uns["numerical_columns"] = list(df.select_dtypes("number").columns)
```
Zethson, line 40: Should we also add a "categorical" columns here which are all columns except these? A simple diff?

I am sure that at some point this will be useful.

tests/api/preprocessing/test_normalization.py
```diff
@@ -28,14 +29,14 @@ def setup_method(self):
...
-
         self.adata = AnnData(
             X=X_data,
             obs=pd.DataFrame(data=obs_data),
             var=pd.DataFrame(data=var_data, index=var_data["Feature"]),
             dtype=np.dtype(object),
+            uns=OrderedDict(),
         )
-
+        self.adata.uns["numerical_columns"] = ["Numeric1", "Numeric2"]
```
Imipenem, line 39: Did not changed it here but to be honest: `Integer1` seems like its a numerical column and will be detected by the new function (but not from autodetect). This is due to the fact, that the sample size here are only 3 observations (unlikely to see such small sets in real world) and autodetect flagged `Integer1` as categorical, that does not need any encoding (since it has numerical dtype), but it only has like 2 different values out of 3, so autodetect flags this as categorical.

That's one of the "potential" breakings I could see (altough not a severe one and technically, this version now is the correct one).

Zethson (approved):
Thanks!
LGTM

ehrapy/api/anndata_ext.py
Imipenem: Will check this one out now. Just a few minutes.

ehrapy/api/anndata_ext.py
Imipenem: Might be useful, could also reuse this then in `autodetect`, for example

ehrapy/api/anndata_ext.py
Imipenem: > "The OrderedDict constructor and update() method both accept keyword arguments, but their order is lost because Python’s function call semantics pass-in keyword arguments using a regular unordered dictionary"

Zethson:
@Imipenem please pull the tutorial notebooks. The MIMIC-II Notebook should now run through with your fix. Please also merge development again into your branch to test this.

ehrapy/api/preprocessing/encoding/_encode.py
```diff
@@ -92,7 +90,7 @@ def undo_encoding(
     This currently resets the AnnData object to its initial state.
 
     Args:
-        adata: The :class:`~anndata.AnnData` object
+        data: The :class:`~anndata.AnnData` object
```
Zethson, line 93: If it's not just adata we should write AnnData or MuData here, right?

ehrapy/api/preprocessing/encoding/_encode.py
```diff
@@ -150,54 +148,61 @@ def _encode(
...
 
         encoded_x = None
         encoded_var_names = adata.var_names.to_list()
+        with Progress(
+            "[progress.description]{task.description}",
+            BarColumn(),
+            "[progress.percentage]{task.percentage:>3.0f}%",
+            refresh_per_second=1500,
+        ) as progress:
+            task = progress.add_task("[red] Running label encode on detected columns ...", total=1)
```
Zethson, line 164: Remove the space after [red]

ehrapy/api/preprocessing/encoding/_encode.py
```diff
@@ -150,54 +148,61 @@ def _encode(
...
+                layers={"original": updated_layer},
+            )
+            encoded_ann_data.uns["var_to_encoding"] = {
+                categorical: "label_encoding" for categorical in categoricals_names
+            }
+            encoded_ann_data.uns["encoding_to_var"] = {"label_encoding": categoricals_names}
 
     # user passed categorical values with encoding mode for each of them
     else:
+        # needed since, through side references, this would be deleted
```
Zethson, line 199: These commas are adventurous :)

ehrapy/api/preprocessing/encoding/_encode.py
```diff
@@ -306,8 +314,7 @@ def _one_hot_encoding(
         Encoded new X and the corresponding new var names
     """
     original_values = _initial_encoding(adata, categories)
-    if progress:
-        progress.update(task, description="[blue]Running one hot encoding encoding on passed columns ...")
+    progress.update(task, description="[blue]Running one hot encoding encoding on passed columns ...")
```
Zethson, line 317: encoding encoding

Zethson (approved):
Please also fix the download progressbar with the 1500 update tick.

PR #276, opened 2022-01-24
Fix #174 and #173

Imipenem:
**PR Checklist**

**Description of changes**

- fixed an issue that stored vars in uns["non_numerical_columns"] or uns["numerical_columns"] that where obs only

- fixed an issue that caused assert_encode to raise an error when an adata object has not been encoded yet (and does not need to be)


Zethson:
LGTM. Did you test the fix for the assert thingy quickly? Know it's annoying, but we need to get this right.

Imipenem:
Yeah sure, I tested it locally.

Issue #283, opened 2022-01-31
Increase performance of Miss Forest Imputation


Imipenem:
**Is your feature request related to a problem? Please describe.**

Miss forest imputation is a nice imputation algorithm with good results. But it comes at the cost of speed when imputing over multiple columns. This can be annoying especially on larger datasets.

**Describe the solution you would like**

> Depending on the size of the dataset and the number of features involved, iterative fitting may take a long time. Try using all your processors for the job by setting n_jobs=-1 for both RandomForestRegressor and RandomForestClassifier and you may try to patch Scikit-Learn with Intel(R) Extension for Scikit-learn as described here: pypi.org/project/scikit-learn-intelex

See https://stackoverflow.com/questions/64900801/implementing-knn-imputation-on-categorical-variables-in-an-sklearn-pipeline if this could be a possible solution and helps with speed.

Zethson:
Out settings object has a number to control the number of CPUs and therefore jobs here. This implementation has to make use of it if it doesn't already.

Zethson:
Multi-output and sparse data is not supported by the intel extension. Is this an issue?

PR #284, opened 2022-02-01
Improve missforest interface

### LINKED ISSUE #279: Nicer interface for miss_forest_impute
**Is your feature request related to a problem? Please describe.**

`var_names: dict[str, list[str]] | None = None` is the input parameter for miss_forest_impute. However, for all other imputation methods it is `var_names: list[str] | None = None`. 

**Describe the solution you would like**

Let's add support for only a list of strings also to miss_forest_impute. We will simply make the assumption that we are only imputing non-mixed data. Just wrap this with a try/except statement and if it crashes denote that a different parameter format is required for mixed data.

Zethson:
Fixes #279 and #283

@Imipenem The dict interface of MissForest is currently not tested, right?


ehrapy/api/preprocessing/_data_imputation.py
```diff
@@ -5,11 +5,12 @@
 from anndata import AnnData
 from rich import print
 from rich.progress import Progress, SpinnerColumn
-from sklearn.ensemble import ExtraTreesRegressor, RandomForestClassifier
 from sklearn.experimental import enable_iterative_imputer  # noqa: F401
-from sklearn.impute import IterativeImputer, KNNImputer, SimpleImputer
```
Imipenem, line 10: are those two removed?

Zethson: Locally imported after the Intel sklearn patch.

Imipenem:
General question: Does this work on other CPU architectures as well? Only asking because it's called `ìntel...`?

Zethson:
> General question: Does this work on other CPU architectures as well? Only asking because it's called `ìntel...`?

I hope so. I'll check again.

Zethson:
image

Should also support AMD.

Zethson:
Apparently a shitshow to install on M1 macs. I will add this as an optional dependency and only use the import when it's installed.

Issue #306, opened 2022-02-10
type_overview improvements


Zethson:
Really useful functions.

Couple of suggestions

1. move to anndata extensions
2. add parameter to sort the output by data type -> first floats, then booleans and then strings last or something
3. Variable names for AnnData object with 173 vars and 82 obs  - Variable names for AnnData object with 82 obs and 173 vars

Imipenem:
4. add an option to somehow display `uns` summary
5.  sort by encoding mode

PR #308, opened 2022-02-12
[FEATURE] Improve type_overview #306

Imipenem:
**PR Checklist**

**Description of changes**

- added new sorting modes

- moved code to anndata_ext

- refactored type_overview code


ehrapy/anndata_ext.py
```diff
@@ -6,6 +6,13 @@
 import numpy as np
 import pandas as pd
 from anndata import AnnData, concat
+from mudata import MuData
+from rich import print
+from rich.text import Text
+from rich.tree import Tree
+
+multi_encoding_modes = {"hash_encoding"}
+available_encodings = {"one_hot_encoding", "label_encoding", "count_encoding", *multi_encoding_modes}
```
Imipenem, line 15: this was required, otherwise there are circular import errors. Might need to move this into its own "encoding_util" file?

Zethson: Or scoped imports? This cannot be here.

Imipenem: Yeah, this might also help to solve the issue. Maybe we dont need the import at all. Will take a look at it.

ehrapy/anndata_ext.py
```diff
@@ -192,6 +196,177 @@ def get_column_values(adata: AnnData, indices: int | list[int]) -> np.ndarray:
...
+
+            adata = ep.dt.mimic_2(encode=True)
+            ep.anndata_ext.type_overview(adata)
+    """
+    if isinstance(data, AnnData):
+        _adata_type_overview(data, sort_by, sort_reversed)
+    elif isinstance(data, MuData):
+        _mudata_type_overview(data, sort_by, sort_reversed)
+    else:
+        print(f"[b red]Unable to present object of type {type(data)}. Can only display AnnData or MuData objects!")
```
Zethson, line 225: I totally get that the rich print looks nicer, but I feel like we should use actual errors whereever possible. They are used in the stacktrace and we're tuning them to be more readable with rich's stacktraceprint anyways.

Using a rich print here actually makes it harder to find the cause if the issue because it does not include the line where it occured etc.

ehrapy/anndata_ext.py
```diff
@@ -192,6 +196,177 @@ def get_column_values(adata: AnnData, indices: int | list[int]) -> np.ndarray:
...
+    data: MuData | AnnData, sort_by: str | None = None, sort_reversed: bool = False
+) -> None:  # pragma: no cover
+    """Prints the current state of an :class:`~anndata.AnnData` or :class:`~mudata.MuData` object in a tree format.
+    Output could be printed in sorted format by using one of `dtype`, `order`, `num_cats` or `None`, which sorts by data type, lexicographical order,
+    number of unique values (excluding NaN's) and unsorted respectively. Note that sorting by `num_cats` only affects
+    encoded variables currently and will display unencoded vars unsorted.
+
+    Args:
+        data: :class:`~anndata.AnnData` or :class:`~mudata.MuData` object to display
+        sort_by: How the tree output should be sorted. One of `dtype`, `order`, `num_cats` or `None` (defaults to None -> unsorted)
```
Zethson, line 209:
```suggestion
        sort_by: How the tree output should be sorted. One of `dtype`, `order`, `num_cats` or None (defaults to None -> unsorted)
```

Can be missleading to use the quotes here for the None?

Zethson (changes requested):
Great, thanks. Tiny request.

ehrapy/anndata_ext.py
Imipenem: True, guess its a remainder from the old days ;)

ehrapy/anndata_ext.py
Imipenem: good catch, yes

Issue #290, opened 2022-02-02
adata=ep.pp.encode(adata,autodetect=True) fails with index out of range for heart failure dataset


Zethson:
Heart Failure dataset from https://github.com/theislab/ehrapy-datasets/

```
adata = ep.dt.heart_failure()
adata=ep.pp.encode(adata,autodetect=True)
```

fails with
```

╭──────────────────────────── Traceback (most recent call last) ────────────────────────────╮
│                                                                                           │
│ /tmp/ipykernel_21248/1532827563.py:1 in <module>                                          │
│                                                                                           │
│ [Errno 2] No such file or directory: '/tmp/ipykernel_21248/1532827563.py'                 │
│ /home/zeth/PycharmProjects/ehrapy/ehrapy/api/preprocessing/encoding/_encode.py:64 in      │
│ encode                                                                                    │
│                                                                                           │
│    61 │   │   # basic type checking for passed parameters when encoding a single AnnData  │
│       object                                                                              │
│    62 │   │   if not _check_anndata_input_type(autodetect, encodings):                    │
│    63 │   │   │   raise EncodingInputValueError                                           │
│ ❱  64 │   │   return _encode(adata=data, autodetect=autodetect, encodings=encodings)      │
│    65 │   elif isinstance(data, MuData):                                                  │
│    66 │   │   # basic type checking for passed parameters when encoding a MuData object   │
│       (collection of multiple AnnData objects)                                            │
│    67 │   │   if not _check_mudata_input_type(autodetect, encodings):                     │
│                                                                                           │
│ /home/zeth/PycharmProjects/ehrapy/ehrapy/api/preprocessing/encoding/_encode.py:168 in     │
│ _encode                                                                                   │
│                                                                                           │
│   165 │   │   ) as progress:                                                              │
│   166 │   │   │   task = progress.add_task("[red]Running label encode on detected columns │
│       ...", total=1)                                                                      │
│   167 │   │   │   # Label encode by default. The primary usage of this is to save unencod │
│       AnnData objects                                                                     │
│ ❱ 168 │   │   │   encoded_x, encoded_var_names = _label_encoding(                         │
│   169 │   │   │   │   adata,                                                              │
│   170 │   │   │   │   encoded_x,                                                          │
│   171 │   │   │   │   encoded_var_names,                                                  │
│                                                                                           │
│ /home/zeth/PycharmProjects/ehrapy/ehrapy/api/preprocessing/encoding/_encode.py:373 in     │
[...]

Issue #287, opened 2022-02-02
Specify encoding mode when autodetect=True


Zethson:
**Is your feature request related to a problem? Please describe.**

`adata=ep.pp.encode(adata,autodetect=True)`

Currently simply label encodes all autodetected columns.

**Describe the solution you would like**

`adata=ep.pp.encode(adata,autodetect=True, encodings="one-hot")` which one-hot encodes all detected columns.

Imipenem:
This should be restricted tough to `one-hot, label` and `count`, since those won't need any additional information (for example `hash-encoding` would require the specification of buckets and since the encoded columns are unknown previously it won't be possible).

Zethson:
Agree. Need to throw appropriate errors if this expectation is not met.

PR #310, opened 2022-02-14
[FIX] Autodetect encode + specify encode mode for autodetect

Imipenem:
**PR Checklist**

**Description of changes**

- fixed an issue that caused ehrapy to crash when there are no columns that needs to be encoded (all numerical columns, #290)

- using autodetect while encoding now allows to pass a specific encoding mode used for the encoding (default is label encoding; works only for single column encoding modes like count, one-hot and label encoding #287 )

- added tests for the fix and feature


Imipenem:
@Zethson: Unfortunately the `type: ignore` are necessary since, we are indeed checking for type before to ensure we work with the correct data type but `mypy` does not recognize this.

tests/preprocessing/test_encode.py
```diff
@@ -42,11 +42,36 @@ def test_autodetect_encode(self):
         }
         assert id(encoded_ann_data.X) != id(encoded_ann_data.layers["original"])
 
+    def test_autodetect_num_only(self, capfd):
+        adata = read(dataset_path=f"{_TEST_PATH}/dataset2.csv")
+        encoded_ann_data = encode(adata, autodetect=True)
+        out, err = capfd.readouterr()
```
Zethson, line 48: I'd prefer a simple assert whether the error was thrown. No need to check the message. If we ever were to change the error message we would also have to change it here. Annoying.

Zethson (changes requested):
Minor changes again, thanks!

tests/preprocessing/test_encode.py
Imipenem: Would have done this normally, yes. The thing is, we do not throw an error here (we just print a warning).

Zethson: Ah okay.I get it. The message is ugly AF though :)

Should be more like:

"No columns to be encoded were detected" :)

Issue #239, opened 2022-01-10
Visualize QC metrics in a nice Rich table


Zethson:
Think we should add this as an option. The data fits rather nicely.

PR #323, opened 2022-02-26
[FEATURE] Display QC metrics of var #239

Imipenem:
**PR Checklist**

**Description of changes**

- display qc metrics of var with extra columns if wanted

- updated docs

@Zethson : are the QC metrics of var (the metric names) fixed or can they vary throughout different datasets? Cause currently the implementation just take every column in var and displays them, but that's not the desired behaviour!


Zethson:
"are the QC metrics of var (the metric names) fixed or can they vary throughout different datasets? Cause currently the implementation just take every column in var and displays them, but that's not the desired behaviour!"

Fixed so far, but maybe add a parameter to add additional columns for the display because people can add custom ones?

Imipenem:
> Fixed so far, but maybe add a parameter to add additional columns for the display because people can add custom ones?

Way to go for now, y

docs/usage/api.rst
```diff
@@ -62,6 +62,7 @@ Quality control
     :toctree: preprocessing
 
     preprocessing.calculate_qc_metrics
+    preprocessing.display_qc_metrics
```
Zethson, line 65: Should this be in the plotting part?

Maybe pp.qc_metrics and pl.qc_metrics?

Zethson (approved):
LGTM, but maybe we can move things around?

docs/usage/api.rst
Zethson: (I'd suggest to remove the "calculate" and "display" as well then.

e.g. umap is tl.umap and pl.umap

Must be reflected in the tutorials as well.

ehrapy/plot/ehrapy_plot/plot_qc.py
```diff
@@ -0,0 +1,51 @@
+from __future__ import annotations
+
+from anndata import AnnData
+from rich.console import Console
+from rich.table import Table
+
+
+def qc_metrics(adata: AnnData, extra_columns: list[str] | None = None) -> None:
```
Zethson, line 8: Think we need to `# pragma: no cover` this function?

Zethson (approved):
LGTM

ehrapy/plot/ehrapy_plot/plot_qc.py
Imipenem: Y, theres no proper way of testing this one.

Issue #314, opened 2022-02-15
Better support for 0 and 1 categoricals


Zethson:
Unfortunately, many datasets use 0 and 1 for booleans in columns. These must be autodetected as categoricals and also need to be e.g. one-hot encoded when requested.

What do we do when columns have 0, 1 and 2 for example? Are these already detected in the current implementation @Imipenem ?

Imipenem:
I guess  0 and 1 are common representations for booleans and might be detected as such by pandas. 


0,1 and 2 are counted as numericals, but that's an endless topic, like: What about 0,1,2,3 or 3-15 (some scores like GCS use this scala). 
We can try to detect such columns, but therefore we also provide the custom encodings besides autodetect.

Zethson:
> 0,1 and 2 are counted as numericals, but that's an endless topic, like: What about 0,1,2,3 or 3-15 (some scores like GCS use this scala). We can try to detect such columns, but therefore we also provide the custom encodings besides autodetect.

Totally get what you're saying. Then I guess this should still not be autodetected and users would have to custom encode this. But the column would be numerical, right? Does this work with out encoding currently?

Imipenem:
> But the column would be numerical, right? Does this work with out encoding currently?

The original column would be numerical (after encoding as well).

Zethson:
As discussed, we need an autodetection for these as well.

PR #327, opened 2022-02-28
[FEATURE] #314: Autodetect binary (0,1) columns

Imipenem:
**PR Checklist**

**Description of changes**

- autodetect now detects numerical columns that are binary
--> only contain 0 and 1 (or 0.0 and 1.0) and maybe some missing values

- columns count as binary when they have at least one 0 (or 0.0) and one 1 (or 1.0) and an arbitrary number (also 0) of missing values

### DIFF
--- a/ehrapy/anndata/anndata_ext.py
+++ b/ehrapy/anndata/anndata_ext.py
@@ -470,15 +470,14 @@ def _update_uns(
 
 
 def _detect_binary_columns(df: pd.DataFrame, numerical_columns: list[str]) -> list[str]:
-    """
-    Detect all columns that contain only 0 and 1 (besides NaNs)
+    """Detect all columns that contain only 0 and 1 (besides NaNs).
 
     Args:
-        df: The dataframe to check
-        numerical_columns: All numerical columns of df
+        df: The dataframe to check.
+        numerical_columns: All numerical columns of the dataframe.
 
     Returns:
-            List of column names that are binary (contiain only 0 and 1 (+NaNs))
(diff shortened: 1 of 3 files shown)


Zethson:
> columns count as binary when they have at least one 0 (or 0.0) and one 1 (or 1.0) and an arbitrary number (also 0) of missing values

at least one? Think this can lead to issues. Should be only 0 and 1 (0.0 and 1.0) right?

Imipenem:
> at least one? Think this can lead to issues. Should be only 0 and 1 (0.0 and 1.0) right?

So what I wanted to express is that the column should not have either only 1 (1.0) or 0 (0.0) so `myframe.mycolumn.nunique() == 2` (number of unique values except `NaN`).

Zethson (approved):
Good work! Do you expect a big performance penalty?

ehrapy/anndata/anndata_ext.py
```diff
@@ -467,6 +469,26 @@ def _update_uns(
         return all_moved_num_columns, list(all_moved_non_num_columns), None
 
 
+def _detect_binary_columns(df: pd.DataFrame, numerical_columns: list[str]) -> list[str]:
+    """
+    Detect all columns that contain only 0 and 1 (besides NaNs)
```
Zethson, line 474: BTW, the first docstring sentence has to be after the triple quotes and must end with a `.`

Imipenem:
> Good work! Do you expect a big performance penalty?

Don't think so, but will test it with a quick profile ;)

Issue #255, opened 2022-01-18
encode mutates input adata


ivirshup:
**Describe the bug**

`ep.pp.encode` mutates it's input `AnnData`

```python
adatas = ep.dt.mimic_3_demo(encoded=False, mudata=False)
adata = adatas["INPUTEVENTS_CV"]
display(adata)
```

```
AnnData object with n_obs × n_vars = 34799 × 20
    obs: 'charttime', 'storetime'
    layers: 'original'
```

```python
adata_encoded = ep.pp.encode(adata, autodetect=True)
display(adata)
```

```
AnnData object with n_obs × n_vars = 34799 × 20
    uns: 'categoricals', 'original_values_categoricals'
    layers: 'original'
```

**Expected behavior**

I didn't expect the input to be mutated if the function returns an encoded output. I think it's unclear how to interpret what happened to it.

This could be intentional, but if so I think it should be documented.

**System [please complete the following information]:**

<details>
<summary> System info </summary>

```
-----
IPython             8.0.0
ehrapy              0.1.0
session_info        1.0.0
-----
PIL                 9.0.0
PyPDF2              1.26.0
anndata             0.7.8
asttokens           NA
attr                21.4.0
backcall            0.2.0
backports           NA
beta_ufunc          NA
binom_ufunc         NA
black               21.12b0
blib2to3            NA
blis                NA
bs4                 4.10.0
camelot             0.10.1
catalogue           NA
category_encoders   2.3.0
certifi             2021.10.08
cffi                1.15.0
chardet             4.0.0
charset_normalizer  2.0.10
click               8.0.3
colorama            0.4.4
cryptography        36.0.1
cv2                 4.5.5
cycler              0.10.0
cymem               NA
cython_runtime      NA
dateutil            2.8.2
decorator           5.1.1
deep_translator     1.6.1
deepl               1.3.1
dill                0.3.4
executing           0.8.2
filelock            3.4.2
gensim              4.1.2
h5py                3.6.0
html5lib            1.1
huggingface_hub     0.0.19
idna                3.3
igraph              0.9.9
jedi                0.18.1
jinja2              3.0.3
joblib              1.1.0
jsonpickle          2.0.0
kiwisolver          1.3.2
leidenalg           0.8.8
llvmlite            0.38.0
markupsafe          2.0.1
matplotlib          3.5.1
medcat              NA
mpl_toolkits        NA
mudata              0.1.1
multiprocess        0.70.12.2
murmurhash          NA
[...]

PR #348, opened 2022-03-10
[FIX] #255: Encode mutates input adata object

Imipenem:
**PR Checklist**

**Description of changes**

- encode does not mutate input adata object anymore

- no copy for the whole adata is needed with this implementation

- fixed an issue that caused non numerical columns to be stored as `str` in `uns` when encoding them

### DIFF
--- a/ehrapy/anndata/anndata_ext.py
+++ b/ehrapy/anndata/anndata_ext.py
@@ -474,8 +474,8 @@ def _update_uns(
     2.) Moving `col1` from `obs` to `X`: `col1` is either numerical or non_numerical, so add it to the corresponding entry in `uns`
     """
     moved_columns_set = set(moved_columns)
-    num_set = set(adata.uns["numerical_columns"])
-    non_num_set = set(adata.uns["non_numerical_columns"])
+    num_set = set(adata.uns["numerical_columns"].copy())
+    non_num_set = set(adata.uns["non_numerical_columns"].copy())
     if not to_x:
         var_num = []
         for var in moved_columns:
--- a/ehrapy/io/_write.py
+++ b/ehrapy/io/_write.py
@@ -60,7 +60,7 @@ def write(
         # dummy encoding when there is non numerical data in X
         if not np.issubdtype(adata.X.dtype, np.number) and extension == "h5ad":
             # flag to indicate an Anndata object has been dummy encoded to write it to .h5ad file
-            # this is when writing an unencoded non numerical AnnData object
+            # Case of writing an unencoded non numerical AnnData object
             encoded_adata = encode(adata, autodetect=True)
             encoded_adata.uns["ehrapy_dummy_encoding"] = True
             encoded_adata.uns["columns_obs_only"] = list(adata.obs.columns)
--- a/ehrapy/preprocessing/encoding/_encode.py
+++ b/ehrapy/preprocessing/encoding/_encode.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from itertools import chain
-from typing import Dict, List
+from typing import Dict, List, Any
 
 import numpy as np
 import pandas as pd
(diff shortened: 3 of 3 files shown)


ehrapy/io/_write.py
```diff
@@ -60,13 +60,11 @@ def write(
...
         if not np.issubdtype(adata.X.dtype, np.number) and extension == "h5ad":
             # flag to indicate an Anndata object has been dummy encoded to write it to .h5ad file
-            # this could be the case when writing to cache file or when writing an unencoded non numerical AnnData object
-            adata_cp = adata.copy()
-            adata_cp.uns["ehrapy_dummy_encoding"] = True
-            adata_cp.uns["columns_obs_only"] = list(adata_cp.obs.columns)
-            # TODO: THIS SHOULD BE FIXED WITH PR #348, SO NO COPY SHOULD BE NEEDED THEN SINCE THE ORIGINAL WILL NOT BE MUTATED
-            encoded_adata_cp = encode(adata_cp, autodetect=True)
-            encoded_adata_cp.write(filename, compression=compression, compression_opts=compression_opts)
+            # this is when writing an unencoded non numerical AnnData object
```
Zethson, line 63:
```suggestion
            # Case of writing an unencoded non numerical AnnData object
```

ehrapy/anndata/anndata_ext.py
```diff
@@ -482,8 +482,8 @@ def _update_uns(
     2.) Moving `col1` from `obs` to `X`: `col1` is either numerical or non_numerical, so add it to the corresponding entry in `uns`
     """
     moved_columns_set = set(moved_columns)
-    num_set = set(adata.uns["numerical_columns"])
-    non_num_set = set(adata.uns["non_numerical_columns"])
+    num_set = set(adata.uns["numerical_columns"].copy())
```
Zethson, line 485: Later down we are copying the whole `.uns`. Are both copies required?

ehrapy/preprocessing/encoding/_encode.py
```diff
@@ -51,7 +51,8 @@ def encode(
         A dict containing the encoding mode and categorical name for the respective column (for each AnnData object in case of MuData object).
 
     Returns:
-        An :class:`~anndata.AnnData` object with the encoded values in X or None (in case of :class:`~mudata.MuData` object)
+        An :class:`~anndata.AnnData` object with the encoded values in X or None (in case of :class:`~mudata.MuData` object). The :class:`~anndata.AnnData`
```
Zethson, line 54: Don't think that this additional comment is required. This is the expected behavior I guess? Might be more confusing than helpful.

ehrapy/preprocessing/encoding/_encode.py
```diff
@@ -832,44 +843,46 @@ def _get_categoricals_old_indices(old_var_names: list[str], encoded_categories:
     return idx_list
 
 
-def _add_categoricals_to_obs(ann_data: AnnData, categorical_names: list[str]) -> None:
+def _add_categoricals_to_obs(from_: AnnData, to: AnnData, categorical_names: list[str]) -> None:
```
Zethson, line 846: Why does from have an underscore at the end and to does not?
The naming is a little bit weird...Can we make this more explicit?

Zethson (commented):
Minor things and comments. Good stuff!

ehrapy/preprocessing/encoding/_encode.py
Imipenem: from is a reserved keyword in python so the underscore was needed. Maybe add it to `to` as well?

ehrapy/anndata/anndata_ext.py
Imipenem: Yes, because were mutating the dicts here (updating them) so we need those copies. Otherwise it would mutate the original lists/sets as well.

Issue #263, opened 2022-01-20
Remove master read function and separate the file formats more clearly


Zethson:
Think this is required because the master read function is hard to document and we need to document some format specific details. Currently we only render the master read function on our docs.

PR #351, opened 2022-03-13
[Feature] New read API #263

Imipenem:
**PR Checklist**

**Description of changes**

- added a new read API for ehrapy.io

- each file type now has its own read function to separate it more clearly and remove unnecessary boilerplate code

- added read support for multiple h5ad files (for example when reading from cache directory)

- reading from cache directory is now only supported via the read_h5ad function

- TODO: add many tests (important here), fix tests, update docs, add custom exceptions

### DIFF
--- a/ehrapy/io/_read.py
+++ b/ehrapy/io/_read.py
@@ -9,21 +9,19 @@ import numpy as np
 import pandas as pd
 from _collections import OrderedDict
 from anndata import AnnData
-from anndata import read as read_h5ad
+from anndata import read as read_h5
 from mudata import MuData
 from rich import print
 
 from ehrapy import ehrapy_settings, settings
 from ehrapy.anndata.anndata_ext import df_to_anndata
 from ehrapy.data._dataloader import download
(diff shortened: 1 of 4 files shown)


ehrapy/io/_read.py
```diff
@@ -56,7 +43,45 @@ def read(
...
+            import ehrapy as ep
+            adata = ep.io.read_cvs("myfile.csv")
+    """
+    file: Path = Path(dataset_path)
+    if not file.exists():
+        file = _get_non_existing_files(file, download_dataset_name, backup_url)
+
+    adata = _read_csv(
+        filename=file,
+        extension=extension,
```
Zethson, line 61: Maybe we can remove the "extension" parameter here now and use the Pandas parameter `sep`? More consistent API with the big packages?

Zethson (approved):
Would be great if we could rename the dataset names to more descriptive names.

"dataset_6" is not really great. One would have to open all of them when looking for a specific dataset to use for a new test.

Zethson:
What's going on with the test failures?

PR #352, opened 2022-03-13
Add ability to compare laboratory measurements to reference values

### LINKED ISSUE #338: Control features with norm values
Would be good to add to QC:

1. Control features (usually lab measurements) with norm ranges and mark how many and which are not in the norm ranges.
2. Maybe also by how much you can be out of bounds of the norm ranges?

Zethson:

fixes #338


Zethson:
I will add the missing tests in a couple of weeks. The ones we have cover the main part.

Issue #307, opened 2022-02-11
KNN imputation improvements


Zethson:
1. Ensure that value K is accessible for numerical and ideally also categorical values
2. Try to find some strategy to impute categorical values with e.g. most frequent -> explore scanpy KNN (neighbors) for custom implementation if required. Likely will involve label encoding the values, imputing and then mapping back.
3. Google around for KNN imputation improvements

PR #365, opened 2022-03-29
[#307] Improve KNN with n_neighbours parameter

Imipenem:
**PR Checklist**

**Description of changes**

- added n_neighbours parameter to knn impute

- encoding now only gets applied to non numerical columns

- added tests for knn impute with n_neighbours parameter

### DIFF
--- a/ehrapy/preprocessing/_data_imputation.py
+++ b/ehrapy/preprocessing/_data_imputation.py
@@ -248,7 +248,9 @@ def knn_impute(
             _knn_impute(adata, var_names, n_neighbours)
             # imputing on encoded columns might result in float numbers; those can not be decoded
             # cast them to int to ensure they can be decoded
-            adata.X[::, column_indices] = adata.X[::, column_indices].astype(int)
+            adata.X[::, column_indices] = np.rint(adata.X[::, column_indices]).astype(int)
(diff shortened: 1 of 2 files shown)


Imipenem:
@Zethson Note: The current implementation just cuts off the float part of the encoded values of non numerical columns and use this for the int value. There is a `numpy` function that does a more natural behaviour (actual rounding: 2.4 -> 2, 2.6 -> 3). But somehow, this didn't work on `float` input with some strange errors (at least for me).

Zethson:
1. What happens in the case of categorical values?
2. What's the error with the floats? Totally not happy with the int casting. For some blood measurements or whatever this is kinda unacceptable, no?

Imipenem:
> 1. What happens in the case of categorical values?
> 
>     2. What's the error with the floats? Totally not happy with the int casting. For some blood measurements or whatever this is kinda unacceptable, no?

Its fixed, but it kind of shows the downside when using numpy array with dtype object. Many of the `ufuncs` of numpy are not defined on object dtype arrays (like `isnan` or `rint` or `round`).

Zethson:
So there's still some casting for ints left. Why is that? Just trying to understand.

Zethson:
Yup, I understood it correctly. Thanks. Approved

Issue #321, opened 2022-02-24
Add X_only option


Zethson:
@HorlavaNastassya encountered a dataset where she wants to much rather choose which data to add to X only and everything else to obs by default.

Besides columns_obs_only we could also add columns_X_only

PR #380, opened 2022-05-16
#321: Add X_only option for reading

Imipenem:
**PR Checklist**

**Description of changes**

- see #321 
- added an option to read with columns that should go to X only and everything else to `.obs`
- datetime columns will be added to `.obs` regardless.

### DIFF
--- a/ehrapy/io/_read.py
+++ b/ehrapy/io/_read.py
@@ -16,6 +16,7 @@ from rich import print
 from ehrapy import ehrapy_settings, settings
 from ehrapy.anndata.anndata_ext import df_to_anndata
 from ehrapy.data._dataloader import download
+from ehrapy.io._utility_io import _check_columns_only_params
 from ehrapy.preprocessing.encoding._encode import encode
 
 
@@ -37,10 +38,10 @@ def read_csv(
         dataset_path: Path to the file or directory to read.
         sep: Separator in the file; either , (default) or \t.
         index_column: The index column of obs. Usually the patient visit ID or the patient ID.
-        columns_obs_only: These columns will be added to obs only and not X. Can not be used together with columns_x_only.
-        columns_x_only: These columns will be added to X only and all remaining columns to obs. Can not be used together with columns_obs_only.
+        columns_obs_only: These columns will be added to obs only and not X.
+        columns_x_only: These columns will be added to X only and all remaining columns to obs. Note that datetime columns will always be added to .obs though.
         return_mudata: Whether to create and return a MuData object. This is primarily used for complex datasets which require several AnnData files.
-        cache: Whether to use the cache when reading.
+        cache: Whether to write to cache when reading or not.
         download_dataset_name: Name of the file or directory in case the dataset is downloaded
         backup_url: URL to download the data file(s) from if not yet existing.
 
@@ -54,8 +55,7 @@ def read_csv(
             import ehrapy as ep
             adata = ep.io.read_csv("myfile.csv")
     """
-    if columns_obs_only and columns_x_only:
-        raise ValueError("Can not use columns_obs_only together with columns_x_only. At least one has to be None!")
+    _check_columns_only_params(columns_obs_only, columns_x_only)
     file: Path = Path(dataset_path)
     if not file.exists():
         file = _get_non_existing_files(file, download_dataset_name, backup_url)
@@ -65,6 +65,7 @@ def read_csv(
         sep=sep,
         index_column=index_column,
         columns_obs_only=columns_obs_only,
+        columns_x_only=columns_x_only,
         return_mudata=return_mudata,
         cache=cache,
         **kwargs,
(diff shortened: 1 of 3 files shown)


ehrapy/io/_read.py
```diff
@@ -37,8 +39,9 @@ def read_csv(
         sep: Separator in the file; either , (default) or \t.
         index_column: The index column of obs. Usually the patient visit ID or the patient ID.
         columns_obs_only: These columns will be added to obs only and not X.
+        columns_x_only: These columns will be added to X only and all remaining columns to obs. Note that datetime columns will always be added to .obs though.
         return_mudata: Whether to create and return a MuData object. This is primarily used for complex datasets which require several AnnData files.
-        cache: Whether to use the cache when reading.
+        cache: Whether to write to cache when reading or not.
```
Zethson, line 44: Could add the default here -> 
```suggestion
        cache: Whether to write to cache when reading or not. (default: False)
```

ehrapy/io/_utility_io.py
```diff
@@ -28,3 +28,31 @@ def _get_file_extension(file_path: Path) -> str:
         {supported_extensions}
         """
     )
+
+
+def _check_columns_only_params(
+    obs_only: dict[str, list[str]] | list[str] | None, x_only: dict[str, list[str]] | list[str] | None
+) -> None:
+    """Check whether columns_obs_only and columns_x_only can be used as passed. For a single anndata object (thus
```
Zethson, line 36: Please adhere to the proper Google Docstring style.

Zethson: "can be used as passed"? This doesn't read like proper English.

Zethson: AnnData object, not anndata. The package is called anndata :)

ehrapy/io/_utility_io.py
```diff
@@ -28,3 +28,31 @@ def _get_file_extension(file_path: Path) -> str:
...
+        return
+    # cannot use both for a single anndata object
+    if obs_only and x_only and isinstance(obs_only, list):
+        raise ValueError(
+            "Can not use columns_obs_only together with columns_x_only with a single AnnData object. At least one has to be None!"
+        )
+    # check for duplicates in the two dicts for multiple AnnData objects
+    else:
+        common_keys = obs_only.keys() & x_only.keys()  # type: ignore
+        # at least one duplicated key has been found
```
Zethson, line 53: Useless comment.

ehrapy/io/_utility_io.py
```diff
@@ -28,3 +28,31 @@ def _get_file_extension(file_path: Path) -> str:
...
+def _check_columns_only_params(
+    obs_only: dict[str, list[str]] | list[str] | None, x_only: dict[str, list[str]] | list[str] | None
+) -> None:
+    """Check whether columns_obs_only and columns_x_only can be used as passed. For a single anndata object (thus
+    parameters being a list of strings) it's not possible to pass both, obs_only and x_only.
+    For multiple anndata objects (thus the parameters being dicts of string keys with a list value), it is possible to pass both. But the keys
+    (unique identifiers of the anndata objects, basically its names) should share no common identifier, thus a single anndata object is either in x_only OR
+    obs_only, but not in both.
+    """
+    # at least one parameter is None
```
Zethson, line 42: Well, that's what the code says. Useless comment.
Rather write something like: "At least one of A and B is None. Simply return because X Y Z"

ehrapy/io/_utility_io.py
```diff
@@ -28,3 +28,31 @@ def _get_file_extension(file_path: Path) -> str:
...
+    if not obs_only or not x_only:
+        return
+    # cannot use both for a single anndata object
```
Zethson, line 45: Useless comment. The ValueError states exactly that.

Zethson (approved):
Not sure why it is not possible to use obs_only and X_only at the same time as long as their sets are disjoint?

Imipenem:
> Not sure why it is not possible to use obs_only and X_only at the same time as long as their sets are disjoint?

As discussed: When the two sets are disjoint, there is no advantage of passing both parameters at the same time: Consider `obs_only = [1,2,3]` and `x_only=[4,5,6]`. If both are passed `1,2,3` goes to `.obs` and `4,5,6` to `X`. Which would also be the case when only passing one of them. 
Maybe raising an error here is a bit "harsh", but it could raise awareness here; so less to do for the user (and its smoother to implement as well).

Zethson:
> > Not sure why it is not possible to use obs_only and X_only at the same time as long as their sets are disjoint?
> 
> As discussed: When the two sets are disjoint, there is no advantage of passing both parameters at the same time: Consider `obs_only = [1,2,3]` and `x_only=[4,5,6]`. If both are passed `1,2,3` goes to `.obs` and `4,5,6` to `X`. Which would also be the case when only passing one of them. Maybe raising an error here is a bit "harsh", but it could raise awareness here; so less to do for the user (and its smoother to implement as well).

You are correct. The heat is getting into my mind already ^_^

Issue #397, opened 2022-07-18
Check for neighbors run before UMAP


Zethson:
`ValueError: Did not find .uns["neighbors"]. Run `sc.pp.neighbors` first.`
When attempting to run `ep.tl.umap` before `ep.pp.neighbors`.

We should catch this issue ourselves and refer to `ep.pp.neighbors`

PR #401, opened 2022-08-11
issue #397 "check for neighbors run before UMAP" fixed

namsaraeva:
**Description of changes**

Added some code to check for neighbors run before running UMAP.


Zethson (approved):
Well done!

PR #404, opened 2022-08-21
Add copy_obs parameter to move_to_obs

namsaraeva:
**Description of changes**

Solved the issue of the existing `move_to_obs` function. I added a `boolean` parameter `copy_obs` that copies the column from X to obs and does not delete it from X. 

**Technical details**

Added a simple test that checks the shape of X before and after copying to make sure nothing got deleted or moved. Also checked if the columns are present in obs.

**Additional context**


ehrapy/anndata/anndata_ext.py
```diff
@@ -176,17 +178,30 @@ def move_to_obs(adata: AnnData, to_obs: list[str] | str, copy: bool = False) ->
...
-    # cast non numerical values from object to either bool (if possible) or category
-    adata.obs = _cast_obs_columns(adata.obs)
-    adata.uns["numerical_columns"] = updated_num_uns
-    adata.uns["non_numerical_columns"] = updated_non_num_uns
+
+    if copy_obs:
+        indices = adata.var_names.isin(to_obs)
+        df = adata[:, indices].to_df()
+        adata.obs = adata.obs.join(df)
+        updated_num_uns, updated_non_num_uns, num_var = _update_uns(adata, to_obs)
```
Imipenem, line 186: By copying columns from X to obs, the columns still remain in X. So they should not got deleted from uns numerical/non numerical (which is what `update_uns` is basically doing). 
This also requires changes to `move_to_x`, which should check whether the columns are already present in X (in that case simply delete those columns from obs).

Imipenem (changes requested):

Imipenem:
One minor change, otherwise LGTM.

ehrapy/anndata/anndata_ext.py
namsaraeva: So I just have to delete the lines below ```update_uns```?

ehrapy/anndata/anndata_ext.py
```diff
@@ -204,13 +212,17 @@ def move_to_x(adata: AnnData, to_x: list[str] | str) -> AnnData:
...
     if isinstance(to_x, str):  # pragma: no cover
         to_x = [to_x]
-    new_adata = concat([adata, AnnData(adata.obs[to_x], dtype="object")], axis=1)
-    new_adata.obs = adata.obs[adata.obs.columns[~adata.obs.columns.isin(to_x)]]
-    # update uns (copy maybe: could be a costly operation but reduces reference cycles)
-    # users might save those as separate AnnData object and this could be unexpected behaviour if we dont copy
-    num_columns_moved, non_num_columns_moved, _ = _update_uns(adata, to_x, True)
-    new_adata.uns["numerical_columns"] = adata.uns["numerical_columns"] + num_columns_moved
-    new_adata.uns["non_numerical_columns"] = adata.uns["non_numerical_columns"] + non_num_columns_moved
+    if set(to_x).issubset(set(adata.var_names)):
```
Imipenem, line 215: This only works while all items of `to_x` are in the `adata.var_names`. We need to also make this work in case only some of the columns in `to_x` are present in `adata.var_names` (and therefore copied before). So I guess we have to check for each column in `to_x` independently, whether it is in `var_names` or not.

And if not: they have to be added to `uns` as well (besides deleting them from `obs`) in the corresponding lists of numerical/non_numerical columns.

Would be cool, if you could add some tests as well for such cases.

Imipenem: rest LGTM

Imipenem (changes requested):

ehrapy/anndata/anndata_ext.py
namsaraeva: Thank you!

namsaraeva:
Could you please have a look? @Imipenem

ehrapy/anndata/anndata_ext.py
```diff
@@ -176,42 +178,74 @@ def move_to_obs(adata: AnnData, to_obs: list[str] | str, copy: bool = False) ->
...
+        adata.uns["numerical_columns"] = updated_num_uns
+        adata.uns["non_numerical_columns"] = updated_non_num_uns
 
     if copy:
         return adata
 
 
 def move_to_x(adata: AnnData, to_x: list[str] | str) -> AnnData:
     """Move features from obs to X inplace.
-
```
Zethson, line 197: We should keep this empty line.

ehrapy/anndata/anndata_ext.py
```diff
@@ -176,42 +178,74 @@ def move_to_obs(adata: AnnData, to_obs: list[str] | str, copy: bool = False) ->
...
     Args:
         adata: The AnnData object
         to_x: The columns to move to X
-
```
Zethson, line 201: We should keep this one as well.

Imipenem (commented):
Code LGTM, good work!

I'd request to add a few more tests (as I think of it now, I did not covered all edge cases in the tests): Maybe one test for columns that do not exist in X or obs when moved/copied -> should raise an exception and explain error). Moving columns from X to obs and back; especially an edge case when one wants to move some columns from obs to X and some of those columns where copied previously from X to obs.

If those tests pass, ready to merge IMO.

namsaraeva:
1) So, for now I added some tests for cases when columns that do not exist in X or obs are moved/copied (i.e. raised the **exceptions in the methods**). 

2) "Test for moving columns from `X` to `obs` and back; especially an edge case when one wants to move some columns from `obs` to `X` and some of those columns where copied previously from `X` to `obs`." -> **Not fully done yet.** In the method move_to_x() I distinguish between **three** different edge cases: 

- Columns passed exist in both `obs` and `X` because they were previously just copied from `X` to `obs`. The only thing to do here is we need to remove this column from `obs` (do we necessarily want it though?). 
- Columns passed exist **only** in `obs`: we want to append the column to `X`, remove the column from `obs` and update `uns`. **This is where the things go wrong.**
- Columns passed are mixed: some are already in `X`, some are not. I divide the list into two separate lists with columns that exist and columns that don't exist in X. Then I handle them like in two cases above.

**So, why do things go wrong?**

The answer is in the function _update_uns() and its usage in the function move_to_x(). There we want to move columns from obs to X and update `uns` accordingly. For this purpose, functions copies numerical and non-numerical values from `uns` of the AnnData passed to the function, and tries to categorise the columns that are being moved (if they are in the num_set, then we append it to new numerical columns, otherwise to non-numerical). But what happens if the column that we want to move from `obs` to `X` does not exist in `X` anymore and therefore *not* in `uns`? We can't categorise the column if it's numerical or not, because if it does not exist  in `X`, it does not exist in `uns` either. That is what we essentially do when using the method move_to_obs(): if we don't want to have one column in `X`, we move it to `obs` inplace and delete it from `uns`.  

**How to fix it?**

We probably need to add one functionality more to the _update_uns() function where it checks the dtype of the passed column and categorises it as numerical or non-numerical and adds it to uns as a new entry. At least I would do that. 

I hope you will understand what I mean. Would be eager to discuss it with you, @Imipenem

Imipenem:
👍 nice work, im currently on vacation, but will get back to you Friday.

namsaraeva:
> You are absolutely right; ok, what about just "do nothing" but warn the user: Column XY already in X. Skipped moving XY to X? As a nice add-on, we could also provide a simple function to delete columns entirely from obs (and every reference to them); might be useful, what do you think?

I added a new function for this: `delete_from_obs()`, a very simple func that deletes the columns inplace, also added a small test for it. In the `move_to_x()`function now we warn the user that moving was skipped and if the user wants to delete the column, she/he should use the `delete_from_obs()`function. But I was thinking if we could make it more interactive, like with Yes/No/Skip on console, but I am not sure if we want to do this here. Might just stick with the small warning. 

Also `move_to_X()` should work fine now, my test with three edge cases passes successfully now, feel free to check it, maybe I forgot something. 

But I have a question for  `move_to_X()`. Currently we return a copy of AnnData, so to see the changes, user needs to assign the output of `move_to_X()` to a new variable/AnnData object. Do we want something like `copy=True/False` like in `move_to_obs()` so the user can choose whether to return a copy or act inplace. 

What do you think? @Imipenem

Zethson:
> But I was thinking if we could make it more interactive, like with Yes/No/Skip on console, but I am not sure if we want to do this here. Might just stick with the small warning.

I would not make it interactive, no.


> But I have a question for move_to_X(). Currently we return a copy of AnnData, so to see the changes, user needs to assign the output of move_to_X() to a new variable/AnnData object. Do we want something like copy=True/False like in move_to_obs() so the user can choose whether to return a copy or act inplace.

Yeah, I think that this would be a good idea.

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,135 @@ def anndata_to_df(
...
+def move_to_obs(adata: AnnData, to_obs: list[str] | str, copy_obs: bool = False, copy: bool = False) -> AnnData:
+    """Move inplace or copy features from X to obs. Note that columns containing boolean values (either 0/1 or True(
+    true)/False(false)) will be stored as boolean columns whereas the other non numerical columns will be stored as
+    category.
 
     Args:
         adata: The AnnData object
         to_obs: The columns to move to obs
+        copy_obs: The values are copied to obs (and therefore kept in X) instead of moved completely
         copy: Whether to return a copy or not
```
Imipenem, line 166: Maybe replace with something like: `copy: Whether to return a copy of the original AnnData object or not.`. So it's clear, we're talking about `AnnData` objects here; much copying going on here so might be useful.

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,135 @@ def anndata_to_df(
...
-    # cast numerical values from object
-    adata.obs[num_var] = adata.obs[num_var].apply(pd.to_numeric, errors="ignore", downcast="float")
-    # cast non numerical values from object to either bool (if possible) or category
-    adata.obs = _cast_obs_columns(adata.obs)
-    adata.uns["numerical_columns"] = updated_num_uns
-    adata.uns["non_numerical_columns"] = updated_non_num_uns
+
+    if not set(to_obs).issubset(set(adata.var_names)):
+        raise ValueError(
+            "Cannot move columns with invalid names."
```
Imipenem, line 183: This error message is a bit general. One could be confused by what exactly makes a name "invalid". Maybe change this to sth. like "Columns XYZ (simple diff of the two sets) are not in var_names".

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,135 @@ def anndata_to_df(
...
+        # cast non numerical values from object to either bool (if possible) or category
+        adata.obs = _cast_obs_columns(adata.obs)
+        adata.uns["numerical_columns"] = updated_num_uns
+        adata.uns["non_numerical_columns"] = updated_non_num_uns
 
     if copy:
         return adata
 
 
+def delete_from_obs(adata: AnnData, to_delete: list[str] | str) -> AnnData:
```
Imipenem, line 226: Add `copy` parameter as well?

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,135 @@ def anndata_to_df(
...
+    """Delete previously copied features from obs.
```
Imipenem, line 227:
```suggestion
    """Delete features from obs.
```

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,135 @@ def anndata_to_df(
...
+    Args:
+        adata: The AnnData object
+        to_delete: The columns to delete from obs
+    Returns:
+        AnnData object with deleted columns from obs.
+    """
+
+    if isinstance(to_delete, str):  # pragma: no cover
+        to_delete = [to_delete]
```
Imipenem, line 236: why not use a set instead? You are already using subset below and `isin` works with sets as well.

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,135 @@ def anndata_to_df(
...
+
+    if not set(to_delete).issubset(set(adata.obs.columns)):
+        raise ValueError(
+            "Cannot delete columns with invalid names."
```
Imipenem, line 240: See comment above.

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,135 @@ def anndata_to_df(
...
-    new_adata.obs = adata.obs[adata.obs.columns[~adata.obs.columns.isin(to_x)]]
-    # update uns (copy maybe: could be a costly operation but reduces reference cycles)
-    # users might save those as separate AnnData object and this could be unexpected behaviour if we dont copy
-    num_columns_moved, non_num_columns_moved, _ = _update_uns(adata, to_x, True)
-    new_adata.uns["numerical_columns"] = adata.uns["numerical_columns"] + num_columns_moved
-    new_adata.uns["non_numerical_columns"] = adata.uns["non_numerical_columns"] + non_num_columns_moved
+
+    if not set(to_x).issubset(set(adata.obs.columns)):
+        raise ValueError(
+            "Cannot move columns with invalid names."
```
Imipenem, line 260: see above as well

Imipenem (commented):
Some minor things; but afterwards lets merge ;)

namsaraeva:
Thank you for the review @Imipenem, I incorporated those changes.

I tried to add the `copy` parameter to `move_to_x()` function, but apparently AnnData's `concat()`function always has to return a new object to avoid corrupting the AnnData object (talked to Isaac), so it was not possible for the time being. This is very niche and maybe we can try it once again sometime in the future, but I personally think it is fine for now. Now the user has to overwrite the object to see the changes that were made by `move_to_x()` i.e. `adata = move_to_x(adata, ['blabla'])`. 

For API consistency purposes I had to delete the `copy` parameter from `move_to_obs()` and `delete_from_obs()`as well, so they always return the original AnnData object now.

Also fixed the tests and formatting. I think we are ready to merge.

Imipenem:
LGTM 🚀.

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,130 @@ def anndata_to_df(
...
+        num_set = set(adata.uns["numerical_columns"].copy())
+        non_num_set = set(adata.uns["non_numerical_columns"].copy())
+        var_num = []
+        var_non_num = []
+        for var in to_obs:
+            if var in num_set:
+                var_num.append(var)
+            elif var in non_num_set:
+                var_non_num.append(var)
+        # cast numerical values from object
```
Zethson, line 196: Usually always write WHY you're doing things. -> "Cast numerical value from object to XYZ"

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,130 @@ def anndata_to_df(
...
+        adata.obs[var_num] = adata.obs[var_num].apply(pd.to_numeric, errors="ignore", downcast="float")
+        # cast non numerical values from object to either bool (if possible) or category
```
Zethson, line 198: Why? Else, the comment is more or less useless because the code line actually says what it does

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,130 @@ def anndata_to_df(
...
+        adata.obs = _cast_obs_columns(adata.obs)
+    else:
+        indices = adata.var_names.isin(to_obs)
+        df = adata[:, indices].to_df()
+        adata._inplace_subset_var(~indices)
+        adata.obs = adata.obs.join(df)
+        updated_num_uns, updated_non_num_uns, num_var = _update_uns(adata, to_obs)
+        # cast numerical values from object
```
Zethson, line 206: See above.

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,130 @@ def anndata_to_df(
...
 
     Args:
         adata: The AnnData object
         to_x: The columns to move to X
+        copy: Whether to return a copy or not
 
     Returns:
         A new AnnData object with moved columns from obs to X. This should not be used for datetime columns currently.
     """
+
```
Zethson, line 249:
```suggestion
```

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,130 @@ def anndata_to_df(
...
+    if not all(elem in adata.obs.columns.values for elem in to_x):
+        raise ValueError(f"Columns `{[x for x in to_x if x not in adata.obs.columns.values]}` are not in obs.")
```
Zethson, line 251: Variable name

ehrapy/anndata/anndata_ext.py
```diff
@@ -154,63 +154,130 @@ def anndata_to_df(
...
+    cols_present_in_x = []
+    cols_not_in_x = []
+
+    for col in to_x:
+        if col in set(adata.var_names):
+            cols_present_in_x.append(col)
+        else:
+            cols_not_in_x.append(col)
+
+    if cols_present_in_x:  # if there are some columns that were already in X, we skip the moving
```
Zethson, line 265: The comment should not explain the "what", but the "why". It might not even be necessary at all.

Zethson (changes requested):
Aaaaaaaaaaaaaalmost there. Just super minor things. Feel free to merge after you've addressed those!

PR #418, opened 2022-11-11
Add winsorize, clip quantiles and filter quantiles

### LINKED ISSUE #335: Add quantile filtering functions
Would be cool to e.g. filter features by quantiles. -> Remove the top 1% largest values or whatever. This is not difficult to code but it might be nice to wrap this into easier filter functions?
Maybe also offer this with absolute values?

So basically something like:

# Removes the top 1 % of all values in the feature 1 and feature 2 columns.
`adata_filtered = ep.pp.filter_quantiles(adata, vars=["feature 1, feature 2"], quantile_top=1, copy=True)`

I don't really like the quantile_top parameter name. Maybe we can some up with something better.

It would also be great if there was a way to just cap the values -> all 1% values don't get removed, but rather squished to the highest value in the column or something.

### LINKED ISSUE #409: Add feature to automatically winsorize/trim numerical values to handle extreme outliers
**Is your feature request related to a problem? Please describe.**

Extreme outliers (values that likely originate from erroneous  data collection rather than real extreme values) really making it hard for downstream analysis, for example when coloring in UMAPs. Normalization exists, but sometimes (as in my case) one wants to plot features, that are ``obs_only`` (and therefore not normalized). See example below:

image

**Describe the solution you would like**

Winsorizing could help with that, by trimming those values to a set limit.
More info:

- https://towardsdatascience.com/detecting-and-treating-outliers-in-python-part-3-dcb54abaf7b0
- https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.winsorize.html

We should implement something like this for ehrapy to handle those values.

(Alternative would be to delete all columns with extreme outliers, which is not what one wants in most cases.)

Zethson:
fixes #409 
fixes #335 


tests/preprocessing/test_outliers.py
```diff
@@ -29,6 +29,19 @@ def test_winsorized_obs(self):
...
+        np.testing.assert_allclose(np.array(clipped_adata[:, "age"].X, dtype=np.float32), expected)
+
+    def test_clip_obs(self):
+        to_clip_obs = ep.ad.move_to_obs(self.mimic_2_10, "age")
+        clipped_adata = ep.pp.clip_quantile(to_clip_obs, obs_cols=["age"], limits=[25, 50], copy=True)
+        expected = np.array([50, 50, 36.5, 44.49191, 25, 36.54657, 25, 50, 50, 25.41667])
+
+        np.testing.assert_allclose(np.array(clipped_adata.obs["age"]), expected)
+
     def test_quantile(self):
```
Zethson, line 45: Think that this one can be killed now?

docs/usage/usage.md
```diff
@@ -157,6 +157,17 @@ Partially overlaps with dataset integration. Note that a simple batch correction
     preprocessing.neighbors
 ```
 
+### Outlier removal
```
Zethson, line 160: I think that we could add these to the "basic processing section".

namsaraeva:
@Imipenem can you have a quick look? Thank you!

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+) -> AnnData:
+    """Returns a Winsorized version of the input array.
+
+    The implementation is based on https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.winsorize.html
+
+    Args:
+        adata: AnnData object to winsorize
+        vars: The features to winsorize
+        obs_cols: Columns in obs with features to winsorize
+        limits: Tuple of the percentages to cut on each side of the array as floats between 0. and 1. (default: 0.01 and 0.99)
```
Imipenem, line 25: there are default limits -> add to parameters as default parameter?

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+        copy: Whether to return a copy or not
+        **kwargs: Keywords arguments get passed to scipy.stats.mstats.winsorize
+
+    Returns:
+        Winsorized AnnData object if copy is True
+    """
+    if vars is not None and not all(elem in adata.var_names.values for elem in vars):
+        raise ValueError(
+            f"Columns `{[col for col in vars if col not in adata.var_names.values]}` are not in var_names."
```
Imipenem, line 34: Does this work for a single string as well; because we're iterating over `vars` and if this is a `str` we're iterating over the single letters in the `str` don't we?

this could be simplified anyways by making vars a set: Something like the following should work
```
if vars is not None:
    diff = vars - set(adata.var_names)
    if diff:  # difference is not an empty set -> vars has some values that are not in the var_names
        raise ValueError(
            f"Columns `{",".join(var for var in diff)}` are not in var_names."
```

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+    if obs_cols is not None and not all(elem in adata.obs.columns.values for elem in obs_cols):
+        raise ValueError(
+            f"Columns `{[col for col in obs_cols if col not in adata.obs.columns.values]}` are not in obs."
```
Imipenem, line 39: same as vars above

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+        vars = [vars]
+
+    if isinstance(obs_cols, str):  # pragma: no cover
+        obs_cols = [obs_cols]
+
+    if copy:  # pragma: no cover
+        adata = adata.copy()
+
+    if limits is None:
+        limits = [0.01, 0.99]
```
Imipenem, line 52: move to default parameters

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+        raise ValueError(
+            f"Columns `{[col for col in vars if col not in adata.var_names.values]}` are not in var_names."
+        )
+
+    if obs_cols is not None and not all(elem in adata.obs.columns.values for elem in obs_cols):
+        raise ValueError(
+            f"Columns `{[col for col in obs_cols if col not in adata.obs.columns.values]}` are not in obs."
+        )
+
+    if isinstance(vars, str):  # pragma: no cover
```
Imipenem, line 42: allow set only?

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+        vars = [vars]
+
+    if isinstance(obs_cols, str):  # pragma: no cover
```
Imipenem, line 45: set only?

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+    if copy:  # pragma: no cover
+        adata = adata.copy()
+
+    if limits is None:
+        limits = [0.01, 0.99]
+
+    if vars:
+        for var in vars:
+            adata[:, var].X = scipy.stats.mstats.winsorize(
+                np.array(adata[:, var].X), limits=limits, nan_policy="omit", **kwargs
```
Imipenem, line 57: is `np.array(...)` necessary here?

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+
+    if obs_cols:
+        for col in obs_cols:
+            obs_array = adata.obs[col].to_numpy()
+            winsorized_array = scipy.stats.mstats.winsorize(obs_array, limits=limits, nan_policy="omit", **kwargs)
+            adata.obs[col] = pd.Series(winsorized_array).values
+
+    if copy:
+        return adata
+
```
Imipenem, line 68: Maybe add something to `.uns` for qc-reasons like how many values (relatively) got winsorized etc.; might be too much effort for too less though.

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+    if vars:
+        for var in vars:
+            adata[:, var].X = scipy.stats.mstats.winsorize(
+                np.array(adata[:, var].X), limits=limits, nan_policy="omit", **kwargs
+            )
+
+    if obs_cols:
+        for col in obs_cols:
+            obs_array = adata.obs[col].to_numpy()
```
Imipenem, line 62: is this necessary? Docs say `sequence`, doesn't this include a pd.Series?

ehrapy/preprocessing/_outliers.py
```diff
@@ -0,0 +1,123 @@
...
+            clipped_array = np.clip(obs_array, limits[0], limits[1])
+            adata.obs[col] = pd.Series(clipped_array).values
+
+    if copy:  # pragma: no cover
+        adata = adata.copy()
+
+    if copy:
+        return adata
```
Imipenem, line 123: What exactly is the difference between clipping and winsorizing here? Seems to be very similar from just looking at the code.

tests/preprocessing/test_outliers.py
```diff
@@ -0,0 +1,43 @@
...
+
+import ehrapy as ep
+
+CURRENT_DIR = Path(__file__).parent
+_TEST_PATH = f"{CURRENT_DIR}/test_preprocessing"
+
+
+class TestOutliers:
+    def setup_method(self):
+        self.mimic_2_10 = ep.dt.mimic_2()[:10]
```
Imipenem, line 13: running the tests will download the dataset, no? I'm not a huge fan of this, to be honest; we could reuse some of the existing `create a test anndata object` functions.

Imipenem (changes requested):

ehrapy/preprocessing/_outliers.py
Zethson: Mutable datastructures should not be default parameters. This circumvents a flake8 issue (and is the correct way to fix this)

ehrapy/preprocessing/_outliers.py
Zethson: Yes

ehrapy/preprocessing/_outliers.py
Zethson: Don't think that this is necessary

tests/preprocessing/test_outliers.py
Zethson: Yup. scanpy does this as well though. It was easier to write the tests like this :)

We can adapt this to create artificial AnnData objects.

ehrapy/preprocessing/_outliers.py
Zethson: Nice one

ehrapy/preprocessing/_outliers.py
Zethson: clip does it by absolute values and winsorize by relative values I think

PR #420, opened 2022-11-21
Negative values offset

Zethson:
Just a very simple and rudimentary implementation. Was thinking whether we should allow this per column, but honestly it's either all or nothing.


ehrapy/preprocessing/_normalization.py
```diff
@@ -396,3 +403,35 @@ def _record_norm(adata: AnnData, vars_list: list[str], method: str) -> None:
     adata.uns["normalization"] = norm_record
 
     return None
+
+
+def offset_negative_values(adata: AnnData, layer: str = None, copy: bool = False) -> AnnData:
```
Imipenem, line 408: Add `var_names` parameter? So users can pass column names, with negative values; if this is None -> use function as it is now.

tests/preprocessing/test_normalization.py
```diff
@@ -232,3 +232,10 @@ def test_norm_record(self):
         adata_norm = ep.pp.maxabs_norm(adata_norm, vars=["Numeric1"], copy=True)
 
         assert adata_norm.uns["normalization"] == {"Numeric1": ["minmax", "maxabs"], "Numeric2": ["minmax"]}
+
+    def test_offset_negative_values(self):
+        """Test for the offset_negative_values method."""
+        to_offset_adata = AnnData(X=np.array([[-1, -5], [5, 6]]))
```
Imipenem, line 238: Add a column with all positive values and multiple negative values.

Imipenem (changes requested):

ehrapy/preprocessing/_normalization.py
Zethson: Was my o.g. plan, but not sure whether should be possible in the first place. Either all or none?

Imipenem: Yeah valid. Was just thinking whether this could be a "speed improvement" when there is for example only one column, that should be corrected (the only one with negative values), as it's the case in the pediatric ICU dataset.

Issue #423, opened 2022-11-25
Custom exceptions do not show error message.


Imipenem:
**Describe the bug**

![image](https://user-images.githubusercontent.com/31141763/204058690-db78afd2-260f-46e2-a356-e2d87b4cae86.png)

**Expected behavior**

See picture above; the custom exceptions should also display the error message in the notebook.

namsaraeva:
@Imipenem do you have a reproducible example?

PR #428, opened 2022-12-02
KeyError fix issue #423

namsaraeva:
@Imipenem could you please test the case that showed the bug in https://github.com/theislab/ehrapy/issues/423? 

My edit will maybe fix it, but not sure.


Imipenem:
Will do asap, ty

Zethson:
Does this issue also occur in other places? Did you check @Imipenem @namsaraeva ?

Imipenem (commented):
image

Ok message showed up again, but the KeyError is printed as well; can we get rid of this one?

Zethson:
> image
> 
> Ok message showed up again, but the KeyError is printed as well; can we get rid of this one?

Discussed with @namsaraeva and we think that in this case keeping the whole traceback is preferable. Please shout at me if you greatly disagree.

Issue #422, opened 2022-11-22
df_to_anndata does not assign binary numerical columns correctly in .uns


Imipenem:
**Describe the bug**

When calling `ep.df_to_anndata(...)` the binary columns (columns with only two values like 1,0) do not get correctly assigned to "numerical_columns" in `uns`.

**To Reproduce**

Example:

```
df = pd.DataFrame({"Col1":[i for i in range(4)], "Col2":["str" + str(i) for i in range(4)], "Col3":[1.0,0.0,np.nan,1.0]})
adata = ep.ad.df_to_anndata(df)
print(adata.uns)

# OverloadedDict, wrapping:
	# OrderedDict([('numerical_columns', ['Col1']), ('non_numerical_columns', ['Col3', 'Col2'])])
# With overloaded keys:
	# ['neighbors'].
```
`Col3` should be numerical!

**Expected behavior**

The binary columns with numerical values such as only 0 and 1 should be assigned to numerical columns.
The bug is caused by the following line: https://github.com/theislab/ehrapy/blob/e1515b0f14b1a24fa230ae83ba1066cea46117f4/ehrapy/anndata/anndata_ext.py#L71

The XOR excludes the binary columns since they are in both sets. This should be fixed.

**System [please complete the following information]:**
-   Ehrapy version: 0.2.0

Zethson:
We are unfortunately still running into this. We can supply an object to test this out and fix this very soon.

PR #429, opened 2022-12-02
df_to_anndata logical XOR to OR, issue #422

namsaraeva:
**Description of changes**

Changed logical XOR to OR, so the binary columns are not excluded when assigning numerical columns.

Are there any other edge cases? I can't think of anything else right now.


Zethson:
One of the tests is failing. Can we also cover what @Imipenem discovered with a super small test?

Imipenem (commented):
As @Zethson  suggested, small test and we're good 👍

Imipenem (approved):
LGTM

PR #437, opened 2023-01-09
Add docs group

Zethson:


Zethson:
@Imipenem the poetry way of doing things. I'm not happy with it, but I guess it's better than before without the extra requirements file?

PR #498, opened 2023-04-11
Added tiny examples to every function

namsaraeva:
Added examples to some functions

### DIFF
--- a/ehrapy/io/_write.py
+++ b/ehrapy/io/_write.py
@@ -35,7 +35,7 @@ def write(
 
             import ehrapy as ep
 
-            adata = eh.data.mimic_2(encode=True)
+            adata = ep.dt.mimic_2(encoded=True)
             ep.io.write("mimic_2.h5ad", adata)
     """
     filename = Path(filename)  # allow passing strings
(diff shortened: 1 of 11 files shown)


ehrapy/io/_write.py
```diff
@@ -59,7 +59,7 @@ def write(
     else:
         # dummy encoding when there is non numerical data in X
         if not np.issubdtype(adata.X.dtype, np.number) and extension == "h5ad":
-            # flag to indicate an Anndata object has been dummy encoded to write it to .h5ad file
+            # flag to indicate an anndata object has been dummy encoded to write it to .h5ad file
```
Zethson, line 62: Why this?

Zethson (commented):
Almost there! Great and you also fixed a few old docstrings

ehrapy/io/_write.py
namsaraeva: Oh, was not really intended. We can rollback

Zethson: yes pls

then merge it :)

PR #502, opened 2023-04-19
Causal Inference with dowhy

timtreis:
tbd


Zethson:
```
Because no versions of sphinx-design match >=0.0.13,<0.3.0 || >0.3.0
 and sphinx-design (0.3.0) depends on sphinx (>=4,<6), sphinx-design (>=0.0.13) requires sphinx (>=4,<6).
And because myst-parser (2.0.0) depends on sphinx (>=6,<8)
 and no versions of myst-parser match >=0.17.0,<2.0.0 || >2.0.0, sphinx-design (>=0.0.13) is incompatible with myst-parser (>=0.17.0).
So, because ehrapy depends on both myst-parser (>=0.17.0) and sphinx-design (>=0.0.13), version solving failed.
```

That's my fault and I'll fix this.

ehrapy/tools/causal/_dowhy.py
```diff
@@ -0,0 +1,362 @@
...
+
+    Returns:
+        A tuple containing the causal estimate and a dictionary of the results of any refutation tests.
+
+    Raises:
+        TypeError: If adata, graph, treatment, outcome, refute_methods, estimation_method, or return_as is not of the expected type.
+        ValueError: If refute_methods or estimation_method contains an unknown value, or if return_as is an unknown value.
+
+    Examples:
+        >>> data = dowhy.datasets.linear_dataset(
```
Zethson, line 102: Guess I should add this to the ehrapy datasets at some point to make it easier

Zethson (changes requested):
1. Please remove causal_model.png
2. Please remove the dowhy.ipynb. Please add the tutorial to the ehrapy-tutorials repository. The notebook also still has errors in it.

ehrapy/tools/causal/_dowhy.py
timtreis: For testing purposes? Through the causal module we'll have dowhy as a dependency anyway, no? Might as well use their function in the test, no?

timtreis:
> Please remove the dowhy.ipynb. Please add the tutorial to the ehrapy-tutorials repository. The notebook also still has errors in it.

The notebook here was just a relict, but thanks for pointing it out 👌 We had already merged the proper notebook into the tutorials repo

https://github.com/theislab/ehrapy-tutorials/blob/main/mimic_2_causal_inference.ipynb

Issue #447, opened 2023-01-30
ep.pp.qc_lab_measurements cannot deal with some measurements


xinyuejohn:
**Describe the bug**

**To Reproduce**

```
import pandas as pd
import ehrapy as ep
import numpy as np
data = pd.DataFrame(np.array([[100, 98], [162, 107]]), columns=['oxygen saturation', 'glucose'], index=[0, 1])
adata = ep.ad.df_to_anndata(data)

# This will cause ValueError (ValueError: could not convert string to float: '0.7 0.3')
ep.pp.qc_lab_measurements(adata, measurements=['oxygen saturation'])

```

**System [please complete the following information]:**

-   OS: e.g. [MacOS 13.1]
-   Language Version: [Python 3.8]
-   Virtual environment: [Conda]

**Additional context**

1) In the above example, measurements=['oxygen saturation'], and the final matched array has 2 rows:

	<img width="234" alt="Screenshot 2023-04-11 at 22 09 25" src="https://user-images.githubusercontent.com/88339105/231277212-391f3b04-2444-441e-aa59-9211120b62a3.png">

2) The final matched array will be transformed to a string:
`check_str: str = np.array2string(check)`
	Then the array will transformed to "0.2-0.7 0.3-0.9"
3) Then the string will be split by:
	```
	min_value = float(check_str.split("-")[0])
    max_value = float(check_str.split("-")[1]) 
	```
	It will case error.

**Suggest solution**
	
Make sure the final matched array has only one row by aggregating several rows in some way. (By warning the user input more information, e.g. sex_col, age_col, ethnicity_col or just simply averaging several rows.

Zethson:
Thank you! Could you please fix it quickly? Thanks

Zethson:
Think that @xinyuejohn mentioned at some point that this is a bit more broken and requires more fixes? Would be awesome if @xinyuejohn and @namsaraeva you could quickly connect on mattermost/IRL and discuss this so that @namsaraeva can fix this?

@namsaraeva this doesn't have a high priority, but should get fixed after you're more free and have some more time.

xinyuejohn:
I will talk to @namsaraeva after I finish Reproduce Eulenberg.

namsaraeva:
@Zethson @xinyuejohn I am working on it

xinyuejohn:
[...]

PR #504, opened 2023-04-24
Error message and minor fixes, issue #447

namsaraeva:
Potential (and temporary) fix for the issue #447. 

As discussed with @Zethson, we should error out the case when multiple reference values are provided, but no additional information, e.g. sex/age/ethnicity. 

I added some code to check if additional columns like sex/age/ethnicity were passed as arguments (at least one is needed to make the value True). Next, if this value is False (no additional column passed) and more than 1 reference value was detected, ValueError is thrown with the message to add additional information. 

Not sure how to make it more specific and flexible, since it is theoretically possible that dataframes other than laposata will be passed onto this method and columns other than sex/age/ethnicity are possible as well.

I also did not really catch what is the difference between our laposata dataframe and the one @xinyuejohn mentioned recently (https://accessmedicine.mhmedical.com/content.aspx?bookid=1069&sectionid=60775149)? Visually they are the same to me, however we have a Sex column.  

Generally, I think we should somehow re-design this method and make it more flexible for other covariates... but that's a big goal. 

Eager to discuss other possible solutions with @Zethson and @xinyuejohn


ehrapy/preprocessing/_quality_control.py
```diff
@@ -263,13 +263,13 @@ def qc_lab_measurements(
         adata: Annotated data matrix.
         reference_table: A custom DataFrame with reference values. Defaults to the laposata table if not specified.
         measurements: A list of measurements to check.
-        unit: The unit of the measurements. Defaults to SI .
+        unit: The unit of the measurements. Defaults to 'traditional'.
```
Zethson, line 266: Hmm, should this be the default? Is it really more common?

Zethson: And is it true? :joy:

Zethson (changes requested):
@namsaraeva could you add a small test for this case, please?

ehrapy/preprocessing/_quality_control.py
namsaraeva: I changed it to `traditional` due to this line: 

`reference_column = "SI Reference Interval" if unit == "SI" else "Traditional Reference Interval"`

But I don't really know what is real default

PR #510, opened 2023-05-06
Add Synthea dataset

namsaraeva:
Adds:
- Synthea dataset (in older DSTU2 format, since it is much smaller than the recent R4 one). This dataset contains 1000 synthetic patient medical records. No reply from fhiry developers on the correct FHIR version yet.

Fixes:
- Some .zip files have different directory name when unpacked, so I added a fix to `_get_non_existing_files()` function to handle this case instead of throwing an error that the directory was not found. The fix looks for the last created file/dir (meaning a .zip file was unpacked and a directory was created) in our dataset directory and changes the returned path to it. 
- The fix leaves out hidden files that begin with `.`

Questions:
- Do we want more Synthea datasets? The COVID-19 one is available only in `.csv` format though. Other datasets are too heavy.


ehrapy/data/_datasets.py
```diff
@@ -681,3 +682,40 @@ def heart_disease(
...
+        encoded: Whether to return an already encoded object.
+        columns_obs_only: Columns to include in obs only and not X.
+
+    Returns:
+        :class:`~anndata.AnnData` object of the 1K Sample Synthetic Patient Records Data Set
+
+    Example:
+        .. code-block:: python
+
+            import ehrapy as ep
```
Zethson, line 706:
```suggestion
            >>> import ehrapy as ep
```

ehrapy/data/_datasets.py
```diff
@@ -681,3 +682,40 @@ def heart_disease(
...
+
```
Zethson, line 707:
```suggestion
```

ehrapy/data/_datasets.py
```diff
@@ -681,3 +682,40 @@ def heart_disease(
...
+            adata = ep.dt.synthea_1k_sample(encoded=True)
```
Zethson, line 708:
```suggestion
            >>> adata = ep.dt.synthea_1k_sample(encoded=True)
```

ehrapy/io/_read.py
```diff
@@ -488,8 +488,14 @@ def _get_non_existing_files(file: Path, download_dataset_name: str, backup_url:
     output_path_name = download_default_name.replace(archive_extension, "") if is_archived else download_default_name
     output_file_or_dir = ehrapy_settings.datasetdir / output_path_name
     moved_path = Path(str(output_file_or_dir)[: str(output_file_or_dir).rfind("/") + 1]) / download_dataset_name
-    shutil.move(output_file_or_dir, moved_path)  # type: ignore
-    file = moved_path
+    if moved_path.exists():
+        shutil.move(output_file_or_dir, moved_path)  # type: ignore
+        file = moved_path
+    elif not moved_path.exists():
```
Zethson, line 494: Can you add a comment above this to explain why we need this case? I know it but other people reading the code may not yet know the answer.

ehrapy/io/_read.py
```diff
@@ -488,8 +488,14 @@ def _get_non_existing_files(file: Path, download_dataset_name: str, backup_url:
...
+        list_of_paths = [path for path in ehrapy_settings.datasetdir.glob("*/") if not path.name.startswith(".")]
+        latest_path = max(list_of_paths, key=lambda x: x.stat().st_ctime)
```
Zethson, line 496: Rename `x`

Zethson (approved):

ehrapy/data/_datasets.py
namsaraeva: What do you mean by this? @Zethson

PR #551, opened 2023-06-26
Dataloader throwing an error using Windows

namsaraeva:
Previous approach for building the download path did not take the Windows platform feature into account (`\\` instead of `/`).


Zethson (approved):
pls merge as soon as Rainer confirms that this one works

PR #567, opened 2023-07-26
Solution for adata.uns["non_numericlal_columns"] being empty in mimic_2 dataset

namsaraeva:
Issue was that `adata.uns["non_numerical_columns"]` was empty after calling `adata = ep.dt.mimic_2(encoded=True)`, making evaluation of categorical data in `rank_features_groups` impossible. 

I added a new slot into `adata.uns` which is called [`encoded_non_numerical_columns`]. This stores the column names of encoded categorical columns. This slot is accessed in `_rank_features_groups`.

Tests were adjusted accordingly. 

I am not sure about the naming of the new slot, so feel free to comment on this.


Zethson (approved):
LGTM!

Issue #416, opened 2022-10-25
Documentation warnings


Zethson:
```
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT": name 'Vocab' is not defined
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.create_vocabulary": name 'Vocab' is not defined
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.load_concept_db": name 'CDB' is not defined
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.load_vocabulary": name 'Vocab' is not defined
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.save_vocabulary": name 'Vocab' is not defined
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.update_cat": name 'Vocab' is not defined
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.update_cat_config": name 'Config' is not defined
/home/zeth/PycharmProjects/ehrapy/docs/usage/tools/ehrapy.tools.MedCAT.rst:48: WARNING: Explicit markup ends without a blank line; unexpected unindent.
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.create_vocabulary": name 'Vocab' is not defined
/home/zeth/PycharmProjects/ehrapy/docs/usage/tools/ehrapy.tools.MedCAT.rst:52: WARNING: Explicit markup ends without a blank line; unexpected unindent.
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.load_concept_db": name 'CDB' is not defined
/home/zeth/PycharmProjects/ehrapy/docs/usage/tools/ehrapy.tools.MedCAT.rst:56: WARNING: Explicit markup ends without a blank line; unexpected unindent.
WARNING: Cannot resolve forward reference in type annotations of "ehrapy.tools.MedCAT.load_vocabulary": name 'Vocab' is not defined
/home/zeth/PycharmProjects/ehrapy/docs/usage/tools/ehrapy.tools.MedCAT.rst:60: WARNING: Explicit markup ends without a blank line; unexpected unindent.
[...]

PR #605, opened 2023-11-01
Add medcat to doc dependency list

eroell:
**PR Checklist**

**Description of changes**
Adresses issue #416 .
This could be a way of omitting the medcat warnings when building the docs.

**Technical details**
None?

**Additional context**
Medcat is a big package, and will add additional installation overhead when intending to install the doc dependencies.
However, it will reduce the number of warnings when building the documentation.


eroell:
Draft PR to deposit this suggestion. Might re-discuss if really want this & whether more warnings can be caught here.

Zethson:
I don't think it works like this. With `ehrapy[medcat]` it will attempt to install all of ehrapy from the current pypi version (that may have different requirements than the development version) and the medcat extra. However, we only want the medcat extra.

Zethson:
CI failures are unrelated and tracked on AnnData

PR #664, opened 2024-03-07
More sa functions

fatisati:
-   [ add 4 more survival analysis functions.]

### DIFF
--- a/tests/tools/test_sa.py
+++ b/tests/tools/test_sa.py
@@ -1,7 +1,7 @@
 import numpy as np
 import pytest
 import statsmodels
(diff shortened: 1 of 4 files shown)


tests/tools/test_sa.py
```diff
@@ -1,7 +1,14 @@
 import numpy as np
 import pytest
 import statsmodels
-from lifelines import CoxPHFitter, KaplanMeierFitter
+from lifelines import (
```
Zethson, line 4: Why import the lifelines versions? You should always use the ehrapy implementation

Zethson (approved):
Thank youuu

PR #672, opened 2024-03-15
change diabetes-130 datasets which are provided

### LINKED ISSUE #671: Diabetes-130 Dataset: Change the dataset instance loaded
### Description of feature

Currently, [ehrapy.data.diabetes_130](https://ehrapy.readthedocs.io/en/latest/usage/data/ehrapy.data.diabetes_130.html#ehrapy.data.diabetes_130) loads a custom preprocessed dataset as indicated there.

There is an alternative pre-processed dataset available [from fairlearn](https://fairlearn.org/v0.10/api_reference/generated/fairlearn.datasets.fetch_diabetes_hospital.html#fairlearn.datasets.fetch_diabetes_hospital): This e.g. better matches the pre-processing in [Strack et al. 2014](https://www.hindawi.com/journals/bmri/2014/781670/), and showcases interesting evaluations.

Switching to this preprocessed dataset, while optionally allowing to load the raw data instead, leverages existing resources better.

eroell:
**PR Checklist**

**Description of changes**
Fixes #671.

**Technical details**
- Deletes `ep.data.diabetes_130(encoded=False, columns_obs_only=None)`
- Adds `ep.data.diabetes_130_raw(encoded=False, columns_obs_only=None)`
- Adds `ep.data.diabetes_130_fairlearn(encoded=False, columns_obs_only=None)`
    - Uses data from [fairlearn.datasets.fetch_diabetes_hospital](https://fairlearn.org/v0.10/api_reference/generated/fairlearn.datasets.fetch_diabetes_hospital.html)

**Questions**
- `ep.data.diabetes_130_raw` downloads a zip (with the csv inside). Is it better to download and extract that, or to store the csv only on figshare?

**To clean up**
- [x] Adapt usage of diabetes_130 in CohortTracker examples
- [x] Update the ehrapy-datasets Repo [here](https://github.com/theislab/ehrapy-datasets/pull/23)


ehrapy/data/_datasets.py
```diff
@@ -168,13 +177,53 @@ def diabetes_130(
...
+
+
 def chronic_kidney_disease(
     encoded: bool = False,
     columns_obs_only: dict[str, list[str]] | list[str] | None = None,
 ) -> AnnData:  # pragma: no cover
     """Loads the Chronic Kidney Disease dataset
 
     More details: https://archive.ics.uci.edu/ml/datasets/Chronic_Kidney_Disease
+
```
eroell, line 226: If no space here, the two lines are just separated by a space instead of a linebreak

eroell:
By the way, wondered after a text search for diabetes_130 what's up with [this line](https://github.com/theislab/ehrapy/blob/627f5dd05c2a3553e02c4dd805837ae5806faa1c/docs/conf.py#L180C5-L180C74)?

ehrapy/data/_datasets.py
```diff
@@ -153,12 +161,13 @@ def diabetes_130(
...
+        >>> adata = ep.dt.diabetes_130_raw(encoded=True)
     """
+
     adata = read_csv(
-        dataset_path=f"{ehrapy_settings.datasetdir}/diabetes_130.csv",
-        download_dataset_name="diabetes_130.csv",
-        backup_url="https://figshare.com/ndownloader/files/33950546",
+        dataset_path=f"{ehrapy_settings.datasetdir}/diabetes_130_raw.csv",
+        download_dataset_name="diabetes_130_raw.csv",
+        backup_url="https://figshare.com/ndownloader/files/33950546",  # TODO: change link, or download zip and extract?
```
eroell, line 170: This is where the question from the first comment applies

Zethson:
Careful:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/docs/checkouts/readthedocs.org/user_builds/ehrapy/checkouts/672/ehrapy/__init__.py", line 12, in <module>
    from ehrapy import data as dt
  File "/home/docs/checkouts/readthedocs.org/user_builds/ehrapy/checkouts/672/ehrapy/data/__init__.py", line 1, in <module>
    from ehrapy.data._datasets import (
  File "/home/docs/checkouts/readthedocs.org/user_builds/ehrapy/checkouts/672/ehrapy/data/_datasets.py", line 6, in <module>
    from fairlearn.datasets import fetch_diabetes_hospital
ModuleNotFoundError: No module named 'fairlearn'
```

> ep.data.diabetes_130_raw downloads a zip (with the csv inside). Is it better to download and extract that, or to store the csv only on figshare?

Probably we could even store parquet? Anyways, the file should be so small that I guess it's okay to store anything be it zip or csv.

> By the way, wondered after a text search for diabetes_130 what's up with [this line](https://github.com/theislab/ehrapy/blob/627f5dd05c2a3553e02c4dd805837ae5806faa1c/docs/conf.py#L180C5-L180C74)?

We once had a tutorial on the dataset. This is an artifact of that which we can remove.

Zethson:
How different are the raw and the preprocessed version? I wonder whether we can condense them into a single link/dataset?

eroell:
> ModuleNotFoundError: No module named 'fairlearn'

Yes, doing it this way only under the assumption that fairlearn becomes a dependency. If not enough reason (their dataset, using some of their things in the Fairness/Bias tutorials), fall back to openml? if that is also an additional dependency we dont want, figshare 

> How different are the raw and the preprocessed version?

raw shape: (101766, 24)  VS preprocessed shape: (101766, 24) See preprocessing [here](https://github.com/fairlearn/talks/blob/main/2021_scipy_tutorial/preprocess.py). With single dataset, you mean downloading raw and preprocess with the script?

Other suggestion might be to load with additional argument `ep.dt.diabetes_130(processing=<"raw" or preprocessed_fairlearn>)`? Not super in sync with other datasets where the only two arguments are "encoded" and "columns_obs_only"

> Probably we could even store parquet? Anyways, the file should be so small that I guess it's okay to store anything be it zip or csv

Sorry that was unclear, not size is issue, just consistency for users: think all `ep.dt.<datasets>` download a `<datasets>.csv` file to a dedicated `/data` directory: thought to stay consistent here as well, maybe doing so by extracting & grabbing that csv? Again, just storing the dataset to figshare would also be a solution

ehrapy/data/_datasets.py
eroell: Starting to think figshare is indeed a neat and quick way instead of dwnloading things from there and extracting

Zethson:
> > ModuleNotFoundError: No module named 'fairlearn'
> 
> Yes, doing it this way only under the assumption that fairlearn becomes a dependency. If not enough reason (their dataset, using some of their things in the Fairness/Bias tutorials), fall back to openml? if that is also an additional dependency we dont want, figshare

https://github.com/fairlearn/fairlearn/blob/main/requirements.txt looks like we already have all dependencies in ehrapy so we'd only be adding fairlearn. However, are you planning to implement some of the fairlearn metrics into ehrapy or do you only want to show the usage of fairlearn as an external package in the bias notebook?

> > How different are the raw and the preprocessed version?
> 
> raw shape: (101766, 24) VS preprocessed shape: (101766, 24) See preprocessing [here](https://github.com/fairlearn/talks/blob/main/2021_scipy_tutorial/preprocess.py). With single dataset, you mean downloading raw and preprocess with the script?
> 
> Other suggestion might be to load with additional argument `ep.dt.diabetes_130(processing=<"raw" or preprocessed_fairlearn>)`? Not super in sync with other datasets where the only two arguments are "encoded" and "columns_obs_only"

We decided to go with 2 dataloaders

eroell:
Do we need a deprecation warning for diabetes_130, or is this a functionality which we allow to be killed without further notice?

eroell:
think this should do it, more suggestions?

ehrapy/data/_datasets.py
```diff
@@ -153,28 +157,74 @@ def diabetes_130(
...
     return adata
 
 
+def diabetes_130_fairlearn(
+    encoded: bool = False,
+    columns_obs_only: dict[str, list[str]] | list[str] | None = None,
+) -> AnnData:
+    """Loads the preprocessed diabetes-130 dataset by fairlearn
+
+    This loads the dataset from the `fairlearn.datasets.fetch_diabetes_hospital` function.
```
Zethson, line 184: Well, it doesn't really use it but rather the same dataset that this function would download. But it's probably fine :)

ehrapy/data/_datasets.py
```diff
@@ -153,28 +157,74 @@ def diabetes_130(
...
+    Examples:
+        >>> import ehrapy as ep
+        >>> adata = ep.dt.diabetes_130_fairlearn()
+
+    References:
+        [1] Beata Strack, Jonathan P. DeShazo, Chris Gennings, Juan L. Olmo, Sebastian Ventura, Krzysztof J. Cios, and John N. Clore, “Impact of HbA1c Measurement on Hospital Readmission Rates: Analysis of 70,000 Clinical Database Patient Records,” BioMed Research International, vol. 2014, Article ID 781670, 11 pages, 2014.
+
+        [2] Bird, S., Dudík, M., Edgar, R., Horn, B., Lutz, R., Milan, V., ... & Walker, K. (2020). Fairlearn: A toolkit for assessing and improving fairness in AI. Microsoft, Tech. Rep. MSR-TR-2020-32.
+    """
+
```
Zethson, line 206:
```suggestion
```

Zethson (approved):
Thank you!

PR #677, opened 2024-03-28
Calculation of feature importances in a supervised setting

### LINKED ISSUE #258: Calculate feature importances easily
Although ehrapy is primarily designed for exploratory analysis, we also want to offer the ability to quickly run prediction models to:

1. Find features that are strongly predictive of the target variable
2. Get features importances to determine biases and imbalances
3. Simple baseline models to justify bigger DL models

Ideally, we'd implement one or two simple models (regression and classification) that are fast (no deep learning) which try to take the nature of EHR data into account (well mostly just requires lot of ehrapy preprocessing but w/e). The API must be intuitive and should resemble (and just be based on) simple scikit-learn stuff. We already depend on it anyways.

We should try to ensure that this is compatible with our https://github.com/fairlearn/fairlearn efforts. See #647

Lilly-May:
**PR Checklist**

-   [X] Referenced issue is linked (closes #258)

**Description of changes**

- Added a new method `feature_importances` to calculate the contribution of each feature to the prediction of a specified feature. For instance, one could be interested in what features are good predictors of disease severity, and depending on the disease, features such as age or pre-existing conditions would have high feature importance.
- Users have the option to choose between a continuous or a categorical prediction, as well as the option to choose between using SVM, Random Forest, or Regression to perform the prediction
- Implemented a basic plotting function for visualizing individual feature importances with a bar plot (see example below)

**ToDos**

- [x] Test performance on the test set. Should we print the resulting loss/accuracy out then? Should that somehow affect the plotting function, e.g. as an indicator of reliability?
- [x] Improve the plotting function. Personally, I would prefer to plot everything in the same color - what do you think?

**Considerations**

- Currently, the importances are only comparable relative to each other; we can't compare them across models or datasets. I'll think more about how we could improve interpretability.
- How do we handle non-numerical features? Currently, I'm simply dropping them and give a warning. Should we instead throw an error stating it's necessary to properly encode features first?
- We should discuss if we want to incorporate bias detection (e.g. FairLearn). 

**Example**
```
adata = ep.dt.mimic_2(encoded=False)
ep.pp.knn_impute(adata, n_neighbours=5)
ep.tl.feature_importances(adata, predicted_feature="tco2_first", prediction_type="continuous", model="rf", input_features="all")
ep.pl.feature_importances(adata)
```
feature_importances


Zethson:
> How do we handle non-numerical features? Currently, I'm simply dropping them and give a warning. Should we instead throw an error stating it's necessary to properly encode features first?

Yes, I would not drop them. Maybe even throw a ValueError with the suggested solution

Lilly-May:
> Yes, I would not drop them. Maybe even throw a ValueError with the suggested solution

I adjusted the code as suggested: Non-numeric features now cause the function to fail with a ValueError, requesting a proper encoding of the features beforehand.

ehrapy/plot/supervised/_feature_importances.py
```diff
@@ -0,0 +1,30 @@
+import matplotlib.pyplot as plt
+import pandas as pd
+import seaborn as sns
+from anndata import AnnData
+
+
+def feature_importances(adata: AnnData, key: str = "feature_importances", n_features: int = 10):
+    """
+    Plot features with greates absolute importances as a barplot.
```
Zethson, line 9: This should be moved up into the first line of the docstring -> `"""Plot features bla.`

ehrapy/plot/supervised/_feature_importances.py
```diff
@@ -0,0 +1,30 @@
...
+
+    Args:
+        adata: :class:`~anndata.AnnData` object storing the data. A key in adata.var should contain the feature
+            importances, calculated beforehand.
+        key: The key in adata.var to use for feature importances. Defaults to 'feature_importances'.
+        n_features: The number of features to plot. Defaults to 10.
+
+    Returns:
```
Zethson, line 17: We remove the `Returns:` section if the function does not return anything. I even asked chatGPT to confirm whether this is fine :)

> 
In Google style docstrings, if a function does not return anything (i.e., it implicitly returns `None`), it is considered good practice to omit the `Returns` section entirely. The absence of a `Returns` section implicitly indicates that the function does not return a value or that its return value is not significant. Including a `Returns` section that states `None` is not wrong, but it is unnecessary and can be seen as redundant. Therefore, for brevity and clarity, you should omit the `Returns` section in such cases.

ehrapy/plot/supervised/_feature_importances.py
```diff
@@ -0,0 +1,30 @@
...
+        None
+    """
+    if key not in adata.var.keys():
+        raise ValueError(f"Key {key} not found in adata.var.")
```
Zethson, line 21: This would raise a `KeyError` anyways. Catching this is more helpful if you specify how to resolve -> calculate the feature importances first by doing blabla

ehrapy/plot/supervised/_feature_importances.py
```diff
@@ -0,0 +1,30 @@
...
+
+    df = pd.DataFrame({"importance": adata.var[key]}, index=adata.var_names)
+    df["absolute_importance"] = df["importance"].abs()
+    df = df.sort_values("absolute_importance", ascending=False)
+    sns.barplot(x=df["importance"][:n_features], y=df.index[:n_features], orient="h")
```
Zethson, line 26: 1. Should we pass `**kwargs` to this?
2. Please use the style of https://github.com/theislab/pertpy/pull/574 to allow for ax to be passed, the figures to be saved etc. The plot currently does not allow for a lot of control or customization...

ehrapy/tools/supervised/_feature_importances.py
```diff
@@ -0,0 +1,157 @@
...
+    input_features: list[str] | Literal["all"] = "all",
+    layer: str | None = None,
+    test_split_size: float = 0.2,
+    key_added: str = "feature_importances",
+    feature_scaling: Literal["standard", "minmax"] | None = "standard",
+    percent_output: bool = False,
+    **kwargs,
+):
+    """
+    Calculate feature importances for predicting a specified feature in adata.var using a given model.
```
Zethson, line 28: Move this one up, please.

ehrapy/tools/supervised/_feature_importances.py
```diff
@@ -0,0 +1,157 @@
...
+            Defaults to 0.2.
+        key_added: The key in adata.var to store the feature importances. Defaults to 'feature_importances'.
+        feature_scaling: The type of feature scaling to use for the input. Choose between 'standard', 'minmax', or None.
+            'standard' uses sklearn's StandardScaler, 'minmax' uses MinMaxScaler. Scaler will be fit and transformed
+            for each feature individually. Defaults to 'standard'.
+        percent_output: Set to True to output the feature importances as percentages. Note that information about positive or negative
+            coefficients for regression models will be lost. Defaults to False.
+        **kwargs: Additional keyword arguments to pass to the model. See the documentation of the respective model in scikit-learn for details.
+
+    Returns:
```
Zethson, line 51: Remove, please.

ehrapy/tools/supervised/_feature_importances.py
```diff
@@ -0,0 +1,157 @@
...
+
+from ehrapy import logging as logg
+
+
+def feature_importances(
+    adata: AnnData,
+    predicted_feature: str,
+    prediction_type: Literal["continuous", "categorical"],
+    model: Literal["regression", "svm", "rf"] = "regression",
+    input_features: list[str] | Literal["all"] = "all",
```
Zethson, line 19: This could probably also be of type `Iterable` instead of `list`. We try to always use the most generic type possible to not constrain users. "possible" here means: Without introducing too much complexity and still KISS it

ehrapy/tools/supervised/_feature_importances.py
```diff
@@ -0,0 +1,157 @@
...
+        raise ValueError(
+            f"Prediction type {prediction_type} not recognized. Please choose either 'continuous' or 'categorical'."
+        )
+
+    if feature_scaling not in ["standard", "minmax", None]:
+        raise ValueError(
+            f"Feature scaling type {feature_scaling} not recognized. Please choose either 'standard', 'minmax', or None."
+        )
+
+    if layer is not None:
```
Zethson, line 75: Consider using https://ehrapy.readthedocs.io/en/latest/usage/anndata/ehrapy.anndata.anndata_to_df.html#ehrapy.anndata.anndata_to_df

ehrapy/tools/supervised/_feature_importances.py
```diff
@@ -0,0 +1,157 @@
...
+        data = adata.layers[layer].to_df()
+    else:
+        data = adata.to_df()
+
+    if prediction_type == "continuous":
```
Zethson, line 80: Since you're doing that check anyways, I wonder whether you can't just infer the `prediction_type` automatically?

ehrapy/tools/supervised/_feature_importances.py
```diff
@@ -0,0 +1,157 @@
...
+                f"feature importances or drop them from the input_features list."
+            ) from e
+
+    x_train, x_test, y_train, y_test = train_test_split(input_data, labels, test_size=test_split_size)
+
+    predictor.fit(x_train, y_train)
+
+    score = predictor.score(x_test, y_test)
+    evaluation_metric = "R2 score" if prediction_type == "continuous" else "accuracy"
+    logg.info(f"Training completed. The model achieved an {evaluation_metric} of {score:.2f} on the test set.")
```
Zethson, line 145: We could consider mentioning the size of the test set? Not sure whether it's all that important...

ehrapy/tools/supervised/_feature_importances.py
```diff
@@ -0,0 +1,157 @@
...
+
+    if model == "regression" or model == "svm":
+        feature_importances = pd.Series(predictor.coef_.squeeze(), index=input_data.columns)
+    else:
+        feature_importances = pd.Series(predictor.feature_importances_.squeeze(), index=input_data.columns)
+
+    if percent_output:
+        feature_importances = feature_importances.abs() / feature_importances.abs().sum() * 100
+
+    # Reorder feature importances to match adata.var order and save importances in adata.var
```
Zethson, line 155: Great!

Zethson (commented):
Really good work! Thank you very much.

Besides my comments, we also need to add this to the documentation of ehrapy somewhere. I would try not to add a new section, but I'll leave it up to you to make a suggestion.

ehrapy/tools/supervised/_feature_importances.py
Lilly-May: I could infer the `prediction_type`, but having it as a separate parameter serves as a kind of sanity check. For instance, if the variable we want to predict is saved numerically, it becomes unclear if the prediction should be continuous or categorically unless it was explicitly saved as a categorical variable. Since I assume that most features are encoded numerically (even when they are categorical), this would happen quite often. But, I also get that it would be nice to remove the parameter, so, if you feel that this double-check is unnecessary, I'll remove it and infer the `prediction_type`.

ehrapy/tools/supervised/_feature_importances.py
Lilly-May: I think it's a nice idea since it provides information on the reliability of the test score. I've added it.

Zethson:
Awesome! Just one more thing to resolve (https://github.com/theislab/ehrapy/pull/677#discussion_r1549682387) and then we're good to go!

ehrapy/tools/supervised/_feature_importances.py
Lilly-May: I finally looked into this and implemented what you suggested, but I think we need to reconsider if this is really what we want. The main point is that if a feature is numerical, it shouldn't necessarily be considered continuous. For example, the `aline_flg` variable in the `mimic_2` dataset is set to either 0 or 1 for each patient. Hence, this should be treated as a classification problem, but as it is stored numerically, determining the `prediction_type` based on `adata.var["ehrapy_column_type"]`, will result in a continuous prediction. Changing this would be super annoying for the user, as they would need to manually change the encoding type in `adata.var["feature_encoding_type"]` or store values "yes" and "no" instead of 1 and 0.

I would suggest keeping the `prediction_type` parameter, but we could add the default option `auto`, where we infer the type using the approach you suggested. We log the inferred type, and if it's not what the user wants (as in the case of `aline_flg`), they still have the option to set the parameter to 'categorical'. Do you think adding the `auto` option just makes things more complex or would it be helpful?

ehrapy/tools/feature_ranking/_feature_importances.py
```diff
@@ -0,0 +1,163 @@
...
+                data[predicted_feature] = data[predicted_feature].astype("category")
+            except ValueError as e:
+                raise ValueError(
+                    f"Feature {predicted_feature} is not categorical and conversion to category failed. Either change the prediction "
+                    f"type to 'continuous' or change the feature data type to a categorical type."
+                ) from e
+
+        if data[predicted_feature].nunique() > 2 and model in ["regression", "svm"]:
+            raise ValueError(
+                f"Feature {predicted_feature} has more than two categories. Please choose random forest (rf) as model for multi-class classification."
```
Zethson, line 113:
```suggestion
                f"Feature {predicted_feature} has more than two categories. Please choose 'rf' as model for multi-class classification."
```

Maybe it's simpler like this?

Zethson (approved):

Issue #647, opened 2024-02-01
Bias detection module


Zethson:
### Description of feature

Pre-requisite: https://github.com/theislab/ehrapy/issues/637

This serves as a meta issue to collect ideas for a bias detection and mitigation module:

- [x] #645 
- [x] Missingness per feature -> missingno stuff
- [x] IQR outliers -> https://github.com/theislab/ehrapy/issues/648 for numerical data @Zethson 
- [x] value counts for categorical counts + warnings? Some things are already in tableone @Zethson 
- [x] sensitivity analysis - easy with ehrapy because the API is so easy @Zethson 
- [x] Imbalanced subsampling through https://imbalanced-learn.org/stable/index.html ? -> only the most important sampling features in ehrapy. Not the generation stuff. @eroell 
- [ ] Propensity score matching @Lilly-May 
- [x] standardized mean differences for numerical data @Lilly-May (#690)
- [x] Correlations between variables @Lilly-May (#690)
- [x] Feature importance -> requires target variable (only run if set) @Lilly-May (#690)
- [x] `generate_bias_report` that runs a lot of methods subsequently and summarizes things?!? @Lilly-May (#690)

Lilly-May:
Regarding propensity score matching, we already have that implemented for causal inference with dowhy [here](https://github.com/theislab/ehrapy/blob/1b5165d925f87304948fffe9b76cc68669ee6ab2/ehrapy/tools/causal/_dowhy.py#L37). If we implement a causal discovery module (#619), implementing a propensity score matching algorithm would be important, but until then, I don't really see the need. However, please feel free to disagree!
Should we close this issue then?

Zethson:
Yup! Let's close

May 4, 2024 12:18:43 PM Lilly May ***@***.***>:

Issue #637, opened 2024-01-15
Add var type retrospectively


Zethson:
### Description of feature

I just ran into the situation where I got an AnnData object which already had encoded data (label encoded) and now I have to manually create the `var` column types (`numeric`, `non-numeric`). I was wondering whether we should offer a convenience function here that will try to guess the column types retrospectively and annotate the var?

Basically just extracting this from `encode` and exposing it with the API

Zethson:
@Lilly-May also ran into this and suggests to validate whether these keys exist early enough and guide the user through fixing it.

Lilly-May:
I was just thinking about this a bit more and here's my suggestion on how to tackle this issue:

**Summary of the problems we're facing:**
- Already encoded data are missing the `adata.var["ehrapy_column_type"]` annotation that is created when calling `ep.pp.encode`, so one has to manually create the var annotation for downstream processing.
- The `adata.var["ehrapy_column_type"]` annotation stores whether a feature is numerical or not, but that doesn't always align perfectly with the continuous/categorical annotation. For instance, a disease flag might be represented as 0 or 1, treated as numerical data, but conceptually it remains categorical. This distinction is important e.g. for downstream bias detection and feature importance calculations.

**Suggestion on how to tackle this:**
We create a new method `ep.pp.feature_specifications` (or `feature_datatypes`?) that stores either the label `continuous` or the label `categorical` for each feature. The method attempts to infer this information: If ehrapy datatypes (`adata.var["ehrapy_column_type"]`, created during encoding) are presented, we can infer some labels from there. Otherwise, in general, numerical implies continuous features, except for some special cases we test for (such as only 0 and 1 values are present, then we save it as categorical). At the end, we print a table with each feature and the detected type. Users must then check the datatypes, and if something is wrong, correct it.

**Discussion points:**
1. How do we name that `var` annotation? We already have `ehrapy_column_type` for the annotation from encoding, but I would consider changing that to something like `encoding_feature_type` and then adding the new continuous/categorical annotation as something else (just `feature_type`?).
[...]

PR #690, opened 2024-04-13
Add bias detection to preprocessing

Lilly-May:
**PR Checklist**

- [X] This comment contains a description of changes (with reason)
- [X] Referenced issue is linked (#647)
- [x]  If you've fixed a bug or added code that should be tested, add tests!
- [x]  Documentation in `docs` is updated

**Description of changes**

This is a very drafty PR to get the discussion on the bias module started. The general idea is to have one method that calls several submethods:

- Pairwise correlations between features
- Standardized mean differences for all features between groups in sensitive features
- Feature importances (Is a sensitive feature an important predictor of another feature?)

I thought the usage to be something along the lines of:

```
# adata is expected to be encoded
ep.pp.bias_detection(adata, sensitive_features=["ethnicity", "sex"])
```

I also made some necessary adjustments on the feature importances method. Firstly, I changed the default model from `regression` to `rf` (Random Forest) as I found this to be a lot more reliable when testing it for the bias detection, so I believe this should be the default. I also added two parameters to include the options to not log during the calculation of feature importances and to return the prediction score (R2 or accuracy).

**Discussion points**

1. Do we make the submethods (`_feature_correlations` and `_standardized_mean_differences`) public as well? If not, should I move everything into one big method since otherwise the submethods (e.g., the correlations one) are quite short, and also we would regenerate a pandas DataFrame from the AnnData several times.
2. It would be great to offer a more explorative way to detect potentially unknown sensitive features and biases in the data. Basically, we would have the option `sensitive_features="all"`. However, I think we can't run the feature importances then because it gets too computationally expensive. Maybe we should add a parameter `run_feature_importances` that is set to `False` by default when `sensitive_features="all"` and set to `True` by default when `sensitive_features` is a list of features?
4. How de we call the big method computing all different measurements? `generate_bias_report`? `bias_detection`?
5. How do we save the results in the AnnData? We will likely need them for downstream plot generation. But for example, for the standardized mean difference, we have one `n_groups_in_feature x var_n matrix` per investigated sensitive feature, so we would have one entry in `varm` per sensitive feature. Do we want that? Or do we only save those for which we find potential biases?
6. How do we show the output? Should we print out a table (similar to scCODA in pertpy)? We could also (but this might be a weird idea) think if it would be possible to generate an (additional) HTML report which already includes plots. But that would be quite different from the rest of the API, so I don't know if we want to go for that.

**ToDos**
- [x] Resolve ToDos in the code
- [x] Look into and adjust the code with respect to differences for continuous and categorical features (waiting on #637 for that)
- [x] Add tests
- [x] Add examples
- [x] Save standardized mean differences in `uns` subdict
- [x] Raise error if data are not already encoded
- [x] Introduce `copy` parameter
- [x] Fix test warnings


Zethson:
Thank you so much!


>  Do we make the submethods (_feature_correlations and _standardized_mean_differences) public as well? If not, should I move everything into one big method since otherwise the submethods (e.g., the correlations one) are quite short, and also we would regenerate a pandas DataFrame from the AnnData several times.

If all of these are fast and don't have too many configurable parameters, it's easier to have a one-stop function. There are ways to circumvent the need to regenerate the Pandas DataFrame such as class variables, cached properties, and other ideas but let's assume that we don't need that for now.

>  It would be great to offer a more explorative way to detect potentially unknown sensitive features and biases in the data. Basically, we would have the option sensitive_features="all". However, I think we can't run the feature importances then because it gets too computationally expensive. Maybe we should add a parameter run_feature_importances that is set to False by default when sensitive_features="all" and set to True by default when sensitive_features is a list of features?

Yeah, that sounds fair to me. Ideally, the fewer parameters the better. That's a rule for every function that we implement. Plotting functions can be exceptions.

>  How de we call the big method computing all different measurements? generate_bias_report? bias_detection?

`detect_bias`? When I hear `report` I always think of an actual HTML/PDF report...

> How do we save the results in the AnnData? We will likely need them for downstream plot generation. But for example, for the standardized mean difference, we have one n_groups_in_feature x var_n matrix per investigated sensitive feature, so we would have one entry in varm per sensitive feature. Do we want that? Or do we only save those for which we find potential biases?

I'd save all of them and not just the ones that we find potential biases for. We just calculate and the interpretation/usage is up to the user (we provide the tools to do so). So yes `varm`.

> How do we show the output? Should we print out a table (similar to scCODA in pertpy)? We could also (but this might be a weird idea) think if it would be possible to generate an (additional) HTML report which already includes plots. But that would be quite different from the rest of the API, so I don't know if we want to go for that.

I veto HTML reports because they're annoying to interpret and it'd be much cooler to generate whole reports for this including the cohort tracking etc in the future. This would be an entirely different subproject.
One learning of mine is that Rich tables are nice, but sometimes troublesome and cannot be resized. Therefore, I always prefer DataFrames over Rich tables and plots over no plots. Ideally, this function could return a summary table and store extensive results in the `varm` slots?

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+import pandas as pd
+from anndata import AnnData
+
+from ehrapy import logging as logg
+from ehrapy.anndata import anndata_to_df
+
+
+def bias_detection(
+    adata: AnnData,
+    sensitive_features: Iterable[str] | Literal["all"],
```
Zethson, line 14: Consider adding a `*` after `sensitive_features` to force the rest to be keyword arguments

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+    corr_threshold: float = 0.5,
+    smd_threshold: float = 0.5,
+    feature_importance_threshold: float = 0.1,
+    prediction_confidence_threshold: float = 0.5,
+    corr_method: Literal["pearson", "spearman"] = "spearman",
+):
+    """Detects bias in the data.
```
Zethson, line 21: Eventually requires a bit more detailed explanation in the docstring :)

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Args:
+        adata: An annotated data matrix containing patient data.
```
Zethson, line 24:
```suggestion
        adata: An annotated data matrix containing EHR data.
```

I know that we have plenty of variants of that in the code and they're all equally bad, but I want to avoid `patient` because it implies a certain structure of the observations (patient vs patient visit view and other things)

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+        sensitive_features: A list of sensitive features to check for bias.
```
Zethson, line 25:
```suggestion
        sensitive_features: A list of sensitive features to check for bias.
```

You've now accidentally typed `sensitive_features` differently by implying that it's a `list` :) Please reword it slightly. Also consider documenting the `all` here.

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+        corr_threshold: The threshold for the correlation coefficient between two features to be considered of interest. Defaults to 0.5.
+        smd_threshold: The threshold for the standardized mean difference between two features to be considered of interest. Defaults to 0.5.
+        feature_importance_threshold: The threshold for the feature importance of a sensitive feature for predicting another feature to be considered
+            of interest. Defaults to 0.1.
+        prediction_confidence_threshold: The threshold for the prediction confidence (R2 or accuracy) of a sensitive feature for predicting another
+            feature to be considered of interest. Defaults to 0.5.
+        corr_method: The correlation method to use. Choose between "pearson" and "spearman". Defaults to "spearman".
+    """
+    from ehrapy.tools import rank_features_supervised
```
Zethson, line 34: Glad to see the scoped import working. Prettier than the isort exception.

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        for feature in sensitive_features:
+            feature_importance = adata.var[f"{prediction_feature}_feature_importances"][feature] / 100
+            if feature_importance > feature_importance_threshold and prediction_score > prediction_confidence_threshold:
+                logg.warning(
+                    f"Feature {feature} has a high feature importance for predicting {prediction_feature} (importance in %: {feature_importance:.3f}, prediction score: {prediction_score:.3f})."
+                )
+
+
+def _feature_correlations(adata: AnnData, method: Literal["pearson", "spearman"] = "spearman"):
```
Zethson, line 77: Misses return type

Zethson: Generally, I've learn that it's fine to keep the docstrings more minimal for such simple functions. If the function name is self explanatory (which it should be here!) we can even omit the whole docstring.

But in any case, you might even integrate this into the bigger function anyways.

ehrapy/preprocessing/_imputation.py
```diff
@@ -208,7 +208,7 @@ def knn_impute(
     imputation ran successfully.
 
     Args:
-        adata: An annotated data matrix containing gene expression values.
+        adata: An annotated data matrix containing patient data.
```
Zethson, line 211: OOoff, Thanks for catching that haha. Same comment about `patient` as above.

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+    corr_matrix = anndata_to_df(adata).corr(method=method)
+    return corr_matrix
+
+
+def _standardized_mean_differences(adata: AnnData, features: Iterable[str]) -> dict:
+    """Computes the standardized mean differences between sensitive features.
+
+    Args:
+        adata: An annotated data matrix containing patient data.
+        features: A list of features to compute the standardized mean differences (SMD) for. For each listed feature, the SMD is computed for each
```
Zethson, line 96: Careful with `list`

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+        method: The correlation method to use. Choose between "pearson" and "spearman". Defaults to "spearman".
+
+    Returns:
+        A pandas DataFrame containing the correlation matrix.
+    """
+    corr_matrix = anndata_to_df(adata).corr(method=method)
+    return corr_matrix
+
+
+def _standardized_mean_differences(adata: AnnData, features: Iterable[str]) -> dict:
```
Zethson, line 91: Can you type the dictionary with more details, please?

Zethson (commented):
Generally, does this function need to differentiate a bit more between the different feature types that we're also discussing?

Overall, I think that this is quite good. We'll see whether we can integrate more of the other functions into this, but I think that this function is going into the right direction for sure.

Lilly-May:
Thanks for the review @Zethson!

I'll summarize the main things I'm going to change here:

- [ ] Put everything into one method
- [ ] Introduce `sensitive_features="all"` option
- [ ] Rename method to `detect_bias`
- [ ] Save all calculations in the adata
- [ ] Return a summary table with detected biases exceeding the respective threshold specified as parameter (I have to think a bit more about how I could summarize everything in ideally one pandas dataframe)


One more thing to think about: Do we also want to offer plots in some way? Would the summary dataframe be the input, or should it work with the values stored in the adata without additional user input?

Zethson:
> One more thing to think about: Do we also want to offer plots in some way? Would the summary dataframe be the input, or should it work with the values stored in the adata without additional user input?

I don't think that we need to offer more plots. I'd much rather consider integrating this with @eroell cohort tracking or TableOne down the road.

ehrapy/preprocessing/_bias.py
Lilly-May: Yeah, I'll integrate it into the bigger function now so I'll delete this code section.

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if sensitive_features == "all":
+        sensitive_features = adata.var_names
+
+    correlations = _feature_correlations(adata, method=corr_method)
+    adata.varp["correlation"] = correlations
+
+    for feature in sensitive_features:
+        for comp_feature in adata.var_names:
+            if correlations.loc[feature, comp_feature] > corr_threshold:
```
eroell, line 44: should also test for < - corr_threshold to take into account negative correlation I think? :)

Lilly-May: True, thanks for catching that!

eroell:
Cool!

Thoughts on the implementing side of things:

> Maybe we should add a parameter run_feature_importances that is set to False by default when sensitive_features="all" and set to True by default when sensitive_features is a list of features?

This function will likely be very verbose with many arguments. It is probably easier to have argument defaults consistent, and not interacting with each other? :)



> How do we save the results in the AnnData? We will likely need them for downstream plot generation. But for example, for the standardized mean difference, we have one n_groups_in_feature x var_n matrix per investigated sensitive feature, so we would have one entry in varm per sensitive feature. Do we want that? Or do we only save those for which we find potential biases?

> I'd save all of them and not just the ones that we find potential biases for. We just calculate and the interpretation/usage is up to the user (we provide the tools to do so). So yes varm.

Hm. A bit more involved, no?
The correlation is a n_var x n_var matrix; but only if we compute a correlation also between (categorical, categorical) and (categorical, numerical), for which we have not paved the way yet right

The feature importance is part of .var if I got this right, since calling a "native" ehrapy function this would be naturally accounted for indeed :)

The standardized mean difference: this is quite a lot going into the direction of rank_features_groups - there, we'd even have things for categoricals/continuous stuff already, and cool test statistics, and so on. Do we want yet another one here? If smd is important, it maybe even should be in rank_features_groups, and from there be called here I think
But to not slow things here down too much, we can also go ahead keeping it here as you suggested - and take care of fancier things later :)

Zethson:
@eroell you are correct :) Thanks!

ehrapy/preprocessing/_bias.py
Lilly-May: "I removed the standalone method and incorporated the code into the larger method. However, I specified the returned dictionary within the larger method.

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+    """Computes pairwise correlations between features in the AnnData object.
+
+    Args:
+        adata: An annotated data matrix containing patient data.
+        method: The correlation method to use. Choose between "pearson" and "spearman". Defaults to "spearman".
+
+    Returns:
+        A pandas DataFrame containing the correlation matrix.
+    """
+    corr_matrix = anndata_to_df(adata).corr(method=method)
```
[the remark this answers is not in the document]
Lilly-May: When using Spearman's rank correlation coefficient (currently the default), we can calculate it between categorical features as well. While it's probably not be the best statistic for identifying correlations between categorical features, it still gives an overview over potential correlations and biases within categorical and, importantly, also between categorical and continuous features.
Personally, I would keep the categorical features in the correlation matrix. We should still look for alternative statistics better suited for detecting differences in categorical features, though! However, I also think this issue would be at least partially addressed by looking at value counts for categorical features as discussed earlier (e.g., Group a of Feature A is always in Group b of Feature B).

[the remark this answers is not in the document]
eroell: Rank correlation only works for ordinal categorical features, and not for nominal categorical variables though..

Cramer's V would likely be what we're looking for here for pairwise categorical-categorical cases of nominal categorical variables: this is the statistic you'd be looking for with your correct intuition for needing value counts by subgroups. A thought here; this does take only values from 0 to 1, and hence might be a bit odd to put into a table with Pearson/spearman Rank Correlation coefficients with values from -1 to 1...

There are multiple cases to take into account though as far as I see, with suggestions in brackets meaning I am not sure and could check out deeper:
- Cont - Cont: Pearson or Spearman
- Categorical Ordinal - Categorical Ordinal: Spearman
- Categorical Nominal - Categorical Nominal: Cramer's V
- Cont - Categorical Nominal: [anything rank_features_groups, ANOVA?]
- Cont - Categorical Ordinal: Spearman
- Categorical Ordinal - categorical Nominal: [anything rank_features_groups, ANOVA?]

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,203 @@
...
+import numpy as np
+import pandas as pd
+from anndata import AnnData
+
+from ehrapy.anndata import anndata_to_df
+
+
+def detect_bias(
+    adata: AnnData,
+    sensitive_features: Iterable[str] | np.ndarray | Literal["all"],
```
Zethson, line 14: Pretty sure that numpy arrays are also of type Iterable?

ehrapy/preprocessing/_bias.py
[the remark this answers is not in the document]
Lilly-May: I've thought about this a bit more, and I think it would be great to use different measurements for the various feature types. However, currently, we don't differentiate between nominal and ordinal categorical features and I'm uncertain if we should incorporate this solely for calculating feature correlations. But maybe there are/will be other use-cases in ehrapy for this? I'm open to adding categorical-ordinal and categorical-nominal feature types as part of addressing issue #701, but I'd like to hear your opinions on whether that makes sense, as it would require the user to carefully review the feature types detected automatically by `ep.ad.infer_feature_types`.

If we choose not to distinguish further between ordinal and nominal categorical features, the best solution I can think of is to use Pearson/Spearman when at least one continuous feature is involved and Cramer's V when we have two categorical features.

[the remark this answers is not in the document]
Zethson: > However, currently, we don't differentiate between nominal and ordinal categorical features and I'm uncertain if we should incorporate this solely for calculating feature correlations.

I guess because the G-test doesn't care that much whether the categorical data is nominal or ordinal.

> I'm open to adding categorical-ordinal and categorical-nominal feature types as part of addressing issue https://github.com/theislab/ehrapy/issues/701, but I'd like to hear your opinions on whether that makes sense, as it would require the user to carefully review the feature types detected automatically by ep.ad.infer_feature_types.

It should be reasonably easy to detect ordinal categorical data, because that's pretty much what label encoded data looks like anyways. The common case of "0" and "1" would probably be nominal categorical. It's not easy to detect other forms of "nominal categorical" I assume, but that should be fine in most cases.

> if we should incorporate this solely for calculating feature correlations

Since we don't reaaally care about it for G-tests, I'm not sure whether we would make use of it for more things (for now) but I always like being future proof if possible.

TLDR:
1. Can we really detect nominal categorical well? Do we want to put another burden on the user to check this? They'd probably have to label those anyways, because we can't easily detect them (besides 0/1s) and at that point they could label them properly.
2. It doesn't like it's toooo much work to also support storing ordinal vs nominal. I enjoy being future proof.
3. It will increase the code complexity by a non-significant amount because we'd always have to grab both categorical keys/types if we want to apply something to only categorical.

I lean more towards supporting it than not.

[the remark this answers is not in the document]
Lilly-May: Just wanted to summarize what I ended up doing to sum up offline and online discussions. As of now, the feature correlations would be the only application in ehrapy where we would need the differentiation between nominal and ordinal features. Because, as soon as data are encoded, detecting this difference automatically is nearly impossible, we would add quite some effort for the user, as they would have to manually declare what features are ordinal and what nominal just to compute the feature correlations using the optimal method (Spearman CC vs. Cramer's V, for instance).

Additionally, just computing Spearman/Pearson CC for all features won't show any correlations that aren't there, but just some correlations between categorical features won't be revealed. However, those should then be detected by the feature importances calculation.

Hence, we decided to stick to Pearson/Spearman CC for all features as of now. If in the future the differentiation between ordinal and nominal categorical features becomes important at other places in ehrapy, it would be easy to adapt the bias detection method accordingly.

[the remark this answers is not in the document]
Zethson: Thanks for the summary! I would kindly suggest that you create a new issue for the whole topic of "ordinal vs nominal categorical recording" where you paste your well formulated comment. I always search issues for context, but rarely PRs.

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,119 @@
...
+            feature to be considered of interest. Defaults to 0.5.
+        corr_method: The correlation method to use. Choose between "pearson" and "spearman". Defaults to "spearman".
+    """
+    from ehrapy.tools import rank_features_supervised
+
+    if sensitive_features == "all":
+        sensitive_features = adata.var_names
+
+    correlations = _feature_correlations(adata, method=corr_method)
+    adata.varp["correlation"] = correlations
```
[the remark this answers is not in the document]
Lilly-May: I ended up storing all results in the `adata` (either in place if `copy=False` or in a copied `adata`), and I return a dictionary with a dataframe for each investigation, storing just the results of interest, i.e., those that exceed the thresholds specified in the parameters.
What do you think about that solution @eroell?

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,244 @@
...
+    feature_importance_threshold: float = 0.1,
+    prediction_confidence_threshold: float = 0.5,
+    corr_method: Literal["pearson", "spearman"] = "spearman",
+    copy: bool = False,
+) -> dict[str, pd.DataFrame] | tuple[dict[str, pd.DataFrame], AnnData]:
+    """Detects biases in the data using feature correlations, standardized mean differences, and feature importances.
+
+    Detects biases with respect to sensitive features, which can be either a specified subset of features or all features in adata.var.
+    The method detects biases by computing:
+    - pairwise correlations between features
```
Zethson, line 31: The bullet points don't seem to render properly:

https://ehrapy--690.org.readthedocs.build/en/690/usage/preprocessing/ehrapy.preprocessing.detect_bias.html

You might need more line breaks or something

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,244 @@
...
+    - standardized mean differences for numeric features between groups of sensitive features
+    - value counts of categorical features between groups of sensitive features
+    - feature importances for predicting one feature with another.
+    Results of the computations are stored in var, varp, and uns of the adata object.
+    Values that exceed the specified thresholds are considered of interest and returned in the results dictionary.
+
+    Args:
+        adata: An annotated data matrix containing EHR data.
+        sensitive_features: Sensitive features to consider for bias detection. If set to "all", all features in adata.var will be considered.
+            If only a subset of features should be considered, provide as an iterable.
```
Zethson, line 41:
```suggestion
```

I think we can get away with removing that because the type hints should imply that.

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,244 @@
...
+            sensitive_features is not set to "all", as this can be computationally expensive. Defaults to None.
+        corr_threshold: The threshold for the correlation coefficient between two features to be considered of interest. Defaults to 0.5.
+        smd_threshold: The threshold for the standardized mean difference between two features to be considered of interest. Defaults to 0.5.
+        categorical_factor_threshold: The threshold for the factor between the value counts (as percentages) of a feature compared between two
+            groups of a sensitive feature. Defaults to 2.
+        feature_importance_threshold: The threshold for the feature importance of a sensitive feature for predicting another feature to be considered
+            of interest. Defaults to 0.1.
+        prediction_confidence_threshold: The threshold for the prediction confidence (R2 or accuracy) of a sensitive feature for predicting another
+            feature to be considered of interest. Defaults to 0.5.
+        corr_method: The correlation method to use. Choose between "pearson" and "spearman". Defaults to "spearman".
```
Zethson, line 52:
```suggestion
        corr_method: The correlation method to use. Defaults to "spearman".
```
I double checked it, the Literal type hints already say that

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,244 @@
...
+            threshold.
+        - "feature_importances": Feature importances for predicting one feature with another that exceed the feature importance and prediction
+            confidence thresholds.
+
+        If copy is set to True, the function returns a tuple with the results dictionary and the updated adata.
+
+    Examples:
+        >>> import ehrapy as ep
+        >>> adata = ep.dt.mimic_2(encoded=True)
+        >>> ep.ad.infer_feature_types(adata, output=None)
```
Zethson, line 70: We have that in every example now and if we don't fully automate it, it might also be good to just add that to the dataloaders that we maintain (not for this PR but something that we should do in a follow up PR)

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,244 @@
...
+import numpy as np
+import pandas as pd
+from anndata import AnnData
+
+from ehrapy.anndata import anndata_to_df, check_feature_types
+from ehrapy.anndata._constants import CATEGORICAL_TAG, CONTINUOUS_TAG, DATE_TAG, FEATURE_TYPE_KEY
+
+
+@check_feature_types
+def detect_bias(
```
Zethson, line 14: Do we want a `layer` parameter? I'm 50/50 but it usually doesn't hurt

ehrapy/tools/feature_ranking/_feature_importances.py
```diff
@@ -47,8 +50,13 @@ def rank_features_supervised(
             for each feature individually. Defaults to 'standard'.
         percent_output: Set to True to output the feature importances as percentages. Note that information about positive or negative
             coefficients for regression models will be lost. Defaults to False.
+        logging: Set to False to disable logging. Defaults to True.
```
Zethson, line 53:
```suggestion
        verbose: Set to False to disable logging. Defaults to True.
```
Is what we use in those cases usually

Zethson (approved):
Lovely! A few minor comments. Thank you very much!

ehrapy/preprocessing/_bias.py
Lilly-May: I added a `layer` parameter - also think it doesn't hurt

ehrapy/preprocessing/_bias.py
Lilly-May: Thanks for spotting that - it's fixed now.

tests/tools/feature_ranking/test_feature_importances.py
```diff
@@ -16,7 +16,7 @@ def test_continuous_prediction():
     adata.var[FEATURE_TYPE_KEY] = [CONTINUOUS_TAG] * 3
 
     for model in ["regression", "svm", "rf"]:
-        rank_features_supervised(adata, "target", model, "all")
+        rank_features_supervised(adata, "target", model=model, input_features="all")
```
eroell, line 19:
```suggestion
        try:
        rank_features_supervised(adata, "target", model=model, input_features="all")
```

ehrapy/preprocessing/_bias.py
```diff
@@ -0,0 +1,247 @@
...
+    # -------------------
+    if run_feature_importances:
+        feature_importances_results: dict[str, list] = {
+            "Sensitive Feature": [],
+            "Predicted Feature": [],
+            "Feature Importance": [],
+            "Prediction Score": [],
+        }
+        for prediction_feature in adata.var_names:
+            prediction_score = rank_features_supervised(
```
eroell, line 218:
```suggestion
            try:
                prediction_score = rank_features_supervised(
```

PR #706, opened 2024-04-25
Enhancement/knni amendments

eroell:
**PR Checklist**

-   [ ] Referenced issue is linked n/a

**Description of changes**

Two suggested changes:
- check for valid backend: it can happen quickly to mispell an argument as long as 'scikit-learn', and then the default being used would be hidden (not speaking of own mishap, checking for a friend)
- omit the `**kwargs` collector: I know... I just am not a fan of grabbing everything that might be a misspelt argument and throw it further down :)

**Technical details**

**Additional context**


Zethson (approved):
Lovely! <3

PR #716, opened 2024-05-06
Feature scaling on training set when computing feature importances

Lilly-May:
**Description of changes**

As @eroell kindly pointed out, features in `ep.tl.feature_importances` were previously scaled based on the data from both the training and test sets. I changed this so that the scaler is now fitted exclusively on the training data and then used to transform the data of both the training and test sets.

### DIFF
--- a/ehrapy/tools/feature_ranking/_feature_importances.py
+++ b/ehrapy/tools/feature_ranking/_feature_importances.py
@@ -120,21 +120,25 @@ def rank_features_supervised(
     input_data = data[input_features]
     labels = data[predicted_feature]
 
+    x_train, x_test, y_train, y_test = train_test_split(input_data, labels, test_size=test_split_size)
+
     for feature in input_data.columns:
         try:
-            input_data.loc[:, feature] = input_data[feature].astype(np.float32)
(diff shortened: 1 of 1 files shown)


Zethson (approved):
LGTM :) Thanks!

ehrapy/tools/feature_ranking/_feature_importances.py
```diff
@@ -120,22 +120,26 @@ def rank_features_supervised(
     input_data = data[input_features]
     labels = data[predicted_feature]
 
+    x_train, x_test, y_train, y_test = train_test_split(input_data, labels, test_size=test_split_size)
```
eroell, line 123: sorry for hopping in - what do you think about adding a random_state here? :)

eroell, line 123:
```suggestion
    x_train, x_test, y_train, y_test = train_test_split(input_data, labels, test_size=test_split_size, random_state=42)
    
```

Lilly-May: Good idea, thanks!

PR #724, opened 2024-05-12
Unify feature type detection

### LINKED ISSUE #701: Harmonize feature type detection
### Description of feature

We currently have three issues (#637, #649, #662) dealing with how we determine feature types and the downstream problems that arise due to different approaches. I'll provide a summary of these issues and associated ToDos here, while closing the other three issues.

**Problem:**
We've introduced PR #697 to accurately determine feature types in ehrapy. With the new method `ep.ad.infer_feature_types`, feature types are guessed based on predefined rules and we prompt the user to review these annotations. Currently, feature determination occurs at multiple stages inconsistently and is saved in the adata at several places. Ideally, we would harmonize ehrapy to use exclusively use `ep.ad.infer_feature_types` for feature annotation, eliminating guesswork in downstream analyses. This means that the new method would be part of the standard preprocessing steps.

**ToDos:**
- [x] Update encoding: The `autodetect` option of encoding [here](https://github.com/theislab/ehrapy/blob/82be78ebbc986c940ee0801f5bbda39c33eaf6ff/ehrapy/preprocessing/_encoding.py#L84) relies on the `adata.var[EHRAPY_TYPE_KEY]` tag, which is set when (1) reading a dataframe [here](https://github.com/theislab/ehrapy/blob/82be78ebbc986c940ee0801f5bbda39c33eaf6ff/ehrapy/anndata/anndata_ext.py#L91) or (2) moving something from obs to X [here](https://github.com/theislab/ehrapy/blob/82be78ebbc986c940ee0801f5bbda39c33eaf6ff/ehrapy/anndata/anndata_ext.py#L602). Ideally, we would get rid of the `adata.var[EHRAPY_TYPE_KEY]` annotation entirely and, if `autodetect` is set to `True` in the encoding method, base the identification of features to encode on the annotation from `ep.ad.infer_feature_types`.
- [x] Ensure that the feature type annotations are not stored elsewhere. For example, during encoding, we store the information in `adata.uns["var_to_encoding"]` and `adata.uns["encoding_to_var"]`.
- [x] Update [CohortTracker](https://github.com/theislab/ehrapy/blob/4090f6fca1265bfc1431a79a66bfd79b89956611/ehrapy/tools/cohort_tracking/_cohort_tracker.py#L27) to be based on `ep.ad.infer_feature_types`.
- [x] Update `ep.tl.rank_features_groups` to be based on `ep.ad.infer_feature_types`.
- [x] Update all methods in adata_ext to be based on `ep.ad.infer_feature_types`.
- [x] Ensure functionality for both encoded and unencoded data, so it can be used when loading an encoded dataset (I don't know why this wouldn't be the case, still I want to double-check).
- [x] Improve date detection in `ep.ad.infer_feature_types` to also work with date(time)s stored as strings and update the FHIR tutorial accordingly.
- [X] Decide how to deal with ordinal categorical features: Should we have explicit annotations for ordinal categorical data? In downstream analyses, these could either be treated as a continuous integer-scaled feature, or as a nominal class feature, depending on the context and analysis. (see #713)
- [x] Look into whether we can combine this with the feature type detection of TableOne [here](https://github.com/tompollard/tableone/blob/bfd6fbaa4ed3e9f59e1a75191c6296a2a80ccc64/tableone/tableone.py#L555)
- [x] Discuss whether we want to automatically incorporate the feature type detection into the dataset-loaders (i.e. without the user specifically calling it, which would be required to maintain the encode parameter for the data loaders)
- [x] Take care of `df_to_anndata`, which also does a lot of type inference.
- [x] If we don't fully automate the feature type detection, it might be good to add that to the dataloaders that we maintain

Lilly-May:
**PR Checklist**

-   [X] Referenced issue is linked (closes #701)
-   [X] Documentation in `docs` is updated

**Description of changes**

- `ehrapy` now fully relies on `ep.ad.infer_feature_types` whenever a distinction between categorical and numerical features is needed.
- The feature types are inferred automatically when not done manually by the user.
- Feature/Encoding types stored previously in `adata.var["ehrapy_column_type"]` were removed.
- Improved date detection in the `ep.ad.infer_feature_types` method: All dates stored in any [ISO-format](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse) as a String are not automatically detected as dates.

**Discussion points**

1. The encoding method stores data in `adata.uns`. With the new feature type detection, we could remove that now. The only reason why I kept it is because it is used to find out if the `adata` is already encoded or not, and, depending on that, the behavior of the encoding method changes. I could manually test that instead by just checking that all features are numerical. However, that would be more computationally expensive than simply checking if a specific key in `adata.uns` is present.
2. When loading data via the dataloader, we have the option to automatically encode them (`encoded=True`). Because the encoding relies on feature types, those will automatically be inferred whenever data are loaded. I think automatically encoding them is a very convenient feature, so I guess we want to keep this behavior, but I still want to confirm that you are aware of and fine with this behavior.
3. Related to the point above, we could also pre-define the correct feature types for the dataset offered by ehrapy, which would be really convenient for the users but not so pleasant for us to do. Just wanted to raise this as an idea and hear what you think about that.
4. Should we offer a convenient function to correct feature type annotations? Something like `ep.ad.set_feature_type(["feature1", "feature2", "feature3"], "categorical")`, where three features that were detected incorrectly would be corrected to the feature type "categorical"?

**ToDos**

- [x] Add more specific warnings for tricky cases when inferring feature types
- [x] Resolve ToDos in the code
- [x] Don't save feature type information several times (specifically the encoding information in `adata.uns`, as described above)
- [x] Store everything encoding related currently stored in `adata.uns` in `adata.var` or remove it
- [x] Pre-safe all feature types in all ehrapy dataloader
- [x] Add method to switch feature types easily and reference it in the warning


tests/anndata/test_anndata_ext.py
```diff
@@ -393,17 +399,19 @@ def test_detect_mixed_binary_columns():
...
-            {EHRAPY_TYPE_KEY: [NUMERIC_TAG, NON_NUMERIC_TAG, NUMERIC_TAG]},
+            {FEATURE_TYPE_KEY: [NUMERIC_TAG, CATEGORICAL_TAG, CATEGORICAL_TAG]},
             index=["Col1", "Col2", "Col3"],
         ),
     )
 
 
 @pytest.fixture
-def adata_numeric():
+def adata_numeric():  # TODO: Delete: I don't think this is used
```
Lilly-May, line 414: I couldn't find any tests using this fixture while going through the tests. Are the tests missing, or should I simply remove the `adata_numeric` method?

Zethson: You may remove it if it's not being used :)

I'll overhaul all fixtures in a couple of weeks so that we have less and more useful ones that are centralized in conftest.py

ehrapy/preprocessing/_encoding.py
```diff
@@ -743,6 +672,7 @@ def _add_categoricals_to_obs(original: AnnData, new: AnnData, categorical_names:
     logger.info(f"The original categorical values `{categorical_names}` were added to obs.")
 
 
+# TODO: Can we delete this and just rely on what is in obs?
```
Lilly-May, line 675: This point is still open to discussion—the last thing that is saved in `adata.uns`. Currently, the original, unencoded columns are stored in `adata.uns`. That information is then used, for example, when reverting encoding. Theoretically, the unencoded columns should also be stored in `adata.obs`, so I assume it would be possible to change the encoding so that it doesn't rely on `adata.uns` at all. 
However, I feel like `obs` is modified by many functions, so I wonder if it's a bit unsafe to rely on the information being stored there until `unencode` is called. I would be curious to know what you @Zethson think - whether I should keep the code as it is or also try to change this last bit stored in `adata.uns`?

ehrapy/anndata/_feature_specifications.py
```diff
@@ -65,17 +129,41 @@ def infer_feature_types(adata: AnnData, layer: str | None = None, output: Litera
...
+            _self = adata
+            adata = args[0]
+            args = args[1:]
+
         if FEATURE_TYPE_KEY not in adata.var.keys():
-            raise ValueError("Feature types are not specified in adata.var. Please run `infer_feature_types` first.")
-        np.all(adata.var[FEATURE_TYPE_KEY].isin([CATEGORICAL_TAG, CONTINUOUS_TAG, DATE_TAG]))
+            infer_feature_types(adata, output=None)
+            logger.warning(
+                f"Feature types were inferred and stored in adata.var[{FEATURE_TYPE_KEY}]. Please verify and adjust if necessary using `ep.ad.correct_feature_types`."
```
Zethson, line 142: Do we need to tell people how to verify?

ehrapy/anndata/_feature_specifications.py
```diff
@@ -86,13 +174,55 @@ def feature_type_overview(adata: AnnData):
...
+                branch.add(f"{unencoded} ({adata.obs[unencoded].nunique()} categories); {enc_mode} encoded")
+
+    else:
+        for categorical in sorted(cat_features):
+            branch.add(f"{categorical} ({df.loc[:, categorical].nunique()} categories)")
 
     print(tree)
+
+
+def correct_feature_types(adata, features: list[str] | str, corrected_type: str):
```
Zethson, line 201: Should this be called `correct` or `replace`?

Should `features` be typed with `Iterable[str]`?

ehrapy/preprocessing/_encoding.py
```diff
@@ -144,30 +151,31 @@ def encode(
...
                 new_var = pd.DataFrame(index=encoded_var_names)
-                new_var[EHRAPY_TYPE_KEY] = adata.var[EHRAPY_TYPE_KEY].copy()
-                new_var.loc[new_var.index.str.contains("ehrapycat")] = NON_NUMERIC_ENCODED_TAG
-                if FEATURE_TYPE_KEY in adata.var.keys():
-                    new_var[FEATURE_TYPE_KEY] = adata.var[FEATURE_TYPE_KEY].copy()
-                    new_var.loc[new_var.index.str.contains("ehrapycat"), FEATURE_TYPE_KEY] = CATEGORICAL_TAG
+                new_var[FEATURE_TYPE_KEY] = adata.var[FEATURE_TYPE_KEY].copy()
+                new_var[FEATURE_TYPE_KEY].loc[new_var.index.str.contains("ehrapycat")] = CATEGORICAL_TAG
+
+                # store unencoded columns in var
```
Zethson, line 157:
```suggestion
```

the following line of code documents itself

ehrapy/preprocessing/_encoding.py
```diff
@@ -144,30 +151,31 @@ def encode(
...
+                new_var["unencoded_var_names"] = unencoded_var_names
+
+                # store encoding mode in var
```
Zethson, line 160:
```suggestion
```

ehrapy/preprocessing/_encoding.py
```diff
@@ -247,22 +250,30 @@ def encode(
...
-            new_var[EHRAPY_TYPE_KEY] = adata.var[EHRAPY_TYPE_KEY].copy()
-            new_var.loc[new_var.index.str.contains("ehrapycat")] = NON_NUMERIC_ENCODED_TAG
-            if FEATURE_TYPE_KEY in adata.var.keys():
-                new_var[FEATURE_TYPE_KEY] = adata.var[FEATURE_TYPE_KEY].copy()
-                new_var.loc[new_var.index.str.contains("ehrapycat"), FEATURE_TYPE_KEY] = CATEGORICAL_TAG
+
+            new_var[FEATURE_TYPE_KEY] = adata.var[FEATURE_TYPE_KEY].copy()
+            new_var[FEATURE_TYPE_KEY].loc[new_var.index.str.contains("ehrapycat")] = CATEGORICAL_TAG
+
+            # store unencoded columns in var
```
Zethson, line 257:
```suggestion
```

ehrapy/preprocessing/_encoding.py
```diff
@@ -628,72 +567,30 @@ def _reorder_encodings(adata: AnnData, new_encodings: dict[str, list[list[str]]
...
 
-    Returns:
-        The updated encoding scheme
-    """
-    updated_encodings = defaultdict(list)  # type: ignore
-    for k, v in chain(new_encodings.items(), filtered_old_encodings.items()):
-        updated_encodings[k] += v
+        encodings[encode_mode] = new_encodings[encode_mode] if encode_mode in new_encodings.keys() else []
+        # add all columns that were encoded with the current mode before and are not reencoded
+        encodings[encode_mode] += [cat for cat in encoded_categoricals_with_mode if cat not in latest_encoded_columns]
```
Zethson, line 588: image

tests/tools/feature_ranking/test_rank_features_groups.py
```diff
@@ -287,6 +293,7 @@ def test_rank_adata_immutability_property(self, field_to_rank):
         adata = read_csv(
             dataset_path=f"{_TEST_PATH}/dataset1.csv", columns_x_only=["station", "sys_bp_entry", "dia_bp_entry"]
         )
+        ep.ad.infer_feature_types(adata, output=None)
```
Zethson, line 296: Why do we always need the explicit calls here?

Zethson (commented):
Thank you so so much for your hard work on this!

> undo_encoding

Does this also need to be removed from the docs?

For the release notes, it'd be great if we had a function which takes the old feature annotations and moves them to the right places.  We would tell people to run that when moving versions. Is this even necessary when it will run our new functions anyways? It might just reannotate the features with the right slots then, right?

ehrapy/anndata/_feature_specifications.py
```diff
@@ -65,17 +129,41 @@ def infer_feature_types(adata: AnnData, layer: str | None = None, output: Litera
...
+            args = args[1:]
+
         if FEATURE_TYPE_KEY not in adata.var.keys():
-            raise ValueError("Feature types are not specified in adata.var. Please run `infer_feature_types` first.")
-        np.all(adata.var[FEATURE_TYPE_KEY].isin([CATEGORICAL_TAG, CONTINUOUS_TAG, DATE_TAG]))
+            infer_feature_types(adata, output=None)
+            logger.warning(
+                f"Feature types were inferred and stored in adata.var[{FEATURE_TYPE_KEY}]. Please verify and adjust if necessary using `ep.ad.correct_feature_types`."
+            )
+        np.all(adata.var[FEATURE_TYPE_KEY].isin([CATEGORICAL_TAG, NUMERIC_TAG, DATE_TAG]))
```
eroell, line 144: this is not doing much, right? :)

ehrapy/preprocessing/_encoding.py
```diff
@@ -85,29 +78,42 @@ def encode(
         if "original" not in adata.layers.keys():
             adata.layers["original"] = adata.X.copy()
 
-        # autodetect categorical values, which could lead to more categoricals
+        # autodetect categorical values based on feature types stored in adata.var[FEATURE_TYPE_KEY]
```
eroell, line 81: in Line 71, instead of if isinstance(adata, AnnData): and then indenting the entire function, it'd be easier to keep track of the indentings when doing 

```py
if not isinstance(adata, AnnData):
    raise ValueError(f"Cannot encode object of type {type(adata)}. Can only encode AnnData objects!")
```

instead, then almost the entire function can be unindented.

Has been like this before, but might as well improve this here :)

ehrapy/anndata/_feature_specifications.py
eroell, line 144:
```suggestion
```

ehrapy/preprocessing/_encoding.py
```diff
@@ -85,29 +78,42 @@ def encode(
...
+                feat
+                for feat in categoricals_names
+                if not np.all(df_adata[feat].apply(type).isin([int, float, complex]))
+            ]
 
             # no columns were detected, that would require an encoding (e.g. non-numerical columns)
             if not categoricals_names:
                 logger.warning("Detected no columns that need to be encoded. Leaving passed AnnData object unchanged.")
                 return adata
             # copy uns so it can be used in encoding process without mutating the original anndata object
```
eroell, line 107: the comment in L107 says "copy uns so it can be ...", now would be "copy obs so it can be ..." to avoid confusion

ehrapy/preprocessing/_encoding.py
```diff
@@ -502,94 +461,74 @@ def _update_encoded_data(
...
     Returns:
         A (partially) encoding reset AnnData object
     """
-    if "var_to_encoding" not in adata.uns.keys():
-        if not suppress_warning:
-            logger.warning("Calling undo_encoding on unencoded AnnData object.")
-        return None
+    if "encoding_mode" not in adata.var.keys():
+        if verbose:
+            logger.warning("Calling undo_encoding on unencoded AnnData object. Returning original AnnData object!")
```
eroell, line 504: If I got this right, `undo_encoding` is dropped: so this message might be a bit confusing;
can it be dropped at this point, as the only case _undo_encoding is called is in encoding, after checking exactly

```py
if "encoding_mode" not in adata.var.keys():
```

Lilly-May:
> _undo_encoding_ Does this also need to be removed from the docs?

I removed it from `usage.md` and looked for other usages. Is there another place I need to delete it from?

> For the release notes, it'd be great if we had a function which takes the old feature annotations and moves them to the right places. We would tell people to run that when moving versions. Is this even necessary when it will run our new functions anyways? It might just reannotate the features with the right slots then, right?

The changes shouldn't cause any function to fail, as `ep.ad.infer_feature_types` will run automatically when feature types are not present. The only noticeable change might be that certain features are detected differently now. For instance, previously, a column with 0/1 values would usually be detected as numeric, whereas now it is categorical. Consequently, analysis results _could_ differ with the new release. Still, I would vote against having a method that transfers the old feature type annotations, as they are not more reliable than the new ones. Also, several methods previously handled this differently and stored the annotations inconsistently, so there isn't really just _one_ old feature type annotation.

ehrapy/anndata/_feature_specifications.py
```diff
@@ -4,15 +4,75 @@
...
-from ehrapy.anndata.anndata_ext import anndata_to_df
+from ehrapy.anndata._constants import CATEGORICAL_TAG, DATE_TAG, FEATURE_TYPE_KEY, NUMERIC_TAG
 
 
-def infer_feature_types(adata: AnnData, layer: str | None = None, output: Literal["tree", "dataframe"] | None = "tree"):
+def _detect_feature_type(col: pd.Series, verbose: bool = True) -> str:
+    """Detect the feature type of a column in a pandas DataFrame.
+
+    Args:
+        col: A pandas Series representing a feature.
```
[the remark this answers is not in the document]
Lilly-May: I changed it to `The column to detect the feature type for.`

ehrapy/anndata/_feature_specifications.py
```diff
@@ -4,15 +4,75 @@
...
+        verbose: Whether to print warnings for uncertain feature types. Defaults to True.
+
+    Returns:
+        The detected feature type. One of 'date', 'categorical', or 'numeric'.
```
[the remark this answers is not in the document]
Lilly-May: True, I changed that!

ehrapy/anndata/_feature_specifications.py
```diff
@@ -4,15 +4,75 @@
...
+    if majority_type == pd.Timestamp:
+        return DATE_TAG
+
+    if majority_type == str:
+        try:
+            col.apply(isoparse)
+            return DATE_TAG
+        except ValueError:
+            try:
+                col = col.apply(int)  # Could be an encoded categorical or a numeric feature
```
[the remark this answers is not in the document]
eroell: If a feature has only NaN's, I think this should raise an Error and fail :)

ehrapy/anndata/_feature_specifications.py
```diff
@@ -4,15 +4,75 @@
...
+    Args:
+        col: A pandas Series representing a feature.
+        verbose: Whether to print warnings for uncertain feature types. Defaults to True.
+
+    Returns:
+        The detected feature type. One of 'date', 'categorical', or 'numeric'.
+    """
+    n_elements = len(col)
+    col = col.dropna()
+    if len(col) == 0:
```
eroell, line 27: if a feature has only nans, I think it should raise an error and fail :)

ehrapy/anndata/_feature_specifications.py
```diff
@@ -4,15 +4,75 @@
...
+                majority_type = int
+            except ValueError:
+                try:
+                    col = col.apply(float)  # Could be an encoded categorical or a numeric feature
+                    majority_type = float
+                except ValueError:
+                    # Features stored as Strings that cannot be converted to float are assumed to be categorical
+                    return CATEGORICAL_TAG
+
+    if majority_type not in [int, float, complex]:
```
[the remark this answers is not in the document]
Lilly-May: Yep! Now that I think about it, we don't really expect complex numbers in EHR data though, do we?🙈😂 I'll remove it...

ehrapy/anndata/_feature_specifications.py
```diff
@@ -4,15 +4,75 @@
...
+        (majority_type == int or (np.all(i.is_integer() for i in col)))
+        and (n_elements != col.nunique())
+        and (
+            (col.min() == 0 and np.all(np.sort(col.unique()) == np.arange(col.nunique())))
+            or (col.min() == 1 and np.all(np.sort(col.unique()) == np.arange(1, col.nunique() + 1)))
+        )
+    ):
+        if verbose:
+            logger.warning(
+                f"Feature {col.name} was detected as a categorical feature stored numerically. Please verify."
```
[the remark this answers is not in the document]
Lilly-May: I was thinking about that too... In general, I like the idea, but the problem is that this warning is raised a lot depending on the dataset (see attached screenshot). So, I'd like to keep the message as short as possible (in one line).

Alternatively, I'm considering if I could combine all warnings into one. For example, we could have one warning: "Features X, Y, and Z were detected as..." Maybe that would solve the problem? What do you think?

Bildschirmfoto 2024-05-18 um 10 35 20

ehrapy/anndata/_feature_specifications.py
Lilly-May: I think it can't hurt, so I added a reference to `ep.ad.feature_type_overview`.

ehrapy/anndata/_feature_specifications.py
Lilly-May: Hmm... I don't really have a preference for the naming. Do you have a preference, @eroell?

Regarding the typing: Absolutely true, I changed it.

ehrapy/anndata/_feature_specifications.py
```diff
@@ -27,32 +87,36 @@ def infer_feature_types(adata: AnnData, layer: str | None = None, output: Litera
...
-            feature_types[feature] = DATE_TAG
-        elif majority_type not in [int, float, complex]:
-            feature_types[feature] = CATEGORICAL_TAG
-        # Guess categorical if the feature is an integer and the values are 0/1 to n-1 with no gaps
-        elif np.all(i.is_integer() for i in col) and (
-            (col.min() == 0 and np.all(np.sort(col.unique()) == np.arange(col.nunique())))
-            or (col.min() == 1 and np.all(np.sort(col.unique()) == np.arange(1, col.nunique() + 1)))
+        if (
+            FEATURE_TYPE_KEY in adata.var.keys()
+            and feature in adata.var[FEATURE_TYPE_KEY]
```
eroell, line 105: Seem to not get it right away, what is checked here?
`feature` is in the index of `adata.var[FEATURE_TYPE_KEY]` (a pd.Series), and I think this always evaluate to `True`

ehrapy/preprocessing/_encoding.py
Lilly-May: Changed the variable name to `categorical`^^ No more cats...🥲

ehrapy/anndata/_feature_specifications.py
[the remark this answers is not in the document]
Zethson: Yes, that'd be much much better. That's a shit ton of warnings. And I hope that our data loader won't spam users with the warning :)

ehrapy/anndata/_feature_specifications.py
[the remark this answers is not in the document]
Zethson: Very likely not haha

Zethson:
> I removed it from usage.md and looked for other usages. Is there another place I need to delete it from?

No, I just probably missed it. Thanks!

OK concerning the "transfer method"

ehrapy/anndata/_feature_specifications.py
Lilly-May: Oops, true—thanks for catching that! It now actually raises a warning.

ehrapy/preprocessing/_encoding.py
Lilly-May: I changed that - thanks!

ehrapy/anndata/_feature_specifications.py
Lilly-May: I kept it, but I now raise a warning if the feature type is neither None, NaN, nor one of the three allowed types.

ehrapy/preprocessing/_encoding.py
Lilly-May: True, `encoding_mode` must be present in `adata.var` at this point. I removed it.

ehrapy/anndata/_feature_specifications.py
Lilly-May: Overall, I want to check if the feature has been previously annotated and, if so, keep that annotation. But you're completely right that the specific condition always evaluates to true, so I removed it.

ehrapy/anndata/_feature_specifications.py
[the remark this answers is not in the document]
Lilly-May: Alright then I'll change that. And nope, logging for the feature type inference is disabled for the dataloaders.😊

ehrapy/anndata/_feature_specifications.py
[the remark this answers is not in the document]
Lilly-May: I think I found a somehow smoother way to do this using `pd.to_numeric` - feel free to double-check :)

tests/tools/feature_ranking/test_rank_features_groups.py
Lilly-May: Good point - we don't. That still remained from when we didn't automatically call `ep.ad.infer_feature_types`. I removed it from all tests where it isn't necessary to explicitly call it.

ehrapy/anndata/_feature_specifications.py
```diff
@@ -4,15 +4,75 @@
...
+                    col = col.apply(float)  # Could be an encoded categorical or a numeric feature
+                    majority_type = float
+                except ValueError:
+                    # Features stored as Strings that cannot be converted to float are assumed to be categorical
+                    return CATEGORICAL_TAG
+
+    if majority_type not in [int, float, complex]:
+        return CATEGORICAL_TAG
+
+    # Guess categorical if the feature is an integer and the values are 0/1 to n-1/n with no gaps
```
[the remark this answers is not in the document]
Lilly-May: > I'd probably vote for having them as numerical features still as default, and asking user to change if they think it should be categories...

Totally see your point! I reviewed all the dataloaders offered by ehrapy and felt like those cases are a lot rarer than the ones where we have already encoded categoricals. Especially the case where a feature serves as a flag, i.e., just storing 0/1 is quite common. We could change it so that 0/1 features are detected as categoricals, and for the rest (numbers from 0 or 1 to an integer without gaps), they are annotated as numerical. However, I think those are exactly the 1% of cases we can't reliably detect the feature type for, because there will be features misdetected either way, and users simply have to provide that information themselves. What do you think @Zethson?


> Can it be problematic downstream to have "None" as encoding label for such categorical feature in adata.var?

I can't think of any case where this would be problematic. All downstream methods that work with the encoding, especially encoding itself if data is re-encoded, check if the encoding mode is `None` and then treat that as an unencoded categorical. We already have those cases as of now (before this PR), where the feature type was categorical but no encoding mode was stored in `uns`.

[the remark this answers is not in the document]
Zethson: TLDR:
Yes. I still think as you outlined @Lilly-May

ehrapy/anndata/_feature_specifications.py
Lilly-May: I now renamed `correct_feature_types` to `replace_feature_types`.

PR #778, opened 2024-07-17
Add rust to build chain

Zethson:
Hopefully fixes `uv`


eroell:
@flying-sheep mentioned with uv we might need to pin minimum versions of some dependencies.... I'll have a look at this soon

flying-sheep:
To elaborate: uv just has a slightly(?) different dependency resolution algorithm from pip. So the versions that get downloaded (potentially built) and then considered for dependency resolutions are different. So:

1. if a failure happens during dependency resolution, uv/pip might be trying to build some ancient version of something. If you use the tool in verbose mode you can catch that and maybe circumvent by setting minimum versions or a [constraint file](https://docs.astral.sh/uv/concepts/resolution/#dependency-constraints), e.g. via the [`UV_CONSTRAINT` env var](https://docs.astral.sh/uv/configuration/environment/).

PR #788, opened 2024-08-20
Get rid of lots of sphinx warnings and hide-output

### LINKED ISSUE #784: Unnecessary underscores in the documentation
### Report

It is not uncommon to see elements of code like this in the documentation:
```
_ = ep.tl.causal_inference(
    adata=adata,
    graph=causal_graph,
    treatment="aline_flg",
    outcome="icu_los_day",
    estimation_method="backdoor.linear_regression",
)
```

What is the point of having `_ = `? This is exactly what happens in Python internally, anyway. If this is done to prevent printing an output returned by the function, a much more elegant way would be to put `;` at the end of the line.

I'm pretty sure I asked this question already, but probably not on GitHub. Sorry for not remembering the answer 

### Version information

The current release of ehrapy at the time of writing is `0.8.0`

Zethson:
Fixes #784


eroell:
3.11 pre might get fixed by just restarting it

PR #791, opened 2024-08-24
Added approximate KNN backend

nicolassidoux:
I had to `from scanpy.neighbors._types import KnnTransformerLike, _KnownTransformer` to make transformer argument of `neighbors()` consistent with scanpy. But _types is protected...

### DIFF
--- a/ehrapy/preprocessing/_scanpy_pp_api.py
+++ b/ehrapy/preprocessing/_scanpy_pp_api.py
@@ -1,29 +1,33 @@
 from __future__ import annotations
 
-from collections.abc import Collection, Mapping, Sequence
 from types import MappingProxyType
-from typing import Any, Callable, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Union
 
 import numpy as np
 import scanpy as sc
-from anndata import AnnData
-from scipy.sparse import spmatrix
-from scanpy.neighbors._types import KnnTransformerLike, _KnownTransformer # _types is protected...
+
+if TYPE_CHECKING:
+    from collections.abc import Collection, Mapping, Sequence
+
+    from anndata import AnnData
+    from scanpy.neighbors._types import KnnTransformerLike, _KnownTransformer
+    from scipy.sparse import spmatrix
+
 AnyRandom = Union[int, np.random.RandomState, None]
 
 
 def pca(
-    data: Union[AnnData, np.ndarray, spmatrix],
-    n_comps: Optional[int] = None,
-    zero_center: Optional[bool] = True,
+    data: AnnData | np.ndarray | spmatrix,
+    n_comps: int | None = None,
+    zero_center: bool | None = True,
     svd_solver: str = "arpack",
     random_state: AnyRandom = 0,
     return_info: bool = False,
     dtype: str = "float32",
     copy: bool = False,
     chunked: bool = False,
-    chunk_size: Optional[int] = None,
-) -> Union[AnnData, np.ndarray, spmatrix]:  # pragma: no cover
+    chunk_size: int | None = None,
+) -> AnnData | np.ndarray | spmatrix:  # pragma: no cover
     """Computes a principal component analysis.
 
     Computes PCA coordinates, loadings and variance decomposition. Uses the implementation of *scikit-learn*.
@@ -93,10 +97,10 @@ def pca(
 
 def regress_out(
     adata: AnnData,
-    keys: Union[str, Sequence[str]],
-    n_jobs: Optional[int] = None,
+    keys: str | Sequence[str],
+    n_jobs: int | None = None,
     copy: bool = False,
-) -> Optional[AnnData]:  # pragma: no cover
+) -> AnnData | None:  # pragma: no cover
     """Regress out (mostly) unwanted sources of variation.
 
     Uses simple linear regression. This is inspired by Seurat's `regressOut` function in R [Satija15].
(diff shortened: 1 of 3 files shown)


ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -1,27 +1,33 @@
-from collections.abc import Collection, Mapping, Sequence
+from __future__ import annotations
+
 from types import MappingProxyType
-from typing import Any, Callable, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Union
```
eroell, line 4:
```suggestion
from typing import TYPE_CHECKING, Any, Callable, Literal, Union
```

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -1,27 +1,33 @@
...
     svd_solver: str = "arpack",
     random_state: AnyRandom = 0,
     return_info: bool = False,
     dtype: str = "float32",
     copy: bool = False,
     chunked: bool = False,
-    chunk_size: Optional[int] = None,
-) -> Union[AnnData, np.ndarray, spmatrix]:  # pragma: no cover
+    chunk_size: int | None = None,
+) -> AnnData | np.ndarray | spmatrix:  # pragma: no cover
```
eroell, line 30:
```suggestion
) -> AnnData | np.ndarray | spmatrix | None:  # pragma: no cover
```
just realized, this has been misstyped before already :)

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -225,6 +232,23 @@ def neighbors(
         method: Use 'umap' [McInnes18]_ or 'gauss' (Gauss kernel following [Coifman05]_ with adaptive width [Haghverdi16]_) for computing connectivities.
                 Use 'rapids' for the RAPIDS implementation of UMAP (experimental, GPU only).
         metric: A known metric’s name or a callable that returns a distance.
+        transformer
```
eroell, line 235:
```suggestion
        transformer: Approximate kNN search implementation.
```
Or similar, as in this docstyle there can be a short description after the argument :)

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -225,6 +232,23 @@ def neighbors(
...
+            Approximate kNN search implementation following the API of
```
eroell, line 236:
```suggestion
            Follows the API of
```

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -225,6 +232,23 @@ def neighbors(
...
+            :class:`~sklearn.neighbors.KNeighborsTransformer`.
+            See :doc:`/how-to/knn-transformers` for more details.
```
eroell, line 238: The link breaks in our documentation as this was from the scanpy doc; but we could insert the full scanpy doc link instead https://scanpy.readthedocs.io/en/latest/how-to/knn-transformers.html & make a quick comment that this is scanpy but works the same for ehrapy

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -225,6 +232,23 @@ def neighbors(
...
+            Also accepts the following known options:
+
+            `None` (the default)
+                Behavior depends on data size.
+                For small data, we will calculate exact kNN, otherwise we use
+                :class:`~pynndescent.pynndescent_.PyNNDescentTransformer`
+            `'pynndescent'`
+                :class:`~pynndescent.pynndescent_.PyNNDescentTransformer`
+            `'rapids'`
+                A transformer based on :class:`cuml.neighbors.NearestNeighbors`.
```
eroell, line 248: Here can either add "deprecated since scanpy version 1.10.0" such that its clear where 1.10.0 comes from, or hide this from the beginning as its deprecated already?
Then, would also need to remove from `_KnownTransformer` if we write this ourselves see question on what to do about private variables below

eroell:
@Zethson what is your stance on importing private variables from scanpy, as we are doing [elsewhere](https://github.com/theislab/ehrapy/blob/423a633c35efdd5be9992cade33443432274006f/ehrapy/plot/_scanpy_pl_api.py#L36) too?

I remember in the past you've [added a fix](https://github.com/theislab/ehrapy/commit/f71f08e9b6ed924a600d88898de9abf6fa3ca301) of a previously import variable by defining the required variable ourselves.

I lean towards doing this here too

eroell (changes requested):

Zethson:
> @Zethson what is your stance on importing private variables from scanpy, as we are doing [elsewhere](https://github.com/theislab/ehrapy/blob/423a633c35efdd5be9992cade33443432274006f/ehrapy/plot/_scanpy_pl_api.py#L36) too?
> 
> I remember in the past you've [added a fix](https://github.com/theislab/ehrapy/commit/f71f08e9b6ed924a600d88898de9abf6fa3ca301) of a previously imported, external private variable by defining the required variable ourselves.
> 
> I lean towards doing this here too

Yeah, we should try to avoid importing internals so I think that's a good option.

eroell:
Thank you so much @nicolassidoux !
One thing I realized here was that we should add tests for the pass through function calls of scanpy...

But this is from our side so far a lacking suite of tests (as we as of now have been closely eyeing scanpy), and hence no test suite at this point to add to here :)

eroell:
ready for me to take a quick look again or are you still doing stuff? :)

nicolassidoux:
Hi @eroell, ready for re-review!

eroell (approved):

eroell:
Added the intersphinx mappings for pynndescent, and linked the reference to the scanpy knn tutorial with a readable text; Also added a more wordy description on the arguments, more wordy than scanpy...
So direct commit of suggestions instead of PR to save your time, what is your take on these additions @nicolassidoux? Would you as user think this is more clear now, or worse?

nicolassidoux:
@eroell 
Yes, direct commit is a good idea indeed! 
In regard with documentation stuff, I really let you decide whatever suits, I definitely don't have the skills to judge from a user perspective :smile: But AFAIK the description looks good!
I let you close the PR if you think all is good!

eroell:
great - this is a nice addition, thank you @nicolassidoux !

Issue #824, opened 2024-11-13
Make all imputation methods consistent in regard to encoding requirements


nicolassidoux:
### Description of feature

Imputation requires proper encoding of the data, but at the moment, there is no consistent strategy about what to do if a passed AnnData is not encoded. Some methods throw an exception, while some others silently encode the data with an arbitrary method.

It would be probably better to just throw an exception and let the caller decide what to do instead of altering the data.

| **Function**         | **Encoding required** | **Action if not encoded** |
|----------------------|-----------------------|---------------------------|
| `explicit_impute`    | No                    | None                      |
| `simple_impute`      | Strategy-dependant    | Raise ValueError          |
| `knn_impute`         | Yes                   | Perform ordinal encoding  |
| `miss_forest_impute` | Yes                   | Perform ordinal encoding  |
| `mice_forest_impute` | Yes                   | Perform ordinal encoding  |

So I suggest to modify `knn_impute`, `miss_forest_impute` and `mice_forest_impute`.

nicolassidoux:
### Comments so far (note: I haven’t tested anything yet)  

1. **Imports from protected**  
   - I noticed that `prepocessing/_imputation.py` contains imports from protected modules or functions:  
     ```python  
     from ehrapy.anndata.anndata_ext import _get_column_indices  
     ```  
     - `_get_column_indices` is primarily used outside its module and could be valuable to the public API. Since the documentation is already in place, I suggest making it public.  

     ```python  
     from ehrapy.core._tool_available import _check_module_importable  
     ```  
     - `_check_module_importable` is only used (outside its module) in `prepocessing/_imputation.py` to check for the presence of sklearnex. If conditional execution based on this is truly necessary, I propose making `_check_module_importable` public across the entire ehrapy module.  

     - For consistency, I also recommend making `_shell_command_accessible` module-public. To align naming conventions, I renamed `_tool_available` to `_utils_available` to match `_utils_doc` (previously `_doc_util`).  

2. **Return Value Behavior**  
   - Some functions return an `AnnData` object only if the `copy` argument is set. I believe this is a mistake. The hints specifices the functions should consistently return the `AnnData` object passed to them, regardless of whether it’s copied at the start.  

3. **Use of `Progress`**  
   - I noticed some `Progress` are being used, but as far as I know, we don’t have any way to track the progress of imputations. To simplify the code, I recommend removing these.  

4. **Improvements to `_get_non_numerical_column_indices`** (`prepocessing/_imputation.py`)
[...]

PR #827, opened 2024-11-24
Make all imputation methods consistent in regard to encoding requirements

nicolassidoux:
See #824 for the details of the change.


ehrapy/_utils_available.py
```diff
@@ -4,7 +4,7 @@
 from subprocess import PIPE, Popen
 
 
-def _check_module_importable(package: str) -> bool:  # pragma: no cover
+def check_module_importable(package: str) -> bool:  # pragma: no cover
```
Zethson, line 7: I read your justification of why you think that these methods should be public across ehrapy. The reason why I thought that they should be underscore methods is because we don't want users to access these. They are not public API. Now currently they are still in a private module so there'd be some safe guard but I'm not sure whether we need to make them public across ehrapy.

Happy to be convinced otherwise though!

ehrapy/_utils_data.py
```diff
@@ -0,0 +1,53 @@
...
+
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from anndata import AnnData
+
+
+def are_dataset_equal(arr1: np.ndarray, arr2: np.ndarray) -> np.bool_:
```
Zethson, line 11:
```suggestion
def are_ndarrays_equal(arr1: np.ndarray, arr2: np.ndarray) -> np.bool_:
```

This would much better reflect the current behavior. I wouldn't mind having this completely generalized though with all matrix types.

ehrapy/_utils_data.py
```diff
@@ -0,0 +1,53 @@
...
+        arr1: First array to compare
+        arr2: Second array to compare
+
+    Returns:
+        True if the two arrays are equal member-wise
+    """
+    return np.all(np.equal(arr1, arr2, dtype=object) | ((arr1 != arr1) & (arr2 != arr2)))
+
+
+def to_missing_data_matrix(data: np.ndarray) -> np.ndarray[Any, np.dtype[np.bool_]]:  # pragma: no cover
```
Zethson, line 26:
```suggestion
def is_val_missing(data: np.ndarray) -> np.ndarray[Any, np.dtype[np.bool_]]:
```

Zethson: I don't understand the name of the function too well. It should be more precise.

 Why the `pragma: no cover`? This function can be quickly tested.

ehrapy/_utils_data.py
```diff
@@ -0,0 +1,53 @@
...
+    """Check if a AnnData value is missing.
+
+    Args:
+        data: The AnnData data to check
```
Zethson, line 30: Let's not call this data but rather `matrix` or so maybe? In the AnnData jargon we are always talking about matrices that are in X or layers.

ehrapy/_utils_data.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Returns:
+        The layer as a dense matrix.
+    """
+    from scipy.sparse import issparse
+
+    if layer is None:
+        return adata.X.toarray() if issparse(adata.X) else adata.X
+    else:
+        return adata.layers[layer] if issparse(adata.layers[layer]) else adata.layers[layer]
```
Zethson, line 53: Is there a `toarray()` missing here?

ehrapy/_utils_data.py
```diff
@@ -0,0 +1,53 @@
...
+    Args:
+        data: The AnnData data to check
+
+    Returns:
+        An array of bool representing the missingness of the original data, with the same shape
+    """
+    return np.isin(data, [None, "", np.nan]) | (data != data)
+
+
+def to_dense_matrix(adata: AnnData, layer: str | None = None) -> np.ndarray:  # pragma: no cover
```
Zethson, line 38: The folks at scanpy have developed a very fast numba kernel for this which is much faster than scipy. See https://github.com/scverse/scanpy/blob/7131500627a3037cedd11d380ae772400c744f9b/src/scanpy/preprocessing/_utils.py#L165

I have already advocated for this to be public API in anndata instead so that all packages can make use of it. @ilan-gold @Intron7 .

ehrapy/preprocessing/_imputation.py
```diff
@@ -2,22 +2,18 @@
 
 import warnings
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, List, Literal
```
Zethson, line 5:
```suggestion
from typing import TYPE_CHECKING, Literal
```

We are using 3.10+ type hints. We don't import primitive types anymore.

`from __future__ import annotations` allow us to use that feature in a forward compatible way.

ehrapy/preprocessing/_imputation.py
```diff
@@ -155,38 +141,30 @@ def simple_impute(
...
-                f"Unknown impute strategy {strategy} for simple Imputation. Choose any of mean, median or most_frequent."
+        except ValueError:
+            raise ValueError(
+                f"Can only impute numerical data using {strategy} strategy. Try to restrict imputation"
+                "to certain columns using var_names parameter or use a different mode."
             ) from None
+    # most_frequent imputation works with non-numerical data as well
+    elif strategy == "most_frequent":
+        _simple_impute(adata, var_names, strategy)
+    # unknown simple imputation strategy
```
Zethson, line 155:
```suggestion
```

This is self documenting code

ehrapy/preprocessing/_imputation.py
```diff
@@ -155,38 +141,30 @@ def simple_impute(
...
+    else:
+        raise ValueError(  # pragma: no cover
```
Zethson, line 157: This could also be tested instead of a `pragma: no cover` but it's a trivial test.

ehrapy/preprocessing/_imputation.py
```diff
@@ -353,12 +312,12 @@ def miss_forest_impute(
...
     See https://academic.oup.com/bioinformatics/article/28/1/112/219101.
-    This requires the computation of which columns in X contain numerical only (including NaNs) and which contain non-numerical data.
+
+    If required, the data needs to be properly encoded as this imputation requires numerical data only.
 
     Args:
         adata: The AnnData object to use MissForest Imputation on.
-        var_names: List of columns to impute or a dict with two keys ('numerical' and 'non_numerical') indicating which var
-                   contain mixed data and which numerical data only.
+        var_names: List of columns to impute
```
Zethson, line 320:
```suggestion
        var_names: Iterable of columns to impute
```

You changed the type to `Iterable` so we can also note this here.

ehrapy/preprocessing/_imputation.py
```diff
@@ -382,10 +341,10 @@ def miss_forest_impute(
         _warn_imputation_threshold(adata, list(adata.var_names), threshold=warning_threshold)
     elif isinstance(var_names, dict):
         _warn_imputation_threshold(adata, var_names.keys(), threshold=warning_threshold)  # type: ignore
-    elif isinstance(var_names, list):
+    elif isinstance(var_names, Iterable) and all(isinstance(item, str) for item in var_names):
```
Zethson, line 344: Why do we need the second check? You surely have a good reason but I'd just like to understand. It's another operation so I wonder whether we need it.

ehrapy/preprocessing/_imputation.py
```diff
@@ -394,72 +353,65 @@ def miss_forest_impute(
...
-                random_state=random_state,
-            )
-            # initial strategy here will not be parametrized since only most_frequent will be applied to non numerical data
-            imp_cat = IterativeImputer(
-                estimator=RandomForestClassifier(n_estimators=n_estimators, n_jobs=settings.n_jobs),
-                initial_strategy="most_frequent",
-                max_iter=max_iter,
-                random_state=random_state,
-            )
+        if settings.n_jobs == 1:  # pragma: no cover
```
Zethson, line 356: I wonder whether we should remove this warning and just default to -1. There's plenty of algorithms that always default to all available cores.

ehrapy/preprocessing/_imputation.py
```diff
@@ -509,47 +463,30 @@ def mice_forest_impute(
...
+                var_names,
+                save_all_iterations_data,
+                random_state,
+                inplace,
+                iterations,
+                variable_parameters,
+                verbose,
+            )
+        else:
+            # Raise exception if we have some non-numerical data
```
Zethson, line 480:
```suggestion
```

It's self documenting here again I think.

ehrapy/preprocessing/_imputation.py
```diff
@@ -616,27 +553,21 @@ def _warn_imputation_threshold(adata: AnnData, var_names: Iterable[str] | None,
     return var_name_to_pct
 
 
-def _get_non_numerical_column_indices(X: np.ndarray) -> set:
+def _get_non_numerical_column_indices(x: np.ndarray) -> set:
```
Zethson, line 556: It used to be called `X` because it was designed to be used for the `X` matrix of AnnData. For such a general function though, it's a terrible name.

```suggestion
def _get_non_numerical_column_indices(arr: np.ndarray) -> set:
```

tests/preprocessing/test_imputation.py
```diff
@@ -64,158 +71,237 @@ def test_median_impute_throws_error_non_numerical(impute_adata):
...
+    adata_imputed = knn_impute(impute_num_adata, copy=True)
+    # col3 didn't have a NaN at row 1, let's simulate it has been modified by mistake
+    adata_imputed.X[1, 2] = 42.0
+    assert not _base_check_imputation(impute_num_adata, adata_imputed)
+
+
+# ----------------------------------------------------------------------------------------------------------------------
+
+
+# Base check for imputation (imputation doesn't leave NaN behind, it doesn't affect non-imputated columns, it doesn't
```
Zethson, line 257: Let's just add this as a proper docstring to this function.

tests/preprocessing/test_imputation.py
```diff
@@ -64,158 +71,237 @@ def test_median_impute_throws_error_non_numerical(impute_adata):
...
+# affect data in imputated columns that wasn't NaN)
+def _base_check_imputation(
```
Zethson, line 259: I'd consider moving this function to the very top. I always try to have fixtures and other test util functions etc at the top so that people see early how the test is set up.

tests/preprocessing/test_imputation.py
```diff
@@ -64,158 +71,237 @@ def test_median_impute_throws_error_non_numerical(impute_adata):
...
+    adata_before_imputation: AnnData,
+    adata_after_imputation: AnnData,
+    before_imputation_layer: str | None = None,
+    after_imputation_layer: str | None = None,
+    imputed_var_names: Iterable[str] | None = None,
+) -> bool:
+    from ehrapy._utils_data import are_dataset_equal, to_dense_matrix, to_missing_data_matrix
```
Zethson, line 266: If these functions are only used here, I really wonder whether we need them in ehrapy itself. Maybe scoping them in the tests module is better? We can also populate the conftest.py a bit with such functions.

Zethson (changes requested):
Thank you very very much! This is great. I am a fan of simplifying the code (by removing semi important features such as the progress bars in general) and really think that generalizing the imputation tests is an awesome idea.

 I have a few questions and concerns though:

1. I am not a fan of adding more `_utils_*` modules. I argue that all and any utils modules are a code smell. It is never clear to developers what is actually in there and it should probably be somewhere else. Let's just have descriptive module names (or scripts) that better describe what can be found in there.
2. I don't know how I feel about the removal of the imputation spinners. We had added them because some imputation functions can run for a looooong time and it just gives feedback that the code is still running. I know - it's not a progress bar but it's at least something? I don't have the strongest opinion on this but wonder what other people think @Imipenem @eroell 
3. Removing some docstrings about raised Errors is okay with me because the errors themselves can document the behavior when raised. The docstrings are more useful for other developers if they want to build on top because they don't have to check the whole code for any potential raised Errors.
4. Instead of printing to stdout and returning a bool in `_base_check_imputation` I wonder whether this function should just raise errors? prints are not for errors - they are for informative messages. Pytest will resolve the stack and point to the part that raised the error.

ehrapy/_utils_data.py
nicolassidoux: Just a precision here: this function returns an array, not a simple bool as your suggestion may let think.

ehrapy/preprocessing/_imputation.py
nicolassidoux: Because while we can check that `var_names` is an `Iterable`, we cannot specifically check if it is a `Iterable[str]` (AFAIK)

nicolassidoux:
> 1. I am not a fan of adding more `_utils_*` modules. I argue that all and any utils modules are a code smell. It is never clear to developers what is actually in there and it should probably be somewhere else. Let's just have descriptive module names (or scripts) that better describe what can be found in there.
Well, this is always a big topic. In a way you're right, having these utils is a hint something could be better organized. But it doesn't mean it is in our control. For example: `_is_val_missing` is a function that would better fit in `AnnData` itself. So I would consider those as "extensions" more than "utils".

> 2. I don't know how I feel about the removal of the imputation spinners. We had added them because some imputation functions can run for a looooong time and it just gives feedback that the code is still running. I know - it's not a progress bar but it's at least something? I don't have the strongest opinion on this but wonder what other people think @Imipenem @eroell
I never saw it spinning 😆 But maybe it's an issue with my stdout.

> 3. Removing some docstrings about raised Errors is okay with me because the errors themselves can document the behavior when raised. The docstrings are more useful for other developers if they want to build on top because they don't have to check the whole code for any potential raised Errors.
Again a bit topic 😄 As I said, it sounds like a good idea, but almost not maintenable in practice. From my experience.

> 4. Instead of printing to stdout and returning a bool in `_base_check_imputation` I wonder whether this function should just raise errors? prints are not for errors - they are for informative messages. Pytest will resolve the stack and point to the part that raised the error.
This is what I would have done for production code, but usually for tests, I like that kind of output. But alright, I'll modify that.

nicolassidoux:
Looks the `Test / test (ubuntu-latest, 3.11, --pre) (pull_request) ` fails...

eroell:
> I don't know how I feel about the removal of the imputation spinners. We had added them because some imputation functions can run for a looooong time and it just gives feedback that the code is still running. I know - it's not a progress bar but it's at least something? I don't have the strongest opinion on this but wonder what other people think @Imipenem @eroell

I'm split... if we were to keep it, I'd move it to a decorator? That would make the functions way more readable, and reduce redundancy

> I never saw it spinning 😆 But maybe it's an issue with my stdout.

You run usually scripts if I recall correctly? If you run it from a jupyter notebook, on this 6 dots on the image, you can see something like a snake walking around to show the spinning
image

ehrapy/_settings.py
```diff
@@ -53,7 +53,7 @@ def __init__(
         figdir: str | Path = "./figures/",
         cache_compression: str | None = "lzf",
         max_memory=15,
-        n_jobs: int = 1,
```
eroell, line 56: Can this have more effects than we hope for..? I vaguely recall a method in a tutorial once having failed with n_jobs=-1, but not sure if really was this or something joblib related (it was a bug in a package though).

Not sure if it would be safer to keep n_jobs as 1, users can still set this to -1 at their wish

ehrapy/_utils_data.py
```diff
@@ -0,0 +1,54 @@
...
+def _is_val_missing(data: np.ndarray) -> np.ndarray[Any, np.dtype[np.bool_]]:
+    """Check if values in a AnnData matrix are missing.
+
+    Args:
+        data: The AnnData matrix to check
+
+    Returns:
+        An array of bool representing the missingness of the original data, with the same shape
+    """
+    return np.isin(data, [None, "", np.nan]) | (data != data)
```
eroell, line 35:
```suggestion
    return np.isin(data, [None, ""]) | (data != data)
```

ehrapy/_utils_data.py
Zethson: I don't think that we can rely on it being where it currently is. I'll lobby for it to be public and then we can start using it. For now, let's just use what we have.

Zethson (changes requested):
> For example: _is_val_missing is a function that would better fit in AnnData itself. So I would consider those as "extensions" more than "utils".

Then @nicolassidoux they should be in https://github.com/theislab/ehrapy/blob/main/ehrapy/anndata/anndata_ext.py right? I still expect a lot of that to eventually make it into @eroell ehrdata but for now it should probably be there.

Could you please move the new functions?

I'm okay with the spinners being removed unless you deem them worthwhile now @nicolassidoux and want to write a decorator.

After the functions have been moved, I'd love to merge this.

Thank you very much!

nicolassidoux:
@Zethson
>Then @nicolassidoux they should be in https://github.com/theislab/ehrapy/blob/main/ehrapy/anndata/anndata_ext.py right? I still expect a lot of that to eventually make it into @eroell ehrdata but for now it should probably be there.

Oh well I should have noticed this file indeed 😄 

>I'm okay with the spinners being removed unless you deem them worthwhile now @nicolassidoux and want to write a decorator.

Let me assess how much work it is to make that done.

ehrapy/preprocessing/_imputation.py
```diff
@@ -155,38 +141,29 @@ def simple_impute(
...
+    if strategy in {"median", "mean"}:
+        try:
             _simple_impute(adata, var_names, strategy)
-        # unknown simple imputation strategy
-        else:
-            raise ValueError(  # pragma: no cover
-                f"Unknown impute strategy {strategy} for simple Imputation. Choose any of mean, median or most_frequent."
+        except ValueError:
+            raise ValueError(
+                f"Can only impute numerical data using {strategy} strategy. Try to restrict imputation"
```
eroell, line 149:
```suggestion
                f"Can only impute numerical data using {strategy} strategy. Try to restrict imputation "
```

ehrapy/preprocessing/_imputation.py
```diff
@@ -274,40 +246,26 @@ def knn_impute(
...
-                adata.X = adata.X.astype("object")
-                # decode ordinal encoding to obtain imputed original data
-                adata.X[::, column_indices] = enc.inverse_transform(adata.X[::, column_indices])
+        if np.issubdtype(adata.X.dtype, np.number):
+            _knn_impute(adata, var_names, n_neighbors, backend=backend, **backend_kwargs)
+        else:
+            # Raise exception since non-numerical data can not be imputed using KNN Imputation
+            raise ValueError(
+                "Can only impute numerical data. Try to restrict imputation to certain columns using "
+                "var_names parameter or."
```
eroell, line 257: alternatively to be in sync with miss_forest and mice_forest:
```suggestion
                "var_names parameter."
```

tests/preprocessing/test_imputation.py
```diff
@@ -20,17 +23,127 @@
...
+    unchanged_mask = ~imputed_mask
+    if not _are_ndarrays_equal(layer_before[:, unchanged_mask], layer_after[:, unchanged_mask]):
+        raise AssertionError("Values outside imputed columns were modified.")
+
+    # Ensure imputation does not alter non-NaN values in the imputed columns
+    imputed_non_nan_mask = (~before_nan_mask) & imputed_mask
+    if not _are_ndarrays_equal(layer_before[imputed_non_nan_mask], layer_after[imputed_non_nan_mask]):
+        raise AssertionError("Non-NaN values in imputed columns were modified.")
+
+    # All checks passed
```
eroell, line 84:
```suggestion
```

nicolassidoux:
Hm of course all notebooks tests or almost fail because of `yaspin` missing. This spinner is nice because it works for both scripts and notebook. What do you guys think?

tests/preprocessing/test_imputation.py
Zethson: I think that's self evident enough but this is a nitpick! You don't have to remove it.

Zethson:
Rich is much much more popular and also used in packages such as `pip` itself. Therefore, users will have it anyways. That's why I think it'd be better if we stuck to Rich?

nicolassidoux:
Ok, it doesn't work well on my stdout, but ok in notebooks.

ehrapy/_settings.py
Zethson: It should be good I think.

nicolassidoux:
@eroell @Zethson Ready for final review, I think

eroell:
Spinners work, cool. Thanks for making the decorator!

Nice to have these additional tests, aim of the PR is fulfilled.

I have one final point on the return type. Before, all imputation methods returned a new `adata` if `copy=True`, else `None`. This is the "scverse"/"scanpy" style behavior. miceforest was an (I suppose accidental) exception, which always returned the adata.

Right now, it has been changed to all imputation methods returning `adata`. I suggest to rather have all functions following the "scverse"/"scanpy" style instead.

Even if this could mean a small breaking change for users doing `adata = ep.pp.mice_forest_impute(adata, copy=False)`, I think going consistent here now helps sharpening our API for the future.

nicolassidoux:
@eroell 
>I have one final point on the return type.

Fine for me, anyway I advocated for removing this copy thing, but that's a separate debate.
I'll modify that right now so the issue can be closed ASAP.

Zethson (approved):

nicolassidoux:
Hmm pre-commit doesn't like there is no return statement if copy=False. Should i modify like:
```python
    ...
    if copy:
        return adata
    else:
        return
```

Zethson:
> Hmm pre-commit doesn't like there is no return statement if copy=False. Should i modify like:
> 
> ```python
>     ...
>     if copy:
>         return adata
>     else:
>         return
> ```

Also how scanpy does it: https://github.com/scverse/scanpy/blob/main/src/scanpy/tools/_dpt.py#L199C5-L199C35

eroell:
Thanks for the final polish here - Like your suggestion, or like scanpy does it, both fine!
Then this looks good to merge for me :)

eroell (approved):

nicolassidoux:
Done!

Zethson:
The RTD build seems to time out because of dependencies somehow urgh

Zethson:
I'll merge this now and I'll take a look myself soon.

PR #835, opened 2024-12-04
Add array type handling for normalization

### LINKED ISSUE #834: Array type handling: Normalization
### Description of feature

Ensure the normalization methods handle the array types (numpy ndarray, dask array, check if we want more?) consistently, and properly fail if an unsupported array is encountered.

`log_norm`
- [ ] Consistent handling
- [ ] Test

`max_abs_norm`
- [ ] Consistent handling
- [ ] Test

`minmax_norm`
- [ ] Consistent handling
- [ ] Test

`power_norm`
- [ ] Consistent handling
- [ ] Test

`quantile_norm`
- [ ] Consistent handling
- [ ] Test

`robust_scale_norm`
- [ ] Consistent handling
- [ ] Test

`scale_norm`
- [ ] Consistent handling
- [ ] Test

`offset_negative_values`
- [ ] Consistent handling
- [ ] Test

eroell:
**Description of changes**

Implements different array types for normalization functions which serve as example and guidance on how we implement this for all other functions.

**Technical details**

Uses singledispatch. Inspired from e.g. scanpy's [scale](https://github.com/scverse/scanpy/blob/main/src/scanpy/preprocessing/_scale.py); but less performance engineering, rather focusing on leveraging frameworks such as dask-ml.


ehrapy/preprocessing/_normalization.py
```diff
@@ -8,9 +9,14 @@
 from ehrapy._compat import is_dask_array
 
 try:
+    import dask.array as da
     import dask_ml.preprocessing as daskml_pp
+
+    DASK_AVAILABLE = True
```
eroell, line 15: What is your take on such a flag?
Could it be worth to introduce dask as a dependency now that we try to introduce it systematically?

ehrapy/preprocessing/_normalization.py
```diff
@@ -69,6 +75,23 @@ def _scale_func_group(
         return None
 
 
+@singledispatch
```
eroell, line 78: What do you think about systematically introducing singledispatch in this manner? It might be a bit of an overkill here; but I think in the long run, using this structure reduces code complexity by introducing a regularly appearing pattern.

eroell: The StandardScaler I think works also with sparse matrices. E.g. the MaxAbsScaler does not I think. For focusing on the repeating patterns in a first step, have been restrictive here in disallowing everything but `np.array` and `dask.array.Array`

tests/preprocessing/test_normalization.py
```diff
@@ -87,12 +89,35 @@ def test_vars_checks(adata_to_norm):
         ep.pp.scale_norm(adata_to_norm, vars=["String1"])
 
 
-@pytest.mark.parametrize("array_type", ARRAY_TYPES)
-def test_norm_scale(array_type, adata_to_norm):
+# TODO: list the supported array types centrally?
```
eroell, line 92: These two lists are not used right now. Should we have a central lookup for which function works with which array type?

Right now, the "ground truth" is tested in the `test_<normalization-method>_array_types` test.

ehrapy/preprocessing/_normalization.py
Zethson: anndata[dask] I guess to be compatible but fine with me

ehrapy/preprocessing/_normalization.py
Zethson: Yes, I agree with your argument

tests/preprocessing/test_normalization.py
Zethson: > Should we have a central lookup for which function works with which array type?

wdym with lookup here?

tests/preprocessing/test_normalization.py
```diff
@@ -87,12 +89,35 @@ def test_vars_checks(adata_to_norm):
...
+norm_scale_supported_types = [np.asarray, da.asarray]
+norm_scale_unsupported_types = [sparse.csc_matrix]
+
+
+# TODO: check this for each function, with just default settings?
+@pytest.mark.parametrize(
```
eroell, line 98: I picked strict three array types to be tested rigidly: `np.ndarray`, `dask.array.Array`, `scipy.sparse.spmatrix`. Not sure if we should test for more. I personally favor being strict about these three things, and not guaranteeing anything else.

Zethson: Yeah totally fine. cupy arrays are for the future

tests/preprocessing/test_normalization.py
eroell: For example a `supported_array_types` dict or dataframe in e.g. `ehrapy.core._supported_array_types.py`
```
supported_array_types = {
    `ep.pp.scale_norm`: {"np.array": True, "dask.array.Array": True, "sp.sparse.spmatrix: False},
    ....
}
```

Could be used in the test parametrization, instead of writing them up individually all the time.

But I have not seen something like that anywhere before, and there might be good reasons to not do this

eroell:
Very happy about your input on any of the questions I raised in the comments @flying-sheep :)

ehrapy/preprocessing/_normalization.py
eroell: Yes, good point. If dask as dependency, then as `anndata[dask]`

eroell:
Also interested what @nicolassidoux thinks here! :)

Zethson (commented):
1. You added lots of `NotImplementedError` messages. I wonder whether we can DRY this more easily?
2. Those error messages should also always suggest which types are support for that function as a sort of solution.
3.    `adata_to_norm_casted = adata_to_norm.copy()` why do we always need those copies? Doesn't the fixture already create new copies?

tests/preprocessing/test_normalization.py
eroell: Very good point. In favor of using dynamically constructed registries indeed.

eroell:
> You added lots of NotImplementedError messages. I wonder whether we can DRY this more easily?

Yess
> Those error messages should also always suggest which types are support for that function as a sort of solution.

Good point, the dynamic registries help here again
> adata_to_norm_casted = adata_to_norm.copy() why do we always need those copies? Doesn't the fixture already create new copies?

because I just not paying attention to this yet, I'll remove :)

ehrapy/preprocessing/_normalization.py
flying-sheep: Depends on how slow dask is to import. This varies wildly depending on the project.

E.g. AnnData is a little slow because of pandas. In scanpy we lazy-import some sklearn subpackage because that would have a huge impact, and without it we’re only slower than anndata because of `sklearn.utils` and `numba`.

ehrapy/preprocessing/_normalization.py
```diff
@@ -113,6 +133,23 @@ def scale_norm(
     )
 
 
+@singledispatch
+def _minmax_norm_function(arr):
+    raise NotImplementedError(f"minmax_norm does not support data to be of type {type(arr)}")
```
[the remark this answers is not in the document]
flying-sheep: you can automate this by accessing `_minmax_norm_function.registry.keys()`

Zethson:
I addressed the feedback raised in the discussions. Hope I didn't miss any. I'll merge this now but if I forgot something @eroell please make me aware. Thank you!

PR #854, opened 2025-01-15
Update power norm test reference values

eroell:
**PR Checklist**

**Description of changes**
Update the expected results of the Power Norm function in the test.
I dont know why the outcome changed slightly, and suspect the small test set causing numeric unstability in the underlying algorithm. The changes happen on the 2nd digit after the comma.

**Technical details**

**Additional context**

### DIFF
--- a/tests/preprocessing/test_normalization.py
+++ b/tests/preprocessing/test_normalization.py
@@ -83,51 +83,6 @@ def adata_to_norm():
     return adata
 
 
-@pytest.fixture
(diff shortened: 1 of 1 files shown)


eroell:
Think this should do it, keeps the fairly strict tests, and the testing dataset minimally small

Zethson:
I think something is either flaky or heavily dependent on some dependency version because I had already changed it once as well. Let's try this.

Issue #861, opened 2025-01-29
Quality Control Rewrite


aGuyLearning:
### Description of feature

I propose a rewrite of the quality control functions to work with single dispatch method rewrites described in #851  and #848. 

To this end the qc_metrics function has to be looked at _obs_qc_metrics and _var_qc_metrics have to be rewritten using single dispatch to handle the different datatypes. 

For now those will be dask.array and numpy.array

- [ ] single dispatch functions for _obs_qc_metrics
- [ ] single dispatch functions for _var_qc_metrics

This will most likely be handled in a big PR with the two issues mentioned.

eroell:
Closed with #865

PR #865, opened 2025-02-05
Datatype Support in Quality Control and Impute

### LINKED ISSUE #848: Compatibility of ep.pp.explicit_impute with different datatypes
### Description of feature

Follow the template workflow of #835:
- [ ] introduce singledispatch to check for allowed data types, or raise an error if others are encountered
- [ ] adapt the tests to check these data types

Create underscore private function for each datatype, and keep the adata object as primary

### LINKED ISSUE #851: Compatibility of ep.pp.mice_forest_impute with different datatypes
### Description of feature

Follow the template workflow of #835:
- [ ] introduce singledispatch to check for allowed data types, or raise an error if others are encountered
- [ ] adapt the tests to check these data types

aGuyLearning:
**PR Checklist**

- [x] Add Tests for the encoding mode 
- [x] add Tests for the deepcopy removal

**Description of changes**
As discussed in #861, this refactors the quality control functions to use single dispatch for future datatype support. 

To this end, the explicit_impute function has been reworked already. closes #848 

mice_forest_impute works on dataframes, so tests have been added to check for dask and np arrays. closes #851 

**Technical details**
This pull request introduces significant changes to the `ehrapy` preprocessing module, focusing on improving imputation and quality control functionalities. The main changes involve the addition of support for `dask` arrays, the use of `singledispatch` for function overloading, and the enhancement of test coverage for different array types.

**Possible enhancements**
Please let's discuss the changes, as I am not yet fully happy with them.


ehrapy/preprocessing/_quality_control.py
```diff
@@ -55,59 +57,52 @@ def qc_metrics(
...
+    var_metrics = _compute_var_metrics(mtx, adata)
+    obs_metrics = _compute_obs_metrics(mtx, adata, qc_vars=qc_vars, log1p=True)
+
     adata.var[var_metrics.columns] = var_metrics
+    adata.obs[obs_metrics.columns] = obs_metrics
 
     return obs_metrics, var_metrics
 
 
-def _missing_values(
```
eroell, line 67: Removed this (from before), it seemed unnecessarily complicated

ehrapy/preprocessing/_quality_control.py
```diff
@@ -55,59 +57,52 @@ def qc_metrics(
...
+    _raise_array_type_not_implemented(_compute_obs_metrics, type(arr))
+    # TODO: add tests for this function
+
+
+@_compute_obs_metrics.register(np.ndarray)
+def _(arr: np.array, adata: AnnData, *, qc_vars: Collection[str] = (), log1p: bool = True):
     obs_metrics = pd.DataFrame(index=adata.obs_names)
     var_metrics = pd.DataFrame(index=adata.var_names)
-    mtx = adata.X if layer is None else adata.layers[layer]
+    mtx = copy.deepcopy(arr.astype(object))
```
eroell, line 102: Is this deepcopy because mtx gets written onto if `"encoding_mode" in adata.var`? Or is there something else you spotted here requiring this?

eroell: From a first glance I think mtx was written to if `"encoding_mode" in adata.var`, which would be a bug. If this is also your reason here @aGuyLearning, we might look for something cheaper than copying the entire mtx

eroell:
As of now, `explicit_impute` raises a NotImplementedError for dask which is solely due to the required function from quality control is not available, right?
I see that you improved `explicit_impute` to support dask, once the quality control block is out of the way 👍 
I'm looking into the quality control right now, there's some potential to daskify further

ehrapy/preprocessing/_quality_control.py
aGuyLearning: i removed the copy part and renamed the method argument to mtx, as it will just directly use the array. In addition, the `mtx.astype(object)` has been moved back into the "encoding_mode" block.

Should i rename the _compute_var_metrics argument as well?

eroell:
Daskification of `qc_metrics`.
Profiling for AnnData of shape 30'000 x 10'000, reduces the peak memory consumption from 2G to 350MB.

**numpy**
```py
import anndata as ad
import numpy as np
import ehrapy as ep

X_np = rng.random((30_000, 1_000))
X_np[X_np <= 0.1] = np.nan

adata_np = ad.AnnData(X_np)
df = ep.pp.qc_metrics(adata_np)[0]
```
image


**dask**
```py
import anndata as ad
import dask.array as da
import numpy as np

import ehrapy as ep

X_dask = da.random.random(size=(30_000, 1_000), chunks=(1000, 1000))
X_dask[X_dask <= 0.1] = np.nan

adata_da = ad.AnnData(X_dask)
df = ep.pp.qc_metrics(adata_da)[0]
```
image

tests/preprocessing/test_quality_control.py
```diff
@@ -82,19 +86,57 @@ def test_var_qc_metrics(missing_values_adata):
     assert (~var_metrics["iqr_outliers"]).all()
 
 
+@pytest.mark.parametrize(
+    "array_type, expected_error",
+    [
+        (np.array, None),
+        # (da.array, NotImplementedError),
```
eroell, line 93: Could you check here @aGuyLearning whether (`da.array`, None) runs smoothly now?

tests/preprocessing/test_imputation.py
```diff
@@ -296,7 +303,26 @@ def test_miceforest_impute_numerical_data(impute_iris_adata):
...
+    ],
+)
+def test_explicit_impute_array_types(impute_num_adata, array_type, expected_error):
+    impute_num_adata.X = array_type(impute_num_adata.X)
+    if expected_error:
+        with pytest.raises(expected_error):
+            explicit_impute(impute_num_adata, replacement=1011, copy=True)
+
+
+@pytest.mark.parametrize(
```
eroell, line 321: Yes, that would be great! If we want to check `da.from_array`, by default passing the argument `chunks=1000` helps, as dask complains when calling `da.from_array` when dtype is simply object in some cases. Would you be interested in adding such a fixture @aGuyLearning?

tests/preprocessing/test_quality_control.py
eroell, line 93: With the new fixture mentioned above, the dask test could be activated! It most likely will fail if just calling `da.from_array`, with no `chunks` argument specified

eroell (changes requested):
See comments - this comes along very well!

ehrapy/preprocessing/_imputation.py
```diff
@@ -19,6 +21,13 @@
 if TYPE_CHECKING:
     from anndata import AnnData
 
+try:
```
Zethson, line 24: We have this check in every file now. Should this be done once somewhere else? Probably just do what scanpy does. If they also check in every file - fine.

tests/preprocessing/test_imputation.py
```diff
@@ -46,6 +48,11 @@ def _base_check_imputation(
     Raises:
         AssertionError: If any of the checks fail.
     """
+    # if .x of the AnnData is a dask array, convert it to a numpy array
```
Zethson, line 51:
```suggestion
    # if .X of the AnnData is a dask array, convert it to a numpy array
```

Zethson: But I can see this in the code. Can you write WHY you're doing this, please? Much more useful.

Zethson (commented):
Cool work !

ehrapy/preprocessing/_imputation.py
aGuyLearning: I just checked what scanpy was doing and i believe we could do the same in ehrapy. 
There will be a separate issue for that, as i might have to touch a few things.

Zethson: That would be great, thank you!

ehrapy/preprocessing/_quality_control.py
eroell: Lets discard the deepcopy here

1. keep the original data that is not affected, instead of copying use sliceing for assembling the "mtx" which is checked for missing values
2. the part which needs to be built from the obs labels again should then be concatenated with the unaltered part above

ehrapy/preprocessing/_imputation.py
aGuyLearning: Done. 
https://github.com/theislab/ehrapy/issues/871

ehrapy/preprocessing/_quality_control.py
```diff
@@ -55,55 +63,59 @@ def qc_metrics(
...
+    qc_vars: Collection[str] = (),
+    log1p: bool = True,
+):
     """Calculates quality control metrics for observations.
 
     See :func:`~ehrapy.preprocessing._quality_control.calculate_qc_metrics` for a list of calculated metrics.
 
     Args:
+        mtx: Data array.
+        obs_metrics: DataFrame to store observation metrics.
```
eroell, line 107: Not an argument, right?
```suggestion
```

ehrapy/preprocessing/_quality_control.py
```diff
@@ -55,55 +63,59 @@ def qc_metrics(
...
+        var_metrics: DataFrame to store variable metrics.
```
eroell, line 108: Not an argument, right?
```suggestion
```

ehrapy/preprocessing/_quality_control.py
```diff
@@ -136,53 +148,70 @@ def _obs_qc_metrics(
...
+            modified_slice = np.tile(
                 np.where(
                     adata.obs[original_values_categorical].astype(object) == "nan",
                     np.nan,
                     adata.obs[original_values_categorical].astype(object),
                 ).reshape(-1, 1),
                 mtx[:, index].shape[1],
             )
+
+            # Concatenate the modified slice with the unchanged part of mtx
```
eroell, line 182: The concatenate operation also copies the data in fact..
I think the behavior might not be working, as the column order is actually relevant, but below the columns are shuffled, right? I might also be missing a point right now about what index is, but think the original order of columns in mtx is not maintained

eroell: It could also be beyond the scope of this PR to make this actually more efficient, and rather have a second issue for this, and address it specifically - if fine with you @aGuyLearning , we can revert this to the deepcopy, and open a new issue!

tests/preprocessing/test_quality_control.py
```diff
@@ -82,19 +86,63 @@ def test_var_qc_metrics(missing_values_adata):
     assert (~var_metrics["iqr_outliers"]).all()
 
 
+@pytest.mark.parametrize(
+    "array_type, expected_error",
+    [
+        (np.array, None),
+        (as_dense_dask_array, None),
+        # (sparse.csr_matrix, NotImplementedError), TODO: Dataset contains string values. Sparse matrices not supported.
```
eroell, line 94:
```suggestion
        # can't test sparse matrices because they don't support string values
```

tests/preprocessing/test_quality_control.py
```diff
@@ -82,19 +86,63 @@ def test_var_qc_metrics(missing_values_adata):
...
+        with pytest.raises(expected_error):
+            _compute_var_metrics(mtx, adata)
+
+
+def test_var_encoding_mode_does_not_modify_original_matrix():
+    adata = read_csv(dataset_path=f"{_TEST_PATH_ENCODE}/dataset1.csv")
+    adata2 = encode(adata, encodings={"one-hot": ["clinic_day"]})
+    mtx_copy = adata.X.copy()
+    _compute_var_metrics(adata2.X, adata2)
+    assert np.array_equal(mtx_copy, adata.X)  # check that the original matrix is not modified
```
eroell, line 137: Hm you're not doing anything with mtx_copy and adata here so yeah they should not change. But it misses the relevant check for making sure adata2.X is not being altered

eroell, line 137:
```suggestion
    assert np.array_equal(mtx_copy, adata.X)
```

tests/preprocessing/test_quality_control.py
```diff
@@ -82,19 +86,63 @@ def test_var_qc_metrics(missing_values_adata):
...
+def test_var_qc_metrics_array_types(array_type, expected_error):
+    adata = read_csv(dataset_path=f"{_TEST_PATH_ENCODE}/dataset1.csv")
+    adata.X = array_type(adata.X)
+    mtx = adata.X
+    if expected_error:
+        with pytest.raises(expected_error):
+            _compute_var_metrics(mtx, adata)
+
+
+def test_var_encoding_mode_does_not_modify_original_matrix():
```
eroell, line 132: For this test, and all other tests of functions that now support both numpy arrays and dask arrays: lets also add a test parametrization for all tests, to check that they do work with both supported array types.

ehrapy/preprocessing/_quality_control.py
aGuyLearning: okay, i will revert it.

tests/preprocessing/test_quality_control.py
aGuyLearning: my bad, that was an oversight.

tests/preprocessing/test_quality_control.py
aGuyLearning: The problem here is, that encode() does not yet support dask arrays

aGuyLearning: While going over the functions supporting multiple array types i found an issue with _replace_explicit(). The dask implementation uses dask.array.isnan function.
dask.array.isnan does NOT support object arrays — it works on numerical types (like float).
So this causes the entrie array to be handled like objects in case of a string being found.

tests/preprocessing/test_quality_control.py
```diff
@@ -8,70 +8,55 @@
...
-@pytest.fixture
-def lab_measurements_simple_adata(obs_data, var_data):
-    X = np.array([[73, 0.02, 1.00], [148, 0.25, 3.55]], dtype=np.float32)
-    return AnnData(
-        X=X,
-        obs=pd.DataFrame(data=obs_data),
-        var=pd.DataFrame(data=var_data, index=["Acetaminophen", "Acetoacetic acid", "Beryllium, toxic"]),
-    )
-
+    # var_metrics tests
```
eroell, line 29:
```suggestion
```

PR #867, opened 2025-02-17
Pass all non-nan features along desired var_names to impute (KNN)

nicolassidoux:
* When the user passes specific features to impute with KNN, pass also all numerical features without missing values to the imputer.
* Check only features to impute for numeric types.

### DIFF
--- a/ehrapy/anndata/anndata_ext.py
+++ b/ehrapy/anndata/anndata_ext.py
@@ -325,15 +325,21 @@ def get_numerical_column_indices(
 
     return [idx for idx in indices if idx not in non_numerical_indices]
 
+
 def get_fully_imputed_column_indices(
     adata: AnnData, layer: str | None = None, column_indices: Iterable[int] | None = None
 ) -> list[int]:
     adata_layer = adata.X if layer is None else adata.layers[layer]
 
-    indices = range(adata_layer.shape[1]) if column_indices is None else [i for i in column_indices if i < adata_layer.shape[1]]
+    indices = (
+        range(adata_layer.shape[1])
+        if column_indices is None
+        else [i for i in column_indices if i < adata_layer.shape[1]]
+    )
     mask = ~np.isnan(adata_layer[:, indices]).any(axis=0)
 
     return np.array(indices)[mask].tolist()
+
 
 def get_column_indices(adata: AnnData, col_names: str | Iterable[str]) -> list[int]:
     """Fetches the column indices in X for a given list of column names
--- a/ehrapy/anndata/anndata_ext.py
+++ b/ehrapy/anndata/anndata_ext.py
@@ -305,19 +305,32 @@ def move_to_x(adata: AnnData, to_x: list[str] | str, copy_x: bool = False) -> An
 
     return new_adata
 
-def get_numerical_column_indices(adata: AnnData, layer: str | None = None, column_indices: Iterable[int] | None = None) -> list[int]:
+def get_numerical_column_indices(adata: AnnData, layer: str | None = None,
+                                     column_indices: Iterable[int] | None = None) -> list[int]:
     adata_layer = adata.X if layer is None else adata[layer]
-    indices = [i for i in range(adata_layer.shape[1])] if column_indices is None else [i for i in column_indices if i < adata_layer.shape[1] - 1]
+    indices = [i for i in range(adata_layer.shape[1])] if column_indices is None else [i for i in column_indices if
+                                                                                       i < adata_layer.shape[1] - 1]
+    non_numerical_indices = []
     for i in indices:
         # Todo: There is probably a better way to do this!
         try:
-            adata.X[::, i].astype("float64")
+            adata_layer[::, i].astype("float64")
         except ValueError:
-            indices.remove(i)
-    return indices
+            non_numerical_indices.append(i)
 
-def get_fully_imputed_column_indices(adata: AnnData, layer: str | None = None, col_indices: Iterable[int] | None = None) -> list[int]:
-    pass
+    return [idx for idx in indices if idx not in non_numerical_indices]
+
+def get_fully_imputed_column_indices(adata: AnnData, layer: str | None = None,
+                                         column_indices: Iterable[int] | None = None) -> list[int]:
+    adata_layer = adata.X if layer is None else adata[layer]
+    indices = [i for i in range(adata_layer.shape[1])] if column_indices is None else [i for i in column_indices if
+                                                                                       i < adata_layer.shape[1] - 1]
+    missing_values_indices = []
+    for i in indices:
+        if np.isnan(adata_layer[::, i].astype("float64")).any():
+            missing_values_indices.append(i)
+
+    return [idx for idx in indices if idx not in missing_values_indices]
 
 def get_column_indices(adata: AnnData, col_names: str | Iterable[str]) -> list[int]:
     """Fetches the column indices in X for a given list of column names
--- a/ehrapy/anndata/anndata_ext.py
+++ b/ehrapy/anndata/anndata_ext.py
@@ -309,17 +309,19 @@ def move_to_x(adata: AnnData, to_x: list[str] | str, copy_x: bool = False) -> An
 def get_numerical_column_indices(
     adata: AnnData, layer: str | None = None, column_indices: Iterable[int] | None = None
 ) -> list[int]:
-    adata_layer = adata.X if layer is None else adata[layer]
+    mtx = adata.X if layer is None else adata[layer]
     indices = (
-        list(range(adata_layer.shape[1]))
+        list(range(mtx.shape[1]))
         if column_indices is None
-        else [i for i in column_indices if i < adata_layer.shape[1] - 1]
+        else [i for i in column_indices if i < mtx.shape[1] - 1]
     )
     non_numerical_indices = []
     for i in indices:
-        # Todo: There is probably a better way to do this!
+        # The astype("float64") call will throw only if the feature’s data type cannot be cast to float64, meaning in
+        # practice it contains non-numeric values. Consequently, it won’t throw if the values are numeric but stored
+        # as an "object" dtype, as astype("float64") can successfully convert them to floats.
         try:
-            adata_layer[::, i].astype("float64")
+            mtx[::, i].astype("float64")
         except ValueError:
             non_numerical_indices.append(i)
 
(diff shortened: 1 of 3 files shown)


ehrapy/anndata/anndata_ext.py
```diff
@@ -306,6 +306,43 @@ def move_to_x(adata: AnnData, to_x: list[str] | str, copy_x: bool = False) -> An
     return new_adata
 
 
+def get_numerical_column_indices(
```
Zethson, line 309: Isn't that already stored by https://ehrapy.readthedocs.io/en/stable/usage/anndata/ehrapy.anndata.infer_feature_types.html at this point? Do we need this function?

ehrapy/anndata/anndata_ext.py
```diff
@@ -306,6 +306,43 @@ def move_to_x(adata: AnnData, to_x: list[str] | str, copy_x: bool = False) -> An
...
+        # Todo: There is probably a better way to do this!
+        try:
+            adata_layer[::, i].astype("float64")
+        except ValueError:
+            non_numerical_indices.append(i)
+
+    return [idx for idx in indices if idx not in non_numerical_indices]
+
+
+def get_fully_imputed_column_indices(
```
Zethson, line 329: This could probably be implemented a bit more nicely with numpy

Something like

```python
def get_fully_imputed_column_indices(
    adata: AnnData, layer: str | None = None, column_indices: Iterable[int] | None = None
) -> list[int]:
    adata_layer = adata.X if layer is None else adata.layers[layer]

    indices = range(adata_layer.shape[1]) if column_indices is None else [i for i in column_indices if i < adata_layer.shape[1]]
    mask = ~np.isnan(adata_layer[:, indices]).any(axis=0)

    return np.array(indices)[mask].tolist()
```

ehrapy/preprocessing/_imputation.py
```diff
@@ -290,13 +281,19 @@ def _knn_impute(
...
-    if isinstance(var_names, Iterable) and all(isinstance(item, str) for item in var_names):
-        column_indices = get_column_indices(adata, var_names)
-        adata.X[::, column_indices] = imputer.fit_transform(adata.X[::, column_indices])
-        # this is required since X dtype has to be numerical in order to correctly round floats
-        adata.X = adata.X.astype("float64")
-    else:
-        adata.X = imputer.fit_transform(adata.X)
+    column_indices = get_column_indices(adata, adata.var_names if var_names is None else var_names)
+    numerical_indices = get_numerical_column_indices(adata)
+    if any(idx not in numerical_indices for idx in column_indices):
```
Zethson, line 286: I wonder whether this part should much rather refer to `infer_feature_types` and how we store that there. I think we should reuse and suggest to fix the `adata.var[‘feature_type’]` slot

ehrapy/preprocessing/_imputation.py
```diff
@@ -290,13 +281,19 @@ def _knn_impute(
...
+        raise ValueError(
+            "Can only impute numerical data. Try to restrict imputation to certain columns using "
+            "var_names parameter or perform an encoding of your data."
+        )
+    fully_imputed_indices = get_fully_imputed_column_indices(adata, column_indices=numerical_indices)
+    imputer_data_indices = column_indices + [i for i in fully_imputed_indices if i not in column_indices]
+    imputer_x = adata.X[::, imputer_data_indices].astype("float64")
+    adata.X[::, imputer_data_indices] = imputer.fit_transform(
+        imputer_x
+    )  # Todo: will cast all involved features to float64, is that desired? we *could* restrict the copy to only the imputed columns...
```
Zethson, line 296: Should we always use float64 or float32?

ehrapy/preprocessing/_imputation.py
```diff
@@ -452,28 +449,24 @@ def mice_forest_impute(
...
-            raise ValueError(
-                "Can only impute numerical data. Try to restrict imputation to certain columns using "
-                "var_names parameter."
-            )
-
-    except ValueError as e:
-        if "Data matrix has wrong shape" in str(e):
-            logger.warning("Check that your matrix does not contain any NaN only columns!")
-        raise
+    if any(
```
Zethson, line 452: See comments about `infer_feature_types` and `adata.var[‘feature_type’]`

Zethson (changes requested):

ehrapy/anndata/anndata_ext.py
nicolassidoux: Well technically, yes. For features inferred as numerical, no problem. But for those inferred as categorical, those whose underlying type is numerical could also be used! So either I check for the datatype being each categorical (and then I need an extra step) or I drop all categorical, and we miss an opportunity to improve the imputation.

ehrapy/anndata/anndata_ext.py
nicolassidoux: I trust you on that!

ehrapy/preprocessing/_imputation.py
nicolassidoux: Ohh great question. If we want the "one-fits-all" approach, higher precision is better? But I wonder if we shouldn't stick to the original dataset type, casting back in the process (and round, truncate whatever if the original type is int)...

Zethson: Yes, I think whenever possible we should use the original dtype and not randomly cast around.

ehrapy/anndata/anndata_ext.py
nicolassidoux: Practical case with the `diabetes_130_fairlearn` dataset:
image
Here `num_procedures` was inferred as categorical, meaning I would drop it even if it's numerical under the hood...
Sure, it's a specific example, but if it happened once...

ehrapy/preprocessing/_imputation.py
Zethson: If you do `adata = ep.dt.diabetes_130_fairlearn(encoded=True)` it works. It is actually correct here because the values in `X` are more or less only strings.

ehrapy/anndata/anndata_ext.py
Zethson: I see that would this mean that your code is better at inferring types than the existing code?

I m also a bit wary here because we re introducing custom code that not all imputers are sharing. Also, do or should we even support the imputation of unencoded values? Probably because one hot encoded columns are also ugly to impute?

Playing devils advocate here. Please forgive me 😅😊😊

Zethson: @Lilly-May may I politely ask for some opinions of yours here, please?

Lilly-May: The first question that comes to mind is: Why is `num_procedures` detected as categorical in the example? My guess is that it has values ranging from 0 to X without any intermediate values missing, so the `infer_feature_type` method assumes it is a label-encoded categorical feature.

Irrespective of that, I think it makes sense to have another method that specifically checks for numerical columns here, as storing a value numerically does not necessarily mean it is a numerical feature. We also want to be able to impute encoded categorical values, which will always have the feature type "categorical," even when encoded (and therefore stored numerically) by ehrapy. So if we rely solely on the features detected by `infer_feature_type`, we’ll never be able to impute encoded categorical values, no?

nicolassidoux: Hi @Lilly-May, first time interacting with you here! Thanks for the input!

> We also want to be able to impute encoded categorical values, which will always have the feature type "categorical," even when encoded (and therefore stored numerically) by ehrapy. 

True, I just checked.

> So if we rely solely on the features detected by infer_feature_type, we’ll never be able to impute encoded categorical values, no?

Indeed... Which gives us two alternatives:
1. We keep this PR implementation. I already made the other imputations method consistent with the check for numerics with `get_numerical_column_indices`, whenever applicable.
2. The user has to provide an encoded dataset (only floats). But even there it would be nice to check for the types stored in X and raise a nice `ValueError("Encode your data")`, which means we still need a kind of `get_numerical_column_indices` at some point.

I would suggest to for 1 now, then switch to 2 later as a design decision.

ehrapy/preprocessing/_imputation.py
nicolassidoux: True! I added this Todo because of `fknni/faiss/faiss.py/fit_transform()`:
```python
if not np.issubdtype(X.dtype, np.floating):
     self.X_full = np.asarray(X, dtype=np.float32)
     ...
     return self.X_full
```
But actually it is more a fknni problem, not ehrapy.

ehrapy/preprocessing/_imputation.py
nicolassidoux: Resolving for clarity as we have a running discussion about that above.

ehrapy/anndata/anndata_ext.py
Zethson: All right - I trust your judgement. Then let's do this.

ehrapy/anndata/anndata_ext.py
```diff
@@ -306,6 +306,41 @@ def move_to_x(adata: AnnData, to_x: list[str] | str, copy_x: bool = False) -> An
...
+) -> list[int]:
+    adata_layer = adata.X if layer is None else adata[layer]
+    indices = (
+        list(range(adata_layer.shape[1]))
+        if column_indices is None
+        else [i for i in column_indices if i < adata_layer.shape[1] - 1]
+    )
+    non_numerical_indices = []
+    for i in indices:
+        # Todo: There is probably a better way to do this!
```
Zethson, line 320:
```
non_numerical_indices = [i for i in indices if not np.issubdtype(adata_layer[:, i].dtype, np.number)]
```

is cleaner

nicolassidoux: Indeed, but doesnt work if dtype is "object" (and here we go again with the encode/not encode discussion).

nicolassidoux: This code:
```python
try:
     adata_layer[::, i].astype("float64")
except ValueError:
...
```
was the only I found that works for everything.

Zethson: It's very pythonic actually haha. I'll think about it again.

Zethson: Can't think about it atm but could you please add one comment justifying why we do it like this? So that no one suggests code like me again that does not take the object case into account.

ehrapy/anndata/anndata_ext.py
```diff
@@ -306,6 +306,41 @@ def move_to_x(adata: AnnData, to_x: list[str] | str, copy_x: bool = False) -> An
...
+        except ValueError:
+            non_numerical_indices.append(i)
+
+    return [idx for idx in indices if idx not in non_numerical_indices]
+
+
+def get_fully_imputed_column_indices(
+    adata: AnnData, layer: str | None = None, column_indices: Iterable[int] | None = None
+) -> list[int]:
+    adata_layer = adata.X if layer is None else adata.layers[layer]
```
Zethson, line 332:
```suggestion
    mtx = adata.X if layer is None else adata.layers[layer]
```

adata_layer could also be a layer key. I think we usually call that mtx.

PR #877, opened 2025-03-07
Improve Tutorial Clarity, Consistency, and Documentation Links

aGuyLearning:
**PR Checklist**

- [ ] This comment contains a description of changes (with reason)
- [ ] Referenced issue is linked
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

**Description of changes**
This pull request includes several changes to improve the clarity and consistency of the text in the tutorials. The most important changes include standardizing the formatting of code identifiers, and improving the readability of certain sections. As well as linking to the documentation for mentioned functions and classes.

PR #880, opened 2025-03-12
Clean up Round - cut down anndata extension functionality

eroell:
**PR Checklist**

- [ ] Referenced issue is linked n/a

**Description of changes**
The module [ehrapy.anndata](https://github.com/theislab/ehrapy/blob/main/ehrapy/anndata/) has many things from the early stage of ehrapy which have been modernized in the meantime. It also has accumulated duplication of functionality. This PR is a cleanup round focusing on this module, but affects also different other parts.
 

**Technical details**
A list of functions in there at the time of opening the PR, and what to do with them:
- [ ] `df_to_anndata`: port to ehrdata (future)
- [ ] `anndata_to_df`: I think we should maintain this with the key additional service that the columns of the generated df should be typed based on the feature datatypes of `infer_feature_types`. We also need this internally, see e.g. changes to ols [here](https://github.com/theislab/ehrapy/pull/877).
- [ ] `move_to_obs`: delete if possible 
- [ ] `_get_var_indices_for_type`: Keep this as ground truth lookup function index of `infer_feature_type` generated types. Move to there?
- [x] `delete_from_obs`: delete, can be done via pandas
- [ ] `move_to_x`: into rank_features_groups where needed, else delete?
- [x] `get_numerical_column_indices`: Delete, replace with `_get_var_indices_for_type`?
- [x] `get_fully_imputed_column_indices`: removed to single place where its used; re-consider spec of knn in future
- [x] `get_column_indices`: delete?
- [x] `_assert_encoded`: delete? I can see the information can be nice to have, but I think we in fact do not need this.
- [x] `get_numeric_vars`: delete, use `_get_var_indices_for_type` directly
- [x] `assert_numeric_vars`: I vote for keeping this, it can be handy in its use of e.g. the normalization methods. Move to somewhere else? Also, should become private.
- [x] `set_numeric_vars` delete
- [x] `_detect binary columns` delete
- [ ]  `_cast_obs_columns`: helper of `move_to_obs`, delete if that one is deleted
- [x] `generate_anndata: deleted - never used
- [x] `get_obs_df`: move to proper get module
- [x] `get_var_df`: move to proper get module
- [x] `get_rank_features_df`: move to proper get module
- [x] `NotEncodedError`: Delete with `_assert_encoded`
- [x] `_are_ndarrays equal`: port to tests where its used? 
- [x] `_is_val_missing`: port to tests where its used? I do see it could be a reference of how we consider arrays to be equal or not, though, with the empty string?
- [x] `_to_dense`:  delete (use from dask array utils once its released)

**Additional context**


Zethson:
Sorry, my recent changes led to conflicts here. When you're back, let's try to get this PR in. I want to make a release then to have the new doc design out. Then we take a 3-4 month window where we make substantial changes and integrate ehrdata.

PR #881, opened 2025-03-13
Links to documentation in CohortTracker Tutorial with nbsphinx

eroell:
test rtd build for nbsphinx link

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -97,7 +97,6 @@ intersphinx_mapping = {
     "jax": ("https://jax.readthedocs.io/en/latest/", None),
     "lamin": ("https://lamin.ai/docs", None),
     "lifelines": ("https://lifelines.readthedocs.io/en/latest/", None),
-    "ehrapy": ("https://ehrapy.readthedocs.io/en/latest/", None),
 }
 
 language = "en"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -8,7 +8,6 @@ from pathlib import Path
 
 HERE = Path(__file__).parent
 sys.path[:0] = [str(HERE.parent), str(HERE / "extensions")]
-sys.path.append(str(HERE / "tutorials" / "notebooks"))
 
 needs_sphinx = "4.3"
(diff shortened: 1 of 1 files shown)


eroell:
Any instant ideas @Zethson?

docs/conf.py
```diff
@@ -8,6 +8,7 @@
 
 HERE = Path(__file__).parent
 sys.path[:0] = [str(HERE.parent), str(HERE / "extensions")]
+sys.path.append(str(HERE / "tutorials" / "notebooks"))
```
flying-sheep, line 11: what’s this for? `sys.path` is for importing python packages, why would you want to do that from the notebook folder?

flying-sheep:
OK, so you’re using `nbsphinx`’ thumbnail gallery which doesn’t exist for myst-nb, otherwise I’d have recommended replacing it for simplicity.

flying-sheep:
OK, got it, the issue is the `cell_type`. If `"cell_type": "markdown"`, nbsphinx won’t interpret rST syntax as rST. You need `"cell_type": "raw"`.

docs/conf.py
eroell: A leftover thing I wanted to delete again!

eroell: Will remove

eroell:
Awesome, this works!
Thanks soo much @flying-sheep!

To summarize what @flying-sheep did here:
a) set the notebook cell type to raw 
b) set the cell metadata "raw_mimetype" to "text/restructuredtext"

Should look like this:
```
   "cell_type": "raw",
   "metadata": {
     "raw_mimetype": "text/restructuredtext"
   },
```

flying-sheep:
Yeah! nbsphinx has docs for this, but they assume you can just do that from UI and don’t have to manually hack around in JSON: https://nbsphinx.readthedocs.io/en/0.2.7/raw-cells.html

I filed https://github.com/microsoft/vscode-jupyter/issues/16490

PR #897, opened 2025-05-17
Add EHRData transition code

Zethson:
We'll use that decorator later as we transition to EHRData


Zethson:
```
    from ehrdata import EHRData
E   ImportError: cannot import name 'EHRData' from 'ehrdata' (/opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/ehrdata/__init__.py)
```

I assume that the latest ehrdata on pypi doesn't have it yet?

@eroell we need a release.

PR #918, opened 2025-06-17
Fix/coverage

eroell:
**PR Checklist**

- [ ] This comment contains a description of changes (with reason) N/A

**Description of changes**
Reactivate coverage check

**Technical details**

**Additional context**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,10 @@ jobs:
                   MPLBACKEND: agg
                   PLATFORM: ${{ matrix.os }}
                   DISPLAY: :42
-              run: pytest -v --cov --color=yes
+              run: pytest -v --cov=ehrapy --cov-report=xml --color=yes
 
             - name: Upload coverage
-              uses: codecov/codecov-action@v3
+              if: matrix.python == '3.11'
+              uses: codecov/codecov-action@v5
+              with:
+                token: ${{ secrets.CODECOV_TOKEN }}
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,4 +1,8 @@
+codecov:
+  require_ci_to_pass: false
+
 comment: false
+
 coverage:
     status:
         project:
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,6 +8,6 @@ coverage:
             default:
                 enabled: no
     ignore:
-        - "test/"
+        - "tests/"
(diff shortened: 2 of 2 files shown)


codecov.yml
```diff
@@ -1,13 +1,17 @@
+codecov:
+  require_ci_to_pass: false
```
Zethson: The template uses "no" here. https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.codecov.yaml

Is this the same?

codecov.yml
```diff
@@ -1,13 +1,17 @@
...
+
 comment: false
+
 coverage:
     status:
         project:
             default:
-                target: auto
+                target: 1 # always succeed
```
Zethson: The comment in https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.codecov.yaml#L9 is a tiny bit better as it explains why it always succeeds.

codecov.yml
```diff
@@ -1,13 +1,17 @@
...
         patch:
```
Zethson, line 12: https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.codecov.yaml#L9

Should we also adapt what the template has here?

.github/workflows/test.yml
```diff
@@ -50,7 +50,10 @@ jobs:
                   MPLBACKEND: agg
                   PLATFORM: ${{ matrix.os }}
                   DISPLAY: :42
-              run: pytest -v --cov --color=yes
+              run: pytest -v --cov=ehrapy --cov-report=xml --color=yes
 
             - name: Upload coverage
-              uses: codecov/codecov-action@v3
+              if: matrix.python == '3.11'
```
Zethson:
```suggestion
              if: matrix.python == '3.13'
```

Maybe as 3.13 should technically finish before 3.11 (but funnily enough in practice often doesn't?) and therefore can earlier execute the additional codecov work

Zethson (approved):

codecov.yml
eroell: I think but a check led me rather to internet arguments and confusion than the rule.
I'll stick to the scverse template, going to `no`

PR #919, opened 2025-06-23
tests for more plots

sueoglu:
Adding more tests for plotting functionalities 

- created tests for the plots in `missing_pl_api `and `scanpy_pl_api `
- generated reference images .png per plotting function

Plotting functions checklist:


tests/plot/test_missingno_pl_api.py
```diff
@@ -0,0 +1,53 @@
...
+    fig = plot.figure
+
+    check_same_image(
+        fig=fig,
+        base_path=f"{_TEST_IMAGE_PATH}/missing_values_matrix",
+        tol=2e-1,
+    )
+
+
+def test_heatmap(mimic_2, check_same_image):
```
eroell: Having the tested function's name in the test function can be helpful, as this function's name will be shown when the test fails. So in this case, the name `test_missing_values_heatmap` would be a good one

eroell:
OK - Run Notebooks won't run since its from a fork; and `ubuntu-latest 3.13, --pre` is failing, no big deal here.

But `ubuntu-latest, 3.11` and `ubuntu-latest 3.13` should be green - and give you right now hopefully helpful error messages on the plotting tests

tests/conftest.py
```diff
@@ -123,6 +123,27 @@ def adata_mini():
...
+            "readmitted",
+            "readmit_binary",
+            "discharge_disposition_id",
+        ]
+    )[:200]
+    return adata
+
+
+@pytest.fixture
+def mimic_2_sample():
```
Zethson: Maybe name that more clearly? Above you also have `dataset_sample` but you subsample down to 200 observations. Here, you pick a few columns. Maybe get the column names into the name of the fixture.

.readthedocs.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 build:
   os: ubuntu-24.04
   tools:
-    python: "3.13"
+    python: "3.12" # dowhy breaks for Python 3.13https://github.com/py-why/dowhy/pull/1328
```
Zethson: The PR has been merged and a release has been made.
```suggestion
    python: "3.13"
```

Zethson: It might still not work but the comment is a bit outdated.

tests/plot/test_scanpy_pl_api.py
```diff
@@ -0,0 +1,633 @@
...
+import os
+
+import numpy as np
+
+import ehrapy as ep
+
+CURRENT_DIR = Path(__file__).parent
+_TEST_IMAGE_PATH = f"{CURRENT_DIR}/_images"
+
+# Set NUMBA_CPU_NAME to "generic" before importing or using numba
```
Zethson: But why? The comment should explain WHY we're doing this because the code already tells us that we're doing just that.

Zethson (commented):
Thank you! Great work.

The pseudotime notebook has an error. Should we keep it with errors or remove it?

tests/plot/test_scanpy_pl_api.py
```diff
@@ -0,0 +1,633 @@
...
+
+    fig = plt.gcf()
+    check_same_image(
+        fig=fig,
+        base_path=f"{_TEST_IMAGE_PATH}/dpt_timeseries",
+        tol=2e-1,
+    )
+
+
+"""
```
eroell: fyi, at the end we'd try to get rid of commented out code :)

tests/plot/test_missingno_pl_api.py
```diff
@@ -0,0 +1,54 @@
...
+    fig.subplots_adjust(left=0.1, right=0.95, top=0.9, bottom=0.15)
+
+    check_same_image(
+        fig=fig,
+        base_path=f"{_TEST_IMAGE_PATH}/missing_values_barplot",
+        tol=2e-1,
+    )
+
+
+def test_missing_values_matrixplot(mimic_2, check_same_image):
```
eroell, line 24: Constructive appreciation: I love tests like these. 
- the test purpose is straightforward understandable, and the code very concise
- the name of the test is on point
- using the dataset and the complex, general test function via a fixture is very neat and readable

tests/plot/test_scanpy_pl_api.py
```diff
@@ -0,0 +1,549 @@
...
+    [
+        "rank_features_groups",
+        "rank_features_groups_violin",
+        "rank_features_groups_stacked_violin",
+        "rank_features_groups_matrixplot",
+        "rank_features_groups_tracksplot",
+    ],
+)
+def test_rank_features_groups_plots(mimic_2_encoded, plotter):
+    # in scanpy tests, there is no testing for rank_genes_groups plot but for tool function,
```
eroell, line 240: Btw did you figure this out in a discussion? Or is there an open issue? If there is an open issue, adding the link to it hear would be neat

tests/plot/test_scanpy_pl_api.py
```diff
@@ -0,0 +1,549 @@
...
+        ],
+    ].copy()
+
+    adata.obs["service_unit"] = pd.Categorical(
+        adata.obs["service_unit"],
+        categories=sorted(pd.unique(adata.obs["service_unit"])),
+        ordered=True,
+    )
+
+    # compute rankings
```
eroell: this comment can be dropped, its almost the name of the function

eroell:
I see this is right now 1 failing test from a complete CI pass; and that is the pre-release test of `ep.pl.diffmap`. I noticed that this seems to be [not tested](https://github.com/scverse/scanpy/blob/9dceb9e1ba5f2d62eea75c4ecc2ad6eb5a2f0df0/tests/test_plotting.py#L1176) in scanpy...

I'm inclined towards opening an issue there, and leave the test out here with a comment to this issue

PR #937, opened 2025-09-04
naive implementation of balanced sampling

### LINKED ISSUE #925: Replace imbalanced-learn with a native implementation
### Description of feature

Imbalanced-learn is often behind scikit-learn releases and breaks. We're only using it for `balanced_sample` which we should be able to implement on our own

sueoglu:
fixes #925 

- implemented a native balanced sample function instead of imbalanced-learns function


eroell:
A first review;
- [ ] replace the code in (balanced_sample)[https://github.com/theislab/ehrapy/blob/94487752bdc367f2c6d49ea1f8cdf6b4dfaf23e3/ehrapy/preprocessing/_balanced_sampling.py#L13], rather than creating a new function `resample_adata`; ideally, there is no change in arguments and functionality (except for a different random seed) visible to users.
- [ ] fix mypy error - but it might even disappear if you stick to the original function name and arguments as above.
- [ ] tests; if you replace the code in `balanced_sample`, it's [tests](https://github.com/theislab/ehrapy/blob/94487752bdc367f2c6d49ea1f8cdf6b4dfaf23e3/tests/preprocessing/test_balanced_sampling.py#L17) should still pass for your new implementation - you can use them also to debug your new code
- [ ] Readthedocs build: I'm on it 

Sidenote: as of now, the mypy error would have been a "helpful" one, it uncovered an edgecase.

ehrapy/preprocessing/_balanced_sampling_native.py
```diff
@@ -0,0 +1,110 @@
...
+from typing import TYPE_CHECKING, Literal
+
+if TYPE_CHECKING:
+    from anndata import AnnData
+
+import numpy as np
+from anndata import AnnData
+
+
+def random_resample(
```
eroell: this might go away when you move your code to replace the current implementation of `balanced_sample`, but as a quick headsup: when we write "helper" functions that users don't see, its common to make them `_function_name(...)`, that is, with an underscore.

ehrapy/preprocessing/_balanced_sampling.py
```diff
@@ -1,83 +0,0 @@
...
-
-if TYPE_CHECKING:
-    from anndata import AnnData
-
-from anndata import AnnData
-from imblearn.over_sampling import RandomOverSampler
-from imblearn.under_sampling import RandomUnderSampler
-
-
-def balanced_sample(
```
eroell, line 13: To not break immediately, could you keep this function header, while calling just `sample` under the hood, and add a deprecation warning that this function has been renamed?
It can be removed from the documentation as you did already

ehrapy/preprocessing/_balanced_sampling.py
```diff
@@ -1,83 +0,0 @@
-from __future__ import annotations
-
-from typing import TYPE_CHECKING, Literal
-
-if TYPE_CHECKING:
-    from anndata import AnnData
-
-from anndata import AnnData
-from imblearn.over_sampling import RandomOverSampler
```
eroell, line 9: imblearn can also be removed from the pyproject.toml file then

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -128,20 +131,74 @@ def subsample(
...
         copy: If an :class:`~anndata.AnnData` is passed, determines whether a copy is returned.
+        balanced: If `True`, balance the groups in `adata.obs[key]` by under- or over-sampling.
+                  Requires `key` to be set. If `False`, simple random sampling is performed.
+        method: The sampling method, either "under" for under-sampling or "over" for over-sampling. Only relevant if `balanced=True`.
+        key: Key in `adata.obs` to use for balancing the groups. Only relevant if `balanced=True`.
+        replace: If `True`, samples are drawn with replacement. Only relevant if `balanced=False`.
+        axis: Axis to sample on. Either `obs` / `0` (observations, default) or `var` / `1` (variables).
+        p: Drawing probabilities (floats) or mask (bools).
+            Either an `axis`-sized array, or the name of a column
+            If p is an array of probabilities, it must sum to 1.
```
eroell, line 172: An example similar to the one that `balanced_sample` had would be great

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -128,20 +131,74 @@ def subsample(
...
+
+
+def sample(
+    data: AnnData | np.ndarray,
+    fraction: float | None = None,
+    *,
+    n_obs: int | None = None,
+    rng: int | None = None,
+    balanced: bool = False,
+    method: Literal["under", "over"] = "under",
```
eroell: Might be more clear to call it explicitly "balanced_method", to make it as clear as possible that this belongs to the balanced sampling. What do you think?

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -128,20 +131,74 @@ def subsample(
...
+    key: str | None = None,
```
eroell: "balanced_key", same as for the comment above?

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -128,20 +131,74 @@ def subsample(
...
+    """Sample a fraction or a number of observations / variables with or without replacement.
 
     Args:
         data: The (annotated) data matrix of shape `n_obs` × `n_vars`. Rows correspond to observations (patients) and columns to features.
-        fraction: Subsample to this `fraction` of the number of observations.
-        n_obs: Subsample to this number of observations.
-        random_state: Random seed to change subsampling.
+        fraction: Sample to this `fraction` of the number of observations.
+        n_obs: Sample to this number of observations.
+        rng: Random seed to change subsampling.
```
eroell: rng: Random seed to change subsampling.
```suggestion
        rng: Random seed.
```

eroell (commented):
`balanced_sample` is also used in the `bias.ipynb` notebook, this can also be replaced there with this new function

sueoglu:
> `balanced_sample` is also used in the `bias.ipynb` notebook, this can also be replaced there with this new function

In the description there, I also remove the link to imblearn's RandomUnderSampler, right?

ehrapy/preprocessing/_scanpy_pp_api.py
sueoglu: Yes, that would be clearer!

sueoglu:
PR from the ehrapy-tutorials repo: https://github.com/theislab/ehrapy/pull/941

ehrapy/preprocessing/_balanced_sampling.py
```diff
@@ -1,83 +1,28 @@
...
-from imblearn.under_sampling import RandomUnderSampler
-
 
 def balanced_sample(
     adata: AnnData,
     *,
     key: str,
     random_state: int = 0,
-    method: Literal["RandomUnderSampler", "RandomOverSampler"] = "RandomUnderSampler",
+    method: Literal["under", "over"] = "under",
```
eroell: To ensure backwards compatibility, please stick with the old arguments, and then map them to the new arguments within the function body

PR #953, opened 2025-09-24
3D enabled implementation of ep.pp.filter_observations, ep.pp.filter_features 

### LINKED ISSUE #809: implement a "ts.filter_features()"
### Description of feature

similar as for static ones, e.g. min count, min persons, ...

sueoglu:
fixes #809 

Implemented and wrote basic tests for:

- ep.pp.filter_observations
- ep.pp.filter_features

Both functions allow users to filter data based on missing value thresholds for both 2D and 3D data structure.
It also includes:
- support for filtering across different layers
- handling of the time mode parameter: all, any, proportion 

Examples:
```
edata_1 = ed.dt.ehrdata_blobs(n_variables=45, n_observations=500, base_timepoints=15, missing_values=0.6)
edata_1.R.shape
(500, 45, 15)
ep.pp.filter_observations(edata_1, min_vars=10, time_mode="all")
edata_1.R.shape
(477, 45, 15)

edata_2 = ed.dt.ehrdata_blobs(n_variables=45, n_observations=500, base_timepoints=15, missing_values=0.6)
edata_2.R.shape
(500, 45, 15)
ep.pp.filter_features(edata_2, min_obs=185, time_mode="all")
edata_2.R.shape
(500, 18, 15)
```

### DIFF
--- a/ehrapy/preprocessing/_filter.py
+++ b/ehrapy/preprocessing/_filter.py
@@ -73,7 +73,7 @@ def filter_features(
         copy: Determines whether a copy is returned.
 
     Returns:
-        Depending on `copy`, subsets and annotates the passed data object and returns `None`
+        Depending on `copy`, subsets and annotates the passed data object and returns a filtered copy of the data object or acts in place
 
     Examples:
         >>> import ehrapy as ep
@@ -173,7 +173,7 @@ def filter_observations(
         copy: Determines whether a copy is returned.
 
     Returns:
-        Depending on `copy`, subsets and annotates the passed data object and returns `None`
+        Depending on `copy`, subsets and annotates the passed data object and returns a filtered copy of the data object or acts in place
 
     Examples:
         >>> import ehrapy as ep
@@ -255,7 +255,9 @@ def _arrays_for_filtering(data: EHRData | AnnData, layer: str | None) -> list[np
     return arrs
 
 
-def _compute_mask(arr: np.ndarray, *, min_count: int, max_count: int, time_mode: str, prop: float, axis: int, caller):
+def _compute_mask(
+    arr: np.ndarray, *, min_count: int, max_count: int, time_mode: str, prop: float, axis: int, caller
+) -> tuple[np.ndarray, np.ndarray, bool]:
     """Compute mask for filtering based on missing data thresholds.
 
     Returns:
--- a/tests/preprocessing/test_filter.py
+++ b/tests/preprocessing/test_filter.py
@@ -16,13 +16,13 @@ def ehr_3d_blobs(scope="module"):
 
 def test_filter_features_invalid_args(ehr_3d_blobs):
     # no threshold
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="at least one of 'min_obs' and 'max_obs"):
         ep.pp.filter_features(ehr_3d_blobs, time_mode="all", copy=False)
     # invalid time_mode
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="must be one of 'all', 'any', 'proportion'"):
         ep.pp.filter_features(ehr_3d_blobs, min_obs=185, time_mode="invalid_mode", copy=False)
     # invalid prop
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="prop must be set to a value between 0 and 1"):
         ep.pp.filter_features(ehr_3d_blobs, min_obs=185, time_mode="proportion", prop=3, copy=False)
 
 
@@ -36,7 +36,7 @@ def test_filter_features_invalid_args(ehr_3d_blobs):
     ],
 )
 def test_filter_features_min_max(request, fixture, shape, kwargs):
-    # generic test for min_obs and max_obs filtering on 2d and 3d data
+    """Generic test for min_obs and max_obs filtering on 2D and 3D data"""
     edata = request.getfixturevalue(fixture)
 
     layer = getattr(edata, shape)
@@ -52,7 +52,7 @@ def test_filter_features_min_max(request, fixture, shape, kwargs):
 
 def test_filter_features_layers(ehr_3d_blobs):
     edata = ehr_3d_blobs
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Invalid layer provided"):
         ep.pp.filter_features(edata, layer="invalid_layer", min_obs=185, time_mode="all", copy=False)
 
     layer_before = edata.layers["R_layer"].copy()
@@ -68,13 +68,13 @@ def test_filter_features_layers(ehr_3d_blobs):
 
 def test_filter_obs_invalid_args(ehr_3d_blobs):
     # no threshold
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="at least one of 'min_vars' and 'max_vars"):
         ep.pp.filter_observations(ehr_3d_blobs, time_mode="all", copy=False)
     # invalid time_mode
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="must be one of 'all', 'any', 'proportion'"):
         ep.pp.filter_observations(ehr_3d_blobs, min_vars=10, time_mode="invalid_mode", copy=False)
     # invalid prop
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="prop must be set to a value between 0 and 1"):
         ep.pp.filter_observations(ehr_3d_blobs, min_vars=10, time_mode="proportion", prop=2, copy=False)
 
 
@@ -88,7 +88,7 @@ def test_filter_obs_invalid_args(ehr_3d_blobs):
     ],
 )
 def test_filter_obs_min_max(request, fixture, shape, kwargs):
-    # generic test for min_obs and max_obs filtering on 2d and 3d data
+    """Generic test for min_vars and max_vars filtering on 2D and 3D data"""
     edata = request.getfixturevalue(fixture)
 
     layer = getattr(edata, shape)
@@ -104,7 +104,7 @@ def test_filter_obs_min_max(request, fixture, shape, kwargs):
 
 def test_filter_obs_layers(ehr_3d_blobs):
     edata = ehr_3d_blobs
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Invalid layer provided"):
         ep.pp.filter_features(edata, layer="invalid_layer", min_obs=185, time_mode="all", copy=False)
 
     layer_before = edata.layers["R_layer"].copy()
--- a/docs/api/preprocessing_index.md
+++ b/docs/api/preprocessing_index.md
@@ -24,6 +24,17 @@ Preprocessing steps usually perform a basic transformation on the data array in 
     preprocessing.clip_quantile
     preprocessing.summarize_measurements
 ```
+## Filtering
+
+```{eval-rst}
+.. autosummary::
+    :toctree: preprocessing
+    :nosignatures:
+
+    preprocessing.filter_features
+    preprocessing.filter_observations
+```
+
 
 ## Quality control
 
@@ -89,14 +100,4 @@ Partially overlaps with dataset integration. Note that a simple batch correction
     :nosignatures:
 
     preprocessing.neighbors
-```
-## Filtering
-
-```{eval-rst}
-.. autosummary::
-    :toctree: preprocessing
-    :nosignatures:
-
-    preprocessing.filter_features
-    preprocessing.filter_observations
 ```


ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+    time_mode: Literal["all", "any", "proportion"] = "all",
+    prop: float | None = None,
+    inplace: bool = True,
+    copy: bool = False,
+) -> AnnData | EHRData | tuple[CSBase, np.ndarray] | None:  # pragma: no cover
+    """Filter features based on counts or number of observations.
+
+    Keep only features which have at least `min_counts` counts or `min_obs` observations
+    or have at most `max_counts` counts or `max_obs` observations.
+    When an `EHRData` is passed, filtering can be done across time points.
```
eroell:
```suggestion
    When a longitudinal `EHRData` is passed, filtering can be done across time points.
```

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+def filter_features(
+    data: AnnData | EHRData | CSBase,
+    *,
+    min_counts: int | None = None,
+    min_obs: int | None = None,
+    max_counts: int | None = None,
+    max_obs: int | None = None,
+    time_mode: Literal["all", "any", "proportion"] = "all",
+    prop: float | None = None,
+    inplace: bool = True,
```
eroell: this function should also feature a `layers` argument

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+                    the feature must pass the filtering criteria in at least 30% of the time points.
+        prop: Proportion of time points in which the feature must pass the filtering criteria. Only relevant if `time_mode='proportion'`.
+        inplace: Performs the operation inplace or returns a copy.
+        copy: If an `EHRData` or `AnnData` is passed, determines whether a copy is returned. Is ignored otherwise.
+
+    Returns:
+        Depending on `inplace`, subsets and annotates the passed data object and returns `None` or returns the following arrays:
+            * feature_mask: A boolean array indicating which features are kept after filtering.
+            * number_per_feature: An array with the number of counts or observations per feature across time points.
+
```
eroell: Adding a 2D and a 3D example would be very helpful here

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+                      the observation must pass the filtering criteria in at least 30% of the time points.
+        prop: Proportion of time points in which the observation must pass the filtering criteria. Only relevant if `time_mode='proportion'`.
+        inplace: Performs the operation inplace or returns a copy.
+        copy: If an `EHRData` or `AnnData` is passed, determines whether a copy is returned. Is ignored otherwise.
+
+    Returns:
+        Depending on `inplace`, subsets and annotates the passed data object and returns `None` or returns the following arrays:
+            * obs_mask: A boolean array indicating which observations are kept after filtering.
+            * number_per_obs: An array with the number of counts or variables per observation across time points.
+
```
eroell: Also, an example here that shows how it can be used would be great

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+                max_counts=max_counts,
+                max_cells=max_obs,
+                inplace=inplace,
+                copy=copy,
+            )
+        else:
+            edata = data.copy() if copy else data
+
+            n_opts = sum(x is not None for x in [min_counts, min_obs, max_counts, max_obs])
+            if n_opts != 1:
```
eroell: Is a min_obs and a max_obs not allowed?

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
     )
 
 
+def filter_features(
+    data: AnnData | EHRData | CSBase,
+    *,
+    min_counts: int | None = None,
```
eroell: I believe the counts are specific to RNAseq processing, and basically sum up the numeric values of one feature.
For us, this argument should be removed, and never passed to the underlying scanpy function

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+    max_counts: int | None = None,
+    max_obs: int | None = None,
+    time_mode: Literal["all", "any", "proportion"] = "all",
+    prop: float | None = None,
+    inplace: bool = True,
+    copy: bool = False,
+) -> AnnData | EHRData | tuple[CSBase, np.ndarray] | None:  # pragma: no cover
+    """Filter features based on counts or number of observations.
+
+    Keep only features which have at least `min_counts` counts or `min_obs` observations
```
eroell: scanpy will consider `min_obs` as how many observations have a value > 0 in this feature. For us, I think it should be about how many values are not one of [ehrdata's missing values](https://github.com/theislab/ehrdata/blob/0756d831a6bc5c32e2e9b11abf259e8ce7a061ca/src/ehrdata/core/constants.py#L20)

I think hence that the scanpy function is not what we'd be using here after all in any case...
By not using it, the function could also be simplified to not take a CSMatrix as input, but only AnnData and EHRData objects, and renaming the function argument to `edata`.

I'm also interested to hear @Zethson on this larger comment

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+            min_counts=min_counts,
+            min_cells=min_obs,
+            max_counts=max_counts,
+            max_cells=max_obs,
+            inplace=inplace,
+            copy=copy,
+        )
+
+
+def filter_observations(
```
eroell: all comments from above apply here, too - but we can discuss them there, and then adjust this here alike

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+            obs_ax, _var_ax, _time_ax = 0, 1, 2
+
+            by_counts = (min_counts is not None) or (max_counts is not None)
+            threshold_min = min_counts if by_counts else min_obs
+            threshold_max = max_counts if by_counts else max_obs
+
+            if by_counts:
+                counts = np.nansum(edata.R, axis=obs_ax)
+            else:
+                present = (~np.isnan(edata.R)) & (edata.R != 0)
```
eroell: Please check for ehrdata's [missing values](https://github.com/theislab/ehrdata/blob/0756d831a6bc5c32e2e9b11abf259e8ce7a061ca/src/ehrdata/core/constants.py#L20); and 0, see larger comment above, we should consider as a real measurement in my opinion :)

eroell (changes requested):

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
     )
 
 
+def filter_features(
+    data: AnnData | EHRData | CSBase,
```
Zethson:
```suggestion
    data: EHRData,
```

We don't need to support anything else, do we?

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+    *,
+    min_counts: int | None = None,
+    min_obs: int | None = None,
+    max_counts: int | None = None,
```
Zethson: See Eljas comment above.

ehrapy/preprocessing/_scanpy_pp_api.py
Zethson: And didn't we at some point decide that we'll only have a `copy` argument and no longer `inplace`? How's the rest of the scanpy looking atm?

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+    max_obs: int | None = None,
+    time_mode: Literal["all", "any", "proportion"] = "all",
+    prop: float | None = None,
+    inplace: bool = True,
+    copy: bool = False,
+) -> AnnData | EHRData | tuple[CSBase, np.ndarray] | None:  # pragma: no cover
```
Zethson:
```suggestion
) -> EHRData | None:  # pragma: no cover
```

ehrapy/preprocessing/_scanpy_pp_api.py
Zethson: Yes I agree with you. I am not sure how much value this function already holds then to be honest. `min_count` and `max_count` however, don't really make a lot of sense.

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+    """Filter features based on counts or number of observations.
+
+    Keep only features which have at least `min_counts` counts or `min_obs` observations
+    or have at most `max_counts` counts or `max_obs` observations.
+    When an `EHRData` is passed, filtering can be done across time points.
+
+    Only provide either `min_counts` or `min_obs` and either `max_counts` or `max_obs` per call.
```
Zethson: The whole docstring must be adapted after the counts business is gone.

ehrapy/preprocessing/_scanpy_pp_api.py
Zethson: I don't think that we should always have a 2D and 3D example. We will have lots of duplications and long examples then. I'd advocate for 3D only. WDYT?

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+            min_cells=min_obs,
+            max_counts=max_counts,
+            max_cells=max_obs,
+            inplace=inplace,
+            copy=copy,
+        )
+
+
+def filter_observations(
+    data: AnnData | EHRData | CSBase,
```
Zethson:
```suggestion
    edata: EHRData,
```

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+    *,
+    min_counts: int | None = None,
```
Zethson: Same here. `counts` is omics business that we don't have here anymore. See above.

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+    min_vars: int | None = None,
+    max_counts: int | None = None,
+    max_vars: int | None = None,
+    time_mode: Literal["all", "any", "proportion"] = "all",
+    prop: float | None = None,
+    inplace: bool = True,
+    copy: bool = False,
+) -> AnnData | EHRData | tuple[np.ndarray, np.ndarray] | None:
```
Zethson:
```suggestion
) -> EHRData | None:
```

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+    """Filter observations based on counts and numbers of variables (features/measurements).
```
Zethson: Docstring needs an overhaul because of counts and edata etc

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+            if threshold_min is not None:
+                msg += f"less than {threshold_min} " + ("counts" if by_counts else "features")
+            else:
+                msg += f"more than {threshold_max} " + ("counts" if by_counts else "features")
+            if time_mode == "proportion":
+                msg += f" in < {prop * 100:.1f}% of time points"
+            else:
+                msg += f" in {time_mode} time points"
+
+            print(msg)
```
Zethson:
```suggestion
            logger.info(msg)
```

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+                if threshold_min is not None:
+                    msg += f"less than {threshold_min} counts"
+                else:
+                    msg += f"more than {threshold_max} counts"
+
+                if time_mode == "proportion":
+                    msg += f" in less than {prop * 100:.1f}% of time points"
+                else:
+                    msg += f" in {time_mode} time points"
+                print(msg)
```
Zethson:
```suggestion
                logger.info(msg)
```

tests/preprocessing/test_filter_features.py
```diff
@@ -0,0 +1,54 @@
...
+import pytest
+import scipy.sparse as sp
+
+import ehrapy as ep
+
+CURRENT_DIR = Path(__file__).parent
+
+
+def test_filter_features_basic(mcar_edata_3d):
+    edata = mcar_edata_3d.copy()
```
Zethson: If this is a fixture, we don't need to create a copy.

tests/preprocessing/test_filter_features.py
```diff
@@ -0,0 +1,54 @@
...
+    with pytest.raises(ValueError):
+        ep.pp.filter_features(edata, min_obs=5, time_mode="proportion", prop=2, inplace=True, copy=False)
+
+    n_vars_before = edata.R.shape[1]
+    ep.pp.filter_features(edata, min_obs=5, time_mode="any", inplace=True, copy=False)
+    n_vars_after = edata.R.shape[1]
+    assert n_vars_after < n_vars_before
+
+
+def test_filter_obs_basic(mcar_edata_3d):
```
Zethson: The test name isn't that descriptive. What is `basic`?

tests/conftest.py
```diff
@@ -144,6 +144,19 @@ def mcar_edata(rng) -> ed.EHRData:
     return ed.EHRData(data)
 
 
+@pytest.fixture
+def mcar_edata_3d(rng) -> ed.EHRData:
```
Zethson: Should we use this or can we use the 3d blobs?

ehrapy/preprocessing/_scanpy_pp_api.py
eroell: good as well, yep

ehrapy/preprocessing/_scanpy_pp_api.py
eroell: Yeah, `inplace` was at one point reserved for functions that modify the dimensionality of anndata right.
I think its not sorted out entirely across scanpy. We can drop `inplace`, all the easier even if we don't have a call to scanpy under the hood here.

sueoglu: If we drop inplace, it returns the arrays anyways when we call scanpy under the hood since its inplace=True by default. Do you think I should just keep it for the 2D case and only drop it for the 3D input?

sueoglu: Oh okay I saw your comment below about not calling scanpy anymore, all good.

eroell: OK, the inplace argument can be removed then

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +401,255 @@ def neighbors(
...
+        min_obs: Minimum number of observations required for a feature to pass filtering.
+        max_counts: Maximum number of counts allowed for a feature to pass filtering.
+        max_obs: Maximum number of observations allowed for a feature to pass filtering.
+        time_mode: How to combine filtering criteria across the time axis. Only relevant if an `EHRData` is passed. Options are:
+                    * `'all'` (default): The feature must pass the filtering criteria in all time points.
+                    * `'any'`: The feature must pass the filtering criteria in at least one time point.
+                    * `'proportion'`: The feature must pass the filtering criteria in at least a proportion `prop` of time points. For example, with `prop=0.3`,
+                    the feature must pass the filtering criteria in at least 30% of the time points.
+        prop: Proportion of time points in which the feature must pass the filtering criteria. Only relevant if `time_mode='proportion'`.
+        inplace: Performs the operation inplace or returns a copy.
```
eroell:
```suggestion
```

ehrapy/preprocessing/_scanpy_pp_api.py
eroell: A 3D example here to illustrate the filtering then :+1:

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -8,9 +8,11 @@
 
 import numpy as np
 import scanpy as sc
+import scanpy.logging as logger
```
Zethson: We're not using the scanpy logger. Please check what we're importing in other places.

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
     )
 
 
+def filter_features(
```
eroell: Your nice new functions are not yet visible in the docs; See https://github.com/theislab/ehrapy/blob/main/docs/api/preprocessing_index.md how to add them, then you can check in the readthedocs build if everything looks as it should

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+    edata: EHRData,
```
eroell:
```suggestion
    edata: EHRData | AnnData
```

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+    *,
+    min_obs: int | None = None,
+    max_obs: int | None = None,
+    time_mode: Literal["all", "any", "proportion"] = "all",
+    prop: float | None = None,
+    copy: bool = False,
+) -> EHRData | None:  # pragma: no cover
```
eroell:
```suggestion
) -> EHRData | AnnData | None:  # pragma: no cover
```

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+    """Filter features based on number of observations.
```
eroell: Filter features based on missing data thresholds

Might be more descriptive. But you can object here

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+
+    Keep only features which have at least `min_obs` observations
```
eroell: Keep only features which have at least `min_obs` observations, meaning non-missing values.

Specify once what a non-missing value is, following the comment above

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+    >>> import ehrapy as ep
+    >>> edata = ed.dt.ehrdata_blobs(n_variables=45, n_observations=500, base_timepoints=15, missing_values=0.6)
+    >>> edata.R.shape
+    (500, 45, 15)
+    >>> ep.pp.filter_features(edata, min_obs=185, time_mode="all")
+    >>> edata.R.shape
+    (500, 18, 15)
+
+    """
+    if not isinstance(edata, EHRData):
```
eroell: or AnnData

ehrapy/preprocessing/_scanpy_pp_api.py
eroell: While this is a new function, I'd be still in favor to add the anndata deprecation warning decorator to allow this to handle AnnData for now

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+        raise TypeError("Data object must be an EHRData object")
+
+    data = edata.copy() if copy else edata
+
+    lower_set = min_obs is not None
```
eroell: `lower_set` and `upper_set` are only used once; in that case you can omit defining these variables, and add `min_obs is not None` and `max_obs is not None` directly into the `if` statement

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+    upper_set = max_obs is not None
+
+    if not (lower_set or upper_set):
+        raise ValueError("You must provide at least one of 'min_obs' and 'max_obs'")
+
+    if time_mode not in {"all", "any", "proportion"}:
```
eroell: I'm always a fan to test these exceptions as well, it can be a 2-line test function like we have it for other functions where we test invalid arguments

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+        raise ValueError(f"time_mode must be one of 'all', 'any', 'proportion', got {time_mode}")
+
+    if time_mode == "proportion" and (prop is None or not (0 < prop <= 1)):
+        raise ValueError("prop must be set to a value between 0 and 1 when time_mode is 'proportion'")
+
+    obs_ax, _var_ax, _time_ax = 0, 1, 2
+
+    threshold_min = min_obs
```
eroell: again, variables `threshold_*` I think only used once

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+    threshold_max = max_obs
+
+    missing_mask = np.isin(data.R, MISSING_VALUES) | np.isnan(data.R)
```
eroell: Huh, I make a note to myself that `np.nan` should be added to MISSING_VALUES

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+    if not (lower_set or upper_set):
+        raise ValueError("You must provide at least one of 'min_obs' and 'max_obs'")
+
+    if time_mode not in {"all", "any", "proportion"}:
+        raise ValueError(f"time_mode must be one of 'all', 'any', 'proportion', got {time_mode}")
+
+    if time_mode == "proportion" and (prop is None or not (0 < prop <= 1)):
+        raise ValueError("prop must be set to a value between 0 and 1 when time_mode is 'proportion'")
+
+    obs_ax, _var_ax, _time_ax = 0, 1, 2
```
eroell: you use `obs_ax` and `_var_ax` sometimes, but sometimes just pass the numbers. Sticking to the numbers here and not have dedicated variables would be more consistent in how we do it across the package :)

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+    (500, 45, 15)
+    >>> ep.pp.filter_features(edata, min_obs=185, time_mode="all")
+    >>> edata.R.shape
+    (500, 18, 15)
+
+    """
+    if not isinstance(edata, EHRData):
+        raise TypeError("Data object must be an EHRData object")
+
+    data = edata.copy() if copy else edata
```
eroell: If we also allow for AnnData, could you add one additional check that the passed layer is 2 dimensional?

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+        edata: Central data object.
+        min_obs: Minimum number of observations required for a feature to pass filtering.
+        max_obs: Maximum number of observations allowed for a feature to pass filtering.
+        time_mode: How to combine filtering criteria across the time axis. Options are:
+                    * `'all'` (default): The feature must pass the filtering criteria in all time points.
+                    * `'any'`: The feature must pass the filtering criteria in at least one time point.
+                    * `'proportion'`: The feature must pass the filtering criteria in at least a proportion `prop` of time points. For example, with `prop=0.3`,
+                    the feature must pass the filtering criteria in at least 30% of the time points.
+        prop: Proportion of time points in which the feature must pass the filtering criteria. Only relevant if `time_mode='proportion'`.
+        copy: Determines whether a copy is returned.
```
eroell: The `layer` argument is missing. I think it's suitable to have, right? I might be overlooking something though.

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+        raise ValueError("prop must be set to a value between 0 and 1 when time_mode is 'proportion'")
+
+    threshold_min = min_vars
+    threshold_max = max_vars
+
+    _obs_ax, var_ax, _time_ax = 0, 1, 2
+    n_obs, n_vars, n_time = edata.R.shape
+    per_time_vals = np.empty((n_obs, n_time), dtype=float)
+
+    for t in range(n_time):
```
eroell: the `for` loop can be vectorized, I hope the below suggestion is correct and please doublecheck :D

```py
missing_mask = np.isin(data.R, MISSING_VALUES) | np.isnan(data.R)
per_time_vals = (~missing_mask).sum(axis=1)
```

tests/preprocessing/test_filter_features.py
```diff
@@ -0,0 +1,61 @@
...
+    with pytest.raises(ValueError):
+        ep.pp.filter_features(edata)
+    # invalid time_mode
+    with pytest.raises(ValueError):
+        ep.pp.filter_features(edata, min_obs=185, time_mode="invalid_mode", copy=False)
+    # invalid prop
+    with pytest.raises(ValueError):
+        ep.pp.filter_features(edata, min_obs=185, time_mode="proportion", prop=2, copy=False)
+
+    # min_obs filtering
```
eroell: this can be split into a separate test - its not doing what the test name says it is doing

tests/preprocessing/test_filter_features.py
```diff
@@ -0,0 +1,61 @@
...
+    with pytest.raises(ValueError):
+        ep.pp.filter_observations(edata)
+    # invalid time_mode
+    with pytest.raises(ValueError):
+        ep.pp.filter_observations(edata, min_vars=10, time_mode="invalid_mode", copy=False)
+    # invalid prop
+    with pytest.raises(ValueError):
+        ep.pp.filter_observations(edata, min_vars=10, time_mode="proportion", prop=2, copy=False)
+
+    # min_vars filtering
```
eroell: same here

tests/preprocessing/test_filter_features.py
```diff
@@ -0,0 +1,61 @@
...
+import numpy as np
+import pytest
+
+import ehrapy as ep
+
+CURRENT_DIR = Path(__file__).parent
+
+
+def test_filter_features_invalid_args_min_max_obs():
+    edata = ed.dt.ehrdata_blobs(n_variables=45, n_observations=500, base_timepoints=15, missing_values=0.6)
```
eroell: When layers are allowed, check that this works for layers. You can check in other test cases how this can be done with `pytest.mark.parametrize`

Also, it make the test more robust if we check that this works in the 2D as well as in the 3D

ehrapy/preprocessing/_scanpy_pp_api.py
eroell: we try to use single-dispatch more consistently across our functions, see the normalization methods. This is helpful to enforce that our functions work regardless of whether a numpy array, a sparse array, or a dask array is passed.

In this case, its enough to make it work for the numpy arrays as you do already - but adding a single-dispatching that raises NotImplementedErrors, should the passed array in `edata` be a scipy.sparse or a dask array, will help users to immediately notice that they'll have to convert their data to a numpy array if they want to use this function

ehrapy/preprocessing/_scanpy_pp_api.py
```diff
@@ -401,6 +403,225 @@ def neighbors(
...
+def filter_observations(
+    edata: EHRData,
+    *,
+    min_vars: int | None = None,
+    max_vars: int | None = None,
+    time_mode: Literal["all", "any", "proportion"] = "all",
+    prop: float | None = None,
+    copy: bool = False,
+) -> EHRData | None:
+    """Filter observations based on numbers of variables (features/measurements).
```
eroell: Many comments from above also apply to this function :)

eroell (changes requested):

ehrapy/preprocessing/_scanpy_pp_api.py
sueoglu: oh sorry it is already included, I just forgot it from my previous implementation

ehrapy/preprocessing/_scanpy_pp_api.py
Zethson: I wonder whether we should put these into a `constants.py` though. I think it's useful to have these be names as they're semantically more readable than magic numbers.

ehrapy/preprocessing/_filter.py
```diff
@@ -0,0 +1,297 @@
...
+            * `'all'` (default): The feature must pass the filtering criteria in all time points.
+            * `'any'`: The feature must pass the filtering criteria in at least one time point.
+            * `'proportion'`: The feature must pass the filtering criteria in at least a proportion `prop` of time points.
+                For example, with `prop=0.3`, the feature must pass the filtering criteria in at least 30% of the time points.
+
+        prop: Proportion of time points in which the feature must pass the filtering criteria. Only relevant if `time_mode='proportion'`.
+        copy: Determines whether a copy is returned.
+
+    Returns:
+        Depending on `copy`, subsets and annotates the passed data object and returns `None`
```
Zethson: Reads a bit weird. Returning `None` isn't particularly interesting. I'd formulate it more in the direction of "Returns a filtered copy and annotated of the passed object or acts in-place". Feel free to change the wording but I think this is a bit clearer.

ehrapy/preprocessing/_filter.py
```diff
@@ -0,0 +1,297 @@
...
+        if layer not in data.layers:
+            raise ValueError(f"Invalid layer provided. Available layers are: {list(data.layers.keys())}")
+        arrs = [data.layers[layer]]
+    else:
+        raise ValueError("layer must be a string or None")
+
+    return arrs
+
+
+def _compute_mask(arr: np.ndarray, *, min_count: int, max_count: int, time_mode: str, prop: float, axis: int, caller):
```
Zethson: Return type, please.

tests/preprocessing/test_filter.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@pytest.fixture
+def ehr_3d_blobs(scope="module"):
+    return ed.dt.ehrdata_blobs(n_variables=45, n_observations=500, base_timepoints=15, missing_values=0.6)
+
+
+def test_filter_features_invalid_args(ehr_3d_blobs):
+    # no threshold
+    with pytest.raises(ValueError):
```
Zethson: Using the "match" parameter you can test quickly for the raised error message. Please do that for all errors.

tests/preprocessing/test_filter.py
```diff
@@ -0,0 +1,118 @@
...
+    "fixture, shape, kwargs",
+    [
+        ("ehr_3d_blobs", "R", {"min_obs": 185}),
+        ("ehr_3d_blobs", "R", {"max_obs": 200}),
+        ("mcar_edata", "X", {"min_obs": 90}),
+        ("mcar_edata", "X", {"max_obs": 60}),
+    ],
+)
+def test_filter_features_min_max(request, fixture, shape, kwargs):
+    # generic test for min_obs and max_obs filtering on 2d and 3d data
```
Zethson: This could be a docstring and not a comment right?

Zethson (approved):
Thank you! 

I'm only picky here so that we establish a high standard for all future contributions. I'll try to annoy a bit less in the future.

1. Please update the PR description with an example of how it works and what it does. We're linking PRs in the changelog and people should understand from the PR description what new cool stuff we have now.
2. You added `Filtering` to the very bottom of the docs. Should we have it at the bottom or the top? I'm not sure but it's something to also think about.

A few more minor comments but after these I would be happy to merge it.

Issue #482, opened 2023-03-14
Add dynamic time warping comparison


Zethson:
### Description of feature

https://github.com/DynamicTimeWarping/dtw-python

PR #954, opened 2025-09-24
add time series distances

Zethson:
Fixes https://github.com/theislab/ehrapy/issues/482

- Adds support for Dynamic Time Warping (DTW) and related timeseries metrics (`dtw`, `soft_dtw`, `gak`) in `ep.pp.neighbors()` for computing neighborhood graphs on longitudinal EHR data.
- Added `timeseries_distance()` function that computes temporal distances between patients across multiple variables
- Extended `ep.pp.neighbors()` to accept time series metrics: `"dtw"`, `"soft_dtw"`, `"gak"`
- Added `_MetricTimeSeries` type for time series-specific metrics
- Moves `neighbors` related code into its own module

## Usage
```python
import ehrdata as ed
import ehrapy as ep

edata = ed.dt.ehrdata_blobs(...)
ep.pp.neighbors(edata, metric="dtw")  # or "soft_dtw", "gak"
```

## Gotchas

- Requires edata.R: All timeseries metrics require the 3D tensor edata.R (patients × variables × timepoints). Raises ValueError if not set.

- `soft_dwt` and `gak` are not necessarily true distance metrics:

soft_dtw: Normalized by subtracting average self-distance to ensure identical sequences yield 0 and preserve symmetry
gak: Returns similarity [0,1], converted to distance via 1 - similarity

- Computational cost: These metrics are significantly slower than Euclidean distance. Not sparse-capable. Use on reasonably-sized datasets (<1000 patients) or with approximate kNN transformers.
Minimum overlap requirement: Requires >3 shared valid timepoints per variable pair, otherwise that variable is excluded from distance calculation.
- requires new `tslearn` dependency. Should we make `tslearn` a required dependency? They pin lots of dependencies but are pretty active. We can make it optional and then it'll error like leiden. See https://github.com/tslearn-team/tslearn/issues/600
- Currently only supports numpy arrays. We'll add sparse support when sparse COO is ready
- We currently don't have a tutorial because I want to overhaul it all at once

Test failures are unrelated. We require a new `EHRData` release.


ehrapy/tools/distances/timeseries.py
```diff
@@ -0,0 +1,67 @@
...
+    obs_i, obs_j = int(obs_indices_x[0]), int(obs_indices_y[0])
+    total_distance = 0
+    valid_variable_count = 0
+
+    for variable_idx in range(R.shape[1]):
+        series_i = R[obs_i, variable_idx, :]
+        series_j = R[obs_j, variable_idx, :]
+        valid_measurements_i = ~np.isnan(series_i)
+        valid_measurements_j = ~np.isnan(series_j)
+        if np.sum(valid_measurements_i) > 3 and np.sum(valid_measurements_j) > 3:
```
sueoglu, line 87: Does this also apply when two patients dont have identical sampling times? Because the docstring specifices >3 valid timepoints per patient but how do we handle misaligned timeseries data?

Zethson: Right, I should add a test for this.
Edit: Actually we already have a test for this.

ehrapy/tools/distances/timeseries.py
```diff
@@ -0,0 +1,67 @@
...
+
+            - "dtw": Standard DTW with exact optimal alignment via dynamic programming
+            - "soft_dtw": Differentiable approximation using soft-min, robust to noise/outliers
+            - "gak": Global Alignment Kernel, similarity-based metric robust to irregular sampling
+
+    Returns:
+        Average temporal distance across valid variable pairs.
+        Returns 0 if no valid variable pairs exist.
+    """
+    match metric:
```
sueoglu, line 60: it would be great to give an error message when the user doesn't have` tslearn` installed so that not the raw error is displayed but an explicit error message for users

Zethson: It's a required dependency at the moment so it will always be available.

tests/conftest.py
```diff
@@ -282,11 +282,26 @@ def edata_small_bias() -> ed.EHRData:
...
+    edata = ed.dt.ehrdata_blobs(
+        n_observations=20,
+        base_timepoints=15,
+        cluster_std=0.5,
+        n_centers=3,
+        seasonality=True,
+        time_shifts=True,
+        variable_length=False,
+    )
+    edata.layers["layer_2"] = edata.X.copy()
```
agerardy, line 301: Does adding a second layer equal adding an edata.R object?

ehrapy/tools/distances/timeseries.py
```diff
@@ -0,0 +1,67 @@
+from typing import Literal
+
+import numpy as np
+
+
+def timeseries_distance(
+    obs_indices_x: np.ndarray,
+    obs_indices_y: np.ndarray,
+    R: np.ndarray,
```
agerardy: What happens if R is not a numpy array, is there an error message?

tests/conftest.py
Zethson: Layers can be 2D or 3D. That's why eventually we want to move to a design where everything is a layer. So yeah we might consider just copying R here instead.

tests/preprocessing/test_neighbors.py
```diff
@@ -0,0 +1,29 @@
...
+    ep.pp.neighbors(edata, n_neighbors=5, metric=metric)
+
+    assert "neighbors" in edata.uns
+    assert "distances" in edata.obsp
+    assert "connectivities" in edata.obsp
+    assert edata.obsp["distances"].shape == (20, 20)
+    assert edata.obsp["connectivities"].shape == (20, 20)
+
+
+# TODO: neighbors does not have layer support. Once X can be 3D (https://github.com/scverse/anndata/pull/1707), this function could however encounter a 3D object in X; then test this
```
agerardy, line 25: Does it not have layers support now?

agerardy (commented):
Not sure what happens when an ehrdata object with a dask array in .R is passed, other than that it looks good to me

tests/preprocessing/test_neighbors.py
Zethson: I don't remember having added this comment. I think it was @eroell . But I think it's mostly about the interface here.

ehrapy/tools/distances/timeseries.py
Zethson: It's single dispatched now so hopefully yes :smile:

Issue #944, opened 2025-09-22
Longitudinal normalization


eroell:
#### This is a function that operates on `.X` or `.layers` so 3D and array-type and encoding is relevant.

This issue addresses all the current normalization methods, since they are very streamlined:

- [ ] minmax_norm
- [ ] maxabs_norm
- [ ] robust_scale_norm
- [ ] quantile_norm
- [ ] power_norm
- [ ] log_norm
- [ ] offset_negative_values

**3D**
If 3D allowed:
- [ ] function handles 3D data
- [ ] the test of the function tests both the expected 2D and 3D implementation

If 3D not allowed:
- [ ] only_2D decorator
- [ ] the test of the function tests for failure if 3D is passe

**array-type**
The available array types are np.array, dask.array, and scipy.sparse matrices.
- [x] function is single-dispatched, with potential not_implemented errors being raised
- [x] test is parametrized to test the different array-types

**Encoding**
If encoded data is required
- [ ] mention in documentation that encoded data is needed, with ep.pp.encode.

eroell:
```py
from functools import singledispatch

import ehrdata as ed
import numpy as np
import xarray as xr

from ehrapy._compat import _raise_array_type_not_implemented


class StandardScaler3D:
    def __init__(self):
        """Standardize features by removing the mean and scaling to unit variance, across all samples and timesteps.
        This class is in concept similar to :class:`sklearn.preprocessing.StandardScaler`, but for a 3D array of shape (n_samples, n_features, n_timesteps).
        """
        self.mean_ = None
        self.std_ = None

    def _fit(self, data: np.ndarray) -> None:
        # Compute mean and std along the (0, 2) axes (samples and timesteps)
        self.mean_ = np.nanmean(data, axis=(0, 2), keepdims=True)
        self.scale_ = np.nanstd(data, axis=(0, 2), keepdims=True)

        self.scale_[self.scale_ == 0] = 1.0

    def fit(self, edata: ed.EHRData) -> None:
        """Fit the StandardScaler3D object to the input data.
        Computes the mean and standard deviation for each feature across all samples and timesteps from the input EHRData's `.r` field.
        If missing values are present, they are ignored during this computation.
        Args:
            edata: Input EHRData, from which's `.r` field, the statistics for the normalization are computed on.
        """

        if not isinstance(edata, ed.EHRData):
            raise ValueError("Input must be an EHRData object.")

        if edata.r.ndim != 3:
[...]

PR #958, opened 2025-10-07
#944 longitudinal normalization 

agerardy:
**PR Checklist**

**Description of changes**
https://github.com/theislab/ehrapy/issues/944
This PR implements normalization support for 3D EHRData objects. The implementation enables all existing normalization functions to work with longitudinal data with shape `(n_obs, n_var, n_timestamps)` but maintains backward compatibility with 2D data. 

**Technical details**
Treats .R as a named layer with 3D structure. Uses helper functions (`_get_target_layer`, `_set_target_layer`, and `normalize_3d_data`, `_normalize_2d_data`) to avoid code duplication. 
Each variable is processed independently by flattening the time dimension `(n_obs x n_timestamps)`, applying the sklearn normalization function, then reshaping to 3D.

Added tests for the new functions, including group functionality and NaN cases

**Examples:**
``` python
edata = ed.dt.ehrdata_blobs(n_observations=100, base_timepoints=24, cluster_std=0.5, n_centers=3, seasonality=True, 
    time_shifts=True, variable_length=False)

# standard scaling
ep.pp.scale_norm(edata)

# log transformation
ep.pp.offset_negative_values(edata)
ep.pp.log_norm(edata)
```

### DIFF
--- a/tests/preprocessing/test_normalization.py
+++ b/tests/preprocessing/test_normalization.py
@@ -544,7 +544,7 @@ def test_norm_power_integers(edata_mini_integers_in_X):
             [-0.31234142],
         ]
     )
-    assert np.allclose(adata_norm.X, in_days_norm)
+    assert np.allclose(adata_norm.X, in_days_norm, rtol=1e-4, atol=1e-4)
 
 
 @pytest.mark.parametrize("array_type", ARRAY_TYPES_NONNUMERIC)


ehrapy/preprocessing/_normalization.py
```diff
@@ -43,25 +48,58 @@ def _scale_func_group(
...
 
-    if layer is None:
-        var_values = edata[:, vars].X.copy()
-    else:
-        var_values = edata[:, vars].layers[layer].copy()
-
-    if group_key is None:
-        var_values = scale_func(var_values)
-
+    if hasattr(edata, "R") and edata.R is not None and edata.R.ndim == 3:
```
eroell: If the edata object has `R`, then it will be used regardless of what the layer argument specified.

This line is a good start to investigate and set a stopping point with a debugger, to see what branch of the if/else statement is actually entered by the code, and whether it matches what you think it should :)

Zethson: @agerardy this hasn't been addressed right? Eljas first part of the comment.

ehrapy/preprocessing/_normalization.py
```diff
@@ -43,25 +48,61 @@ def _scale_func_group(
...
+        # Convert variable names to indices
+        from ehrapy.anndata.anndata_ext import _get_var_indices
 
-    if group_key is None:
-        var_values = scale_func(var_values)
+        var_indices = _get_var_indices(edata, vars)
 
+        if layer is None:
+            var_values = edata.R[:, var_indices, :].copy()
```
Zethson: Can we get away without making copies?

ehrapy/preprocessing/_normalization.py
```diff
@@ -43,25 +48,61 @@ def _scale_func_group(
...
+        else:
+            var_values = edata.layers[layer][:, var_indices, :].copy()
```
Zethson: Can we get away without making copies?

ehrapy/preprocessing/_normalization.py
```diff
@@ -113,15 +157,21 @@ def scale_norm(
...
     Returns:
-        `None` if `copy=False` and modifies the passed edata, else returns an updated AnnData object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated edata object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
         >>> edata = ed.dt.mimic_2()
         >>> edata_norm = ep.pp.scale_norm(edata, copy=True)
+        >>> # Works automatically with both 2D and 3D data
```
Zethson:
```suggestion
```

ehrapy/preprocessing/_normalization.py
```diff
@@ -113,15 +157,21 @@ def scale_norm(
...
+        >>> edata_3d_norm = ep.pp.scale_norm(edata_3d, copy=True)
```
Zethson: Let's keep it simple and not distinguish between 2D and 3D. We should rather finally have a proper 3D test dataset @eroell .

ehrapy/preprocessing/_normalization.py
```diff
@@ -113,15 +157,21 @@ def scale_norm(
...
     """
-    scale_func = _scale_norm_function(edata.X if layer is None else edata.layers[layer], **kwargs)
+    if hasattr(edata, "R") and edata.R is not None and edata.R.ndim == 3:
```
Zethson: See Eljas above.

ehrapy/preprocessing/_normalization.py
```diff
@@ -177,15 +230,21 @@ def minmax_norm(
...
     Returns:
-        `None` if `copy=False` and modifies the passed edata, else returns an updated AnnData object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated data object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
         >>> edata = ed.dt.mimic_2()
         >>> edata_norm = ep.pp.minmax_norm(edata, copy=True)
+        >>> # Works automatically with both 2D and 3D data
```
Zethson: See above.

ehrapy/preprocessing/_normalization.py
```diff
@@ -177,15 +230,21 @@ def minmax_norm(
...
+        >>> edata_3d_norm = ep.pp.minmax_norm(edata_3d, copy=True)
     """
-    scale_func = _minmax_norm_function(edata.X if layer is None else edata.layers[layer], **kwargs)
+    if hasattr(edata, "R") and edata.R is not None and edata.R.ndim == 3:
```
Zethson: See above.

ehrapy/preprocessing/_normalization.py
```diff
@@ -177,15 +230,21 @@ def minmax_norm(
         **kwargs: Additional arguments passed to the MinMaxScaler.
 
     Returns:
-        `None` if `copy=False` and modifies the passed edata, else returns an updated AnnData object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated data object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
```
Zethson:
```suggestion
        `None` if `copy=False` and modifies the passed edata, else returns an updated object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
```

tests/conftest.py
```diff
@@ -287,6 +287,22 @@ def edata_blob_small() -> ed.EHRData:
     return edata
 
 
+@pytest.fixture
+def edata_blob_small_3d() -> ed.EHRData:
```
Zethson: Do we need another fixture here or can you reuse the `edata_blobs_timeseries_small`?

Zethson (changes requested):
Thank you! Already looks pretty good.

1. Many of comments are repetitive so I stopped repeating them after some time 😄 
2. Many of your tests have tons of useless comments. Let the code speak for itself and clean up any LLM leftovers, please.
3. Please also follow the comments that I make in Öyku's PRs. One of them is to improve the PR description and add some usage examples.

Just a first quick pass. I'll let @eroell have a go and then I might have a look again.

Thanks!

tests/conftest.py
agerardy: @Zethson I made one specifically with .R and my tests rely on that, but maybe they shouldn't...

Zethson: `edata_blobs_timeseries_small` also has an R

agerardy: ah, you're right my tests are failing for a different reason, nevermind

eroell: If `edata_blobs_timeseries_small` works, you can a) remove this new fixture and b) click "resolve this comment". Then it disappears and helps us focus on remaining open comments!

tests/preprocessing/test_normalization.py
```diff
@@ -720,3 +720,31 @@ def test_norm_numerical_only():
...
 
     assert np.array_equal(expected_adata.X, ep.pp.log_norm(to_normalize_adata, copy=True).X)
+
+
+def test_scale_norm_3d(edata_blob_small_3d):
+    """Test scale_norm normalization on 3D EHRData (edata.R)."""
+    edata = edata_blob_small_3d
+    # Save original shape and dtype
+    orig_shape = edata.R.shape
+    print("Original shape:", orig_shape)
```
eroell: No print statements in test functions, we test many things and print statements make the test output too large

tests/preprocessing/test_normalization.py
```diff
@@ -720,3 +720,31 @@ def test_norm_numerical_only():
...
+    orig_dtype = edata.R.dtype
+    print("Original dtype:", orig_dtype)
+    print("mean:", np.mean(edata.R), "std:", np.std(edata.R))
+    # Run normalization
```
eroell: No need for such small comments - in most cases the code can explain itself

tests/preprocessing/test_normalization.py
```diff
@@ -720,3 +720,31 @@ def test_norm_numerical_only():
...
+    # print("attr:", getattr(edata, 'R', None))
+    ep.pp.scale_norm(edata)
+    print("mean:", np.mean(edata.R), "std:", np.std(edata.R))
+
+    # After normalization, R should still be 3D and same shape
+    assert edata.R.shape == orig_shape
+    assert edata.R.dtype == orig_dtype or np.issubdtype(edata.R.dtype, np.floating)
+
+    # Check that each variable (axis=1) is zero mean and unit variance across all samples and timestamps
+    n_obs, n_var, n_timestamps = edata.R.shape
```
eroell: n_obs, n_var, n_timestamps never used? think this line can be deleted?

tests/preprocessing/test_normalization.py
```diff
@@ -720,3 +720,31 @@ def test_norm_numerical_only():
...
+    for var_idx in range(n_var):
+        flat = edata.R[:, var_idx, :].reshape(-1)
+        # Allow small numerical error
```
eroell: Just to give some more "calibration"; this is again a comment that can be removed. `atol=1e-6` is quite explicit about what is happening

tests/preprocessing/test_normalization.py
```diff
@@ -720,3 +720,31 @@ def test_norm_numerical_only():
     expected_adata = AnnData(X=np.array([[0.6931472, 0, 0], [0, 0, 0.6931472]], dtype=np.float32))
 
     assert np.array_equal(expected_adata.X, ep.pp.log_norm(to_normalize_adata, copy=True).X)
+
+
+def test_scale_norm_3d(edata_blob_small_3d):
```
eroell: Every changed function should be tested - are you preferring to make one classic example here, which you then expand to every other normalization function once we have iterated, or create them now already? Both options are OK

ehrapy/preprocessing/_normalization.py
```diff
@@ -537,20 +614,31 @@ def _record_norm(edata: EHRData | AnnData, vars: Sequence[str], method: str) ->
...
-        `None` if `copy=False` and modifies the passed edata, else returns an updated data object.
+        `None` if `copy=False` and modifies the passed edata, else returns an updated edata object.
+
+    Examples:
+        >>> import ehrdata as ed
+        >>> import ehrapy as ep
+        >>> edata = ed.dt.mimic_2()
+        >>> edata_offset = ep.pp.offset_negative_values(edata, copy=True)
+        >>> # Works automatically with both 2D and 3D data
+        >>> edata_3d_offset = ep.pp.offset_negative_values(edata_3d, copy=True)
```
eroell: Gathering a few comments from other places here, to make it less dispersed
a) Showing some output is very helpful, see e.g. [here](https://github.com/theislab/ehrapy/blob/7a42d0e1a71b6c858337063ccef75a419aa9a897/ehrapy/preprocessing/_filter.py#L178)
b) One test of 2D or 3D is enough; Can you use a 3D here, ideally the physionet2012 dataset?
c) edata_3d variable would never have been introduced

ehrapy/preprocessing/_normalization.py
```diff
@@ -417,13 +487,15 @@ def power_norm(
...
     Returns:
-        `None` if `copy=False` and modifies the passed edata, else returns an updated data object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated edata object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
         >>> edata = ed.dt.mimic_2()
         >>> edata_norm = ep.pp.power_norm(edata, copy=True)
+        >>> # Works automatically with both 2D and 3D data
```
eroell: Gathering a few comments from other places here, to make it less dispersed
a) Showing some output is very helpful, see e.g. [here](https://github.com/theislab/ehrapy/blob/7a42d0e1a71b6c858337063ccef75a419aa9a897/ehrapy/preprocessing/_filter.py#L178)
b) One test of 2D or 3D is enough; Can you use a 3D here, ideally the physionet2012 dataset?
c) edata_3d variable would never have been introduced

ehrapy/preprocessing/_normalization.py
```diff
@@ -360,13 +425,15 @@ def quantile_norm(
...
     Returns:
-        `None` if `copy=False` and modifies the passed edata, else returns an updated data object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated edata object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
         >>> edata = ed.dt.mimic_2()
         >>> edata_norm = ep.pp.quantile_norm(edata, copy=True)
+        >>> # Works automatically with both 2D and 3D data
```
eroell: Gathering a few comments from other places here, to make it less dispersed
a) Showing some output is very helpful, see e.g. [here](https://github.com/theislab/ehrapy/blob/7a42d0e1a71b6c858337063ccef75a419aa9a897/ehrapy/preprocessing/_filter.py#L178)
b) One test of 2D or 3D is enough; Can you use a 3D here, ideally the physionet2012 dataset?
c) edata_3d variable would never have been introduced

ehrapy/preprocessing/_normalization.py
```diff
@@ -296,13 +356,15 @@ def robust_scale_norm(
...
     Returns:
-        `None` if `copy=False` and modifies the passed edata, else returns an updated AnnData object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated edata object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
         >>> edata = ed.dt.mimic_2()
         >>> edata_norm = ep.pp.robust_scale_norm(edata, copy=True)
+        >>> # Works automatically with both 2D and 3D data
```
eroell: Gathering a few comments from other places here, to make it less dispersed
a) Showing some output is very helpful, see e.g. [here](https://github.com/theislab/ehrapy/blob/7a42d0e1a71b6c858337063ccef75a419aa9a897/ehrapy/preprocessing/_filter.py#L178)
b) One test of 2D or 3D is enough; Can you use a 3D here, ideally the physionet2012 dataset?
c) edata_3d variable would never have been introduced

ehrapy/preprocessing/_normalization.py
```diff
@@ -231,13 +286,15 @@ def maxabs_norm(
...
     Returns:
-        `None` if `copy=False` and modifies the passed edata, else returns an updated AnnData object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated edata object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
         >>> edata = ed.dt.mimic_2()
         >>> edata_norm = ep.pp.maxabs_norm(edata, copy=True)
+        >>> # Works automatically with both 2D and 3D data
```
eroell: Gathering a few comments from other places here, to make it less dispersed
a) Showing some output is very helpful, see e.g. [here](https://github.com/theislab/ehrapy/blob/7a42d0e1a71b6c858337063ccef75a419aa9a897/ehrapy/preprocessing/_filter.py#L178)
b) One test of 2D or 3D is enough; Can you use a 3D here, ideally the physionet2012 dataset?
c) edata_3d variable would never have been introduced

ehrapy/preprocessing/_normalization.py
eroell: Gathering a few comments from other places here, to make it less dispersed
a) Showing some output is very helpful, see e.g. [here](https://github.com/theislab/ehrapy/blob/7a42d0e1a71b6c858337063ccef75a419aa9a897/ehrapy/preprocessing/_filter.py#L178)
b) One test of 2D or 3D is enough; Can you use a 3D here, ideally the physionet2012 dataset?
c) edata_3d variable would never have been introduced

ehrapy/preprocessing/_normalization.py
agerardy: I've moved this logic to its own function _get_target_layer and changed it to check for layers first.
```python
    if layer is None:
        if hasattr(edata, "R") and edata.R is not None:
            return edata.R, "R"
        else:
            return edata.X, "X"
    else:
        return edata.layers[layer], layer
```
I hope this is how it needs to work

tests/preprocessing/test_normalization.py
agerardy: I can't tell anymore which function this referred to exactly, but all instances of `n_obs, n_var, n_timestamps` i can find use range(n_var) directly after

ehrapy/preprocessing/_normalization.py
agerardy: I'll just reply to this one and mark the duplicates resolved :) I've written the examples with fake numbers for now, but will try to get it actually running with the physionet2012 dataset. it certainly looks better but doesnt work yet

tests/preprocessing/test_normalization.py
agerardy: I'm not sure what you mean by this. I have written a 3D test function for every normalization function?

ehrapy/preprocessing/_normalization.py
```diff
@@ -177,15 +215,21 @@ def minmax_norm(
...
-        `None` if `copy=False` and modifies the passed edata, else returns an updated AnnData object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
-        >>> edata = ed.dt.mimic_2()
-        >>> edata_norm = ep.pp.minmax_norm(edata, copy=True)
+        >>> import numpy as np
+        >>> edata = ed.dt.physionet2012()
```
eroell: Please use the longitudinal layers array instead of the 2D .X field for the examples

tests/preprocessing/test_normalization.py
```diff
@@ -101,31 +95,111 @@ def test_norm_scale_group(array_type, edata_mini_normalization):
     edata_mini_casted = edata_mini_normalization.copy()
     edata_mini_casted.X = array_type(edata_mini_casted.X)
 
-    with pytest.raises(KeyError):
-        ep.pp.scale_norm(edata_mini_casted, group_key="invalid_key", copy=True)
+    if "dask" in array_type.__name__:
```
eroell: See #975 for how to check for the type with `isinstance`

tests/preprocessing/test_normalization.py
```diff
@@ -101,31 +95,111 @@ def test_norm_scale_group(array_type, edata_mini_normalization):
...
+            ]
+        )
+        col2_norm = col1_norm
+        assert np.allclose(edata_mini_norm.X[:, 0], edata_mini_casted.X[:, 0])
+        assert np.allclose(edata_mini_norm.X[:, 1], col1_norm)
+        assert np.allclose(edata_mini_norm.X[:, 2], col2_norm)
+
+
+@pytest.mark.parametrize("copy", [True, False])
+def test_3d_norm_copy_behavior(edata_blobs_timeseries_small, copy):
```
eroell: please use the naming convention

`test_<ehrapy function>_3D_...`

And only if its more than 1 ehrapy function tested `test_norm_....`

tests/preprocessing/test_normalization.py
```diff
@@ -101,31 +95,111 @@ def test_norm_scale_group(array_type, edata_mini_normalization):
...
+    [
+        ep.pp.scale_norm,
+        ep.pp.minmax_norm,
+        ep.pp.maxabs_norm,
+        ep.pp.robust_scale_norm,
+        ep.pp.quantile_norm,
+        ep.pp.power_norm,
+    ],
+)
+def test_3d_norm_shape_preservation(edata_blobs_timeseries_small, norm_func):
```
eroell: this is on the right track:
I suggest to create a test function 

`test_norm_3D` where very basic things such as this one is evaluated

tests/preprocessing/test_normalization.py
```diff
@@ -627,14 +683,6 @@ def test_norm_log_norm_array_types(adata_to_norm, array_type, expected_error):
             ep.pp.log_norm(adata_to_norm)
 
 
-def test_norm_log_3D_edata(edata_blob_small):
```
eroell, line 630: I'm aware this was not sticking to the naming convention we're trying to stick to now

tests/preprocessing/test_normalization.py
```diff
@@ -708,15 +756,274 @@ def test_offset_negative_values():
...
-
 def test_norm_numerical_only():
     """Test for the log_norm method."""
     to_normalize_adata = AnnData(X=np.array([[1, 0, 0], [0, 0, 1]], dtype=np.float32))
     expected_adata = AnnData(X=np.array([[0.6931472, 0, 0], [0, 0, 0.6931472]], dtype=np.float32))
 
     assert np.array_equal(expected_adata.X, ep.pp.log_norm(to_normalize_adata, copy=True).X)
+
+
+def test_scale_norm_3d(edata_blobs_timeseries_small):
```
eroell: The below tests could be packed into one
`test_norm_3D` function, which is parametrized by the different function names? This would reduce a lot of repetition, and can be merged with the function I mentioned above to be a basic test.
Also, this function could be parametrized with `@pytest.mark.parametrize("array_type", ARRAY_TYPES_NUMERIC_3D_ABLE)` as the basic function in #975 is.

Then, only the precise tests would be left over here

tests/preprocessing/test_normalization.py
```diff
@@ -708,15 +756,274 @@ def test_offset_negative_values():
...
+
+    if norm_func == ep.pp.power_norm:
+        edata.layers[DEFAULT_TEM_LAYER_NAME] = np.abs(edata.layers[DEFAULT_TEM_LAYER_NAME]) + 0.1
+
+    with pytest.raises(ValueError):
+        norm_func(edata, layer=DEFAULT_TEM_LAYER_NAME, vars=["nonexistent_var"])
+
+
+def test_3d_norm_variable_selection(edata_blobs_timeseries_small):
+    """Test variable selection with 3D normalization."""
```
eroell: Also, no need to comment tests with docstrings

eroell (changes requested):
Dropped a first intermediate review already, to be considered together with @sueoglu's :)

tests/preprocessing/test_normalization.py
Zethson: They can be helpful but then need to be formulated in a different way like: "XYZ must exhibit ABC behavior". It's obvious then that this is what the test tests for

agerardy: I think we discussed this before, I'll delete most of the docstrings under tests. maybe keep some for the more complicated ones

ehrapy/preprocessing/_normalization.py
```diff
@@ -387,12 +466,12 @@ def _power_norm_function(arr, **kwargs):
 
 
 @_power_norm_function.register
+@_apply_over_time_axis
 def _(arr: np.ndarray, **kwargs):
-    return sklearn_pp.PowerTransformer(**kwargs).fit_transform
+    return sklearn_pp.PowerTransformer(**kwargs).fit_transform(arr)
```
sueoglu, line 463: is it intentional that this and _maxabs_norm_function do not support Dask arrays yet? If it is, then maybe a short note that those two onl expect non Dask arrays

ehrapy/preprocessing/_normalization.py
```diff
@@ -7,9 +7,16 @@
 import sklearn.preprocessing as sklearn_pp
 from ehrdata.core.constants import NUMERIC_TAG
 
-from ehrapy._compat import DaskArray, _raise_array_type_not_implemented, function_2D_only, use_ehrdata
+from ehrapy._compat import (
+    DaskArray,
+    _apply_over_time_axis,
+    _raise_array_type_not_implemented,
+    function_2D_only,
```
sueoglu: I think we dont need this decorator anymore right? because all normalization functions now support 3D

ehrapy/preprocessing/_normalization.py
```diff
@@ -537,31 +660,48 @@ def _record_norm(edata: EHRData | AnnData, vars: Sequence[str], method: str) ->
...
-        if minimum < 0:
-            edata.layers[layer] = edata.layers[layer] + np.abs(minimum)
-    else:
-        minimum = np.min(edata.X)
-        if minimum < 0:
-            edata.X = edata.X + np.abs(minimum)
+    X = edata.X if layer is None else edata.layers[layer]
+    minimum = np.nanmin(X)
+    if minimum < 0:
+        offset_arr = X + np.abs(minimum)
```
sueoglu: just to understand how it works: is a single minimum computed across all variables and all time steps and is the same offset applied to every element for 3D data?

ehrapy/preprocessing/_normalization.py
agerardy: Those functions are not available in dask unfortunately. I added a note to the docs and raise a NotImplementedError if it is still attempted

ehrapy/preprocessing/_normalization.py
agerardy: yes, it takes a global minimum across all axis and applies the same offset everywhere

agerardy: yes, that's exactly how it works

tests/preprocessing/test_normalization.py
agerardy: I packed the basic tests into one function and the precise tests (as in, those that need to check outputs are certain values) into another. the basic test function turned out kinda long but it works fine

ehrapy/preprocessing/_normalization.py
```diff
@@ -113,15 +137,22 @@ def scale_norm(
...
+        `None` if `copy=False` and modifies the passed edata, else returns an updated object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
-        >>> edata = ed.dt.mimic_2()
-        >>> edata_norm = ep.pp.scale_norm(edata, copy=True)
+        >>> import numpy as np
+        >>> edata = ed.dt.ehrdata_blobs(n_observations=20, base_timepoints=15)
+        >>> np.nanmean(edata.layers[DEFAULT_TEM_LAYER_NAME])
```
eroell: Lets always make self-contained examples: this should run through if you copy the code into an environment with ehrdata and ehrapy, but `DEFAULT_TEM_LAYER_NAME` will raise an error since its not defined

ehrapy/preprocessing/_normalization.py
```diff
@@ -417,15 +518,37 @@ def power_norm(
...
+        >>> edata.layers[DEFAULT_TEM_LAYER_NAME] = skewed_data
+        >>> stats.skew(edata.layers[DEFAULT_TEM_LAYER_NAME].flatten())
+        13.528100
+        >>> ep.pp.power_norm(edata, layer=DEFAULT_TEM_LAYER_NAME)
+        >>> stats.skew(edata.layers[DEFAULT_TEM_LAYER_NAME].flatten())
+        -0.041263
     """
-    scale_func = _power_norm_function(edata.X if layer is None else edata.layers[layer], **kwargs)
+    X = edata.X if layer is None else edata.layers[layer]
+    if isinstance(X, DaskArray) and group_key is None:
```
eroell: Let's raise [this central error message](https://github.com/theislab/ehrapy/blob/fae75a147b20768dd084e03a407c469e1d95ba27/ehrapy/_compat.py#L32) that is used in other places as well

eroell: this also gets us rid of the complex case distinction

ehrapy/preprocessing/_normalization.py
```diff
@@ -438,8 +561,44 @@ def power_norm(
...
 
+@singledispatch
+def _log_norm_function(arr, offset: int | float = 1, base: int | float | None = None):
+    _raise_array_type_not_implemented(_log_norm_function, type(arr))
+
+
+@_log_norm_function.register
+@_apply_over_time_axis
+def _(arr: np.ndarray, offset: int | float = 1, base: int | float | None = None) -> np.ndarray:
+    result = arr.copy()  # Work on a copy to avoid modifying input
```
eroell: why only here work on copy?

there is an argument `copy` to the function, can this be used to guide when to work on a copy and when to not do so?

ehrapy/preprocessing/_normalization.py
agerardy: makes sense, i added the import for it

ehrapy/preprocessing/_normalization.py
agerardy: of  course, should have thought of that, i just start doing things way too complicated after I work on them for too long

ehrapy/preprocessing/_normalization.py
agerardy: that was a leftover from an old example i think. fixed it and another instance of using copy in offset_negative_values, thanks

ehrapy/preprocessing/_normalization.py
Zethson: I wouldn't import it but just hardcode the value here. Simplifies the example.

Zethson: I wouldn't import it but just hardcode the value here. Simplifies the example.

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +803,151 @@ def test_norm_record(adata_to_norm):
...
+        (
+            ep.pp.power_norm,
+            lambda flat: (
+                np.allclose(np.nanmean(flat), 0, atol=1e-5),
+                np.allclose(np.nanstd(flat), 1, atol=1e-5),
+            ),
+        ),
+    ],
+)
+def test_norm_3D_precise(edata_blobs_timeseries_small, array_type, norm_func, check_func):
```
eroell: As quickly discussed: split into different tests to avoid very complicated fixture and case distinction

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +803,151 @@ def test_norm_record(adata_to_norm):
...
+    """Test precise mathematical correctness of 3D normalization functions.
```
eroell: no need for docstring

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +803,151 @@ def test_norm_record(adata_to_norm):
...
+    edata = edata_blobs_timeseries_small.copy()
+    edata.layers[DEFAULT_TEM_LAYER_NAME] = array_type(edata.layers[DEFAULT_TEM_LAYER_NAME])
+    if isinstance(edata.layers[DEFAULT_TEM_LAYER_NAME], da.Array) and norm_func in (
+        ep.pp.maxabs_norm,
+        ep.pp.power_norm,
+    ):
+        with pytest.raises(NotImplementedError, match="does not support array type.*dask"):
+            norm_func(edata, layer=DEFAULT_TEM_LAYER_NAME)
+        return
+    orig_shape = edata.layers[DEFAULT_TEM_LAYER_NAME].shape
```
eroell, line 739: no need to check shape again: just precise test, all other things tested earlier

agerardy: this doesnt check the shape of the data itself but checks whether the shape is the same after for example logarithmic normalization was done

eroell: But you do a shape test already in the first, basic test function. This test here is called precise test, and should do as much as possible only this.
And it can do mostly only this, since you test already in `test_norm_3d` for many of these things

ehrapy/preprocessing/_normalization.py
```diff
@@ -43,53 +52,63 @@ def _scale_func_group(
...
-    if layer is None:
-        var_values = edata[:, vars].X.copy()
-    else:
-        var_values = edata[:, vars].layers[layer].copy()
+    var_indices = _get_var_indices(edata, vars)
+    X = edata.X if layer is None else edata.layers[layer]
 
-    if group_key is None:
-        var_values = scale_func(var_values)
+    # group wise normalization doesnt work with dask arrays
```
eroell: Why does group-wise normalization not work with dask arrays?
I see no part in the group computations that cannot be computed lazily. 
Next to unlocking a feature here, The whole testing logic would become much easier once this is enabled, too.

Can you please check or show here with a small example what is not working, after fixing the currently broken group-wise normalization?

ehrapy/preprocessing/_normalization.py
```diff
@@ -360,15 +437,22 @@ def quantile_norm(
...
+        `None` if `copy=False` and modifies the passed edata, else returns an updated object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
-        >>> edata = ed.dt.mimic_2()
-        >>> edata_norm = ep.pp.quantile_norm(edata, copy=True)
+        >>> import numpy as np
+        >>> edata = ed.dt.physionet2012(layer="tem_data")
+        >>> edata.layers["tem_data"] = edata.layers["tem_data"] * 8 + 2
```
eroell: Why the *8 + 2?

ehrapy/preprocessing/_normalization.py
```diff
@@ -537,31 +697,43 @@ def _record_norm(edata: EHRData | AnnData, vars: Sequence[str], method: str) ->
...
     Returns:
-        `None` if `copy=False` and modifies the passed edata, else returns an updated data object.
+        `None` if `copy=False` and modifies the passed edata, else returns an updated object.
+
+    Examples:
+        >>> import ehrdata as ed
+        >>> import ehrapy as ep
+        >>> import numpy as np
+        >>> edata = ed.dt.physionet2012(layer="tem_data")
+        >>> edata_shifted = edata.copy()
```
eroell: I think the example doesn't need edata.copy()?

ehrapy/preprocessing/_normalization.py
```diff
@@ -537,31 +697,43 @@ def _record_norm(edata: EHRData | AnnData, vars: Sequence[str], method: str) ->
...
+        >>> edata_shifted.layers["tem_data"] = edata_shifted.layers["tem_data"] - 0.5
```
eroell: why -0.5? Looks below as if already -17.8 has been the minimal original value. The less extra fluff in examples, the better! :)

tests/preprocessing/test_normalization.py
```diff
@@ -8,14 +8,13 @@
 from ehrdata.core.constants import DEFAULT_TEM_LAYER_NAME
 
 import ehrapy as ep
-from tests.conftest import ARRAY_TYPES_NONNUMERIC
+from tests.conftest import ARRAY_TYPES_NONNUMERIC, ARRAY_TYPES_NUMERIC_3D_ABLE
```
eroell, line 11: `# TODO: check this for each function, with just default settings?`

can be deleted below

ehrapy/preprocessing/_normalization.py
agerardy: i think thats a leftover from the old example, will remove, thx

ehrapy/preprocessing/_normalization.py
agerardy: good question. i had cursor edit these all at onceand missed this, can't figure out how it came up with that. I'll check over all the examples again

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +664,246 @@ def test_norm_record(adata_to_norm):
...
+    [
+        ep.pp.scale_norm,
+        ep.pp.minmax_norm,
+        ep.pp.maxabs_norm,
+        ep.pp.robust_scale_norm,
+        ep.pp.quantile_norm,
+        ep.pp.power_norm,
+    ],
+)
+def test_norm_3D(edata_blobs_timeseries_small, array_type, norm_func):
```
eroell, line 727: This function is very long and complicated. Please make this slimmer. the codecoverage report can guide which lines of code are not tested, and I think here a lot of nested redundancy is present which does not enhance coverage and is so complex that its difficult to maintain the overview: I'll add a few comments that highlight this

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +664,246 @@ def test_norm_record(adata_to_norm):
...
+        norm_func(edata_select, layer=DEFAULT_TEM_LAYER_NAME, vars=selected_vars)
+
+        assert not np.allclose(
+            edata_select.layers[DEFAULT_TEM_LAYER_NAME][:, 0, :], layer_before_select[:, 0, :], equal_nan=True
+        )
+        assert not np.allclose(
+            edata_select.layers[DEFAULT_TEM_LAYER_NAME][:, 1, :], layer_before_select[:, 1, :], equal_nan=True
+        )
+
+        if edata_select.layers[DEFAULT_TEM_LAYER_NAME].shape[1] > 2:
```
eroell: The exact same condition is checked already before entering this branch on line 739

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +664,246 @@ def test_norm_record(adata_to_norm):
...
+
+    layer_original = edata.layers[DEFAULT_TEM_LAYER_NAME].copy()
+    norm_func(edata, layer=DEFAULT_TEM_LAYER_NAME)
+
+    assert edata.layers[DEFAULT_TEM_LAYER_NAME].shape == orig_shape
+    assert edata.layers[DEFAULT_TEM_LAYER_NAME].dtype == orig_dtype or np.issubdtype(
+        edata.layers[DEFAULT_TEM_LAYER_NAME].dtype, np.floating
+    )
+
+    assert edata.obs.shape == orig_obs_shape
```
eroell: obs and var should not be affected by 3D operations, this does not need to be tested here. this improves the focus on the 3D relevant checks

ehrapy/preprocessing/_normalization.py
agerardy: true, that was to show the difference between before and after at some point, not needed any more

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +664,246 @@ def test_norm_record(adata_to_norm):
...
+    assert edata.var.shape[0] == orig_var_shape[0]
+
+    edata.layers["test_isolated_layer"] = layer_original.copy() * 2 + 5
```
eroell: again, this is not very specific to 3D normalization - that the layer is respected during the function does not need to be tested here again

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +664,246 @@ def test_norm_record(adata_to_norm):
...
+    isolated_original = edata.layers["test_isolated_layer"].copy()
+    norm_func(edata, layer=DEFAULT_TEM_LAYER_NAME)
+    assert np.allclose(edata.layers["test_isolated_layer"], isolated_original, equal_nan=True)
+
+    assert "normalization" in edata.uns
+    assert len(edata.uns["normalization"]) > 0
+
+    edata_invalid = edata_blobs_timeseries_small.copy()
```
eroell: this whole test of `edata_invalid` for simply a non-existent var can be removed - the variable finding is handled the same for 2D and 3D in our code, and this test does not need to check it again - it won't improve the test coverage

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +664,246 @@ def test_norm_record(adata_to_norm):
...
+        assert not np.allclose(
+            edata_select.layers[DEFAULT_TEM_LAYER_NAME][:, 1, :], layer_before_select[:, 1, :], equal_nan=True
+        )
+
+        if edata_select.layers[DEFAULT_TEM_LAYER_NAME].shape[1] > 2:
+            assert np.allclose(
+                edata_select.layers[DEFAULT_TEM_LAYER_NAME][:, 2, :], layer_before_select[:, 2, :], equal_nan=True
+            )
+
+    edata_copy = edata_blobs_timeseries_small.copy()
```
eroell: there is so many copies, I have troubles keeping track :)

If you consider [simple_impute](https://github.com/theislab/ehrapy/blob/7c521027bc072ec4fcf74a3a9bd2359bfc2e3d3a/tests/preprocessing/test_imputation.py#L215) tests, a few lines that are quite quickly understood are preferred over concatenations of many copies of an ehrdata object chaining tests together

ehrapy/preprocessing/_normalization.py
agerardy: dask doesn't support in place assignment with fancy indexing so in line 76 for example `X[np.ix_(group_mask, var_indices)] = scale_func(X[np.ix_(group_mask, var_indices)])` doesn't work, same for line 80. The most reasonable approach I've found is to convert to numpy arrays, normalize and convert back. The dask native approaches look rather complex. Is that a reasonable approach or should we preserve the dask arrays at all cost?

Zethson: Any time you convert to numpy you are materializing (and loading) the whole array at once. Therefore, this must be avoided.

tests/preprocessing/test_normalization.py
agerardy: you're right this function got way out of hand, I'll rewrite it based on simple_impute

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +647,241 @@ def test_norm_record(adata_to_norm):
...
+    if isinstance(layer_data, da.Array):
+        layer_data = layer_data.compute()
+
+    group_a_data = layer_data[:group_size]
+    group_b_data = layer_data[group_size:]
+
+    group_a_flat = group_a_data.flatten()[~np.isnan(group_a_data.flatten())]
+    group_b_flat = group_b_data.flatten()[~np.isnan(group_b_data.flatten())]
+
+    if len(group_a_flat) > 0 and len(group_b_flat) > 0:
```
Zethson: Nit: You could use the new cool match case syntax here.

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +647,241 @@ def test_norm_record(adata_to_norm):
...
+    [
+        ep.pp.scale_norm,
+        ep.pp.minmax_norm,
+        ep.pp.maxabs_norm,
+        ep.pp.robust_scale_norm,
+        ep.pp.quantile_norm,
+        ep.pp.power_norm,
+    ],
+)
+def test_norm_group_3D(edata_blobs_timeseries_small, array_type, norm_func):
```
Zethson, line 867: Because this test is so long, I'd be fine or happy with a few on line comments for blocks that test specific behavior or things.

ehrapy/preprocessing/_normalization.py
```diff
@@ -43,53 +52,105 @@ def _scale_func_group(
...
-            var_values[group_idx] = scale_func(var_values[group_idx])
+        import dask.array as da
+
+        if isinstance(X, DaskArray):
+            result = X
+            for group in edata.obs[group_key].unique():
+                group_idx = edata.obs[group_key] == group
+                group_mask = np.where(group_idx)[0]
+
+                group_data = X[group_mask][:, var_indices] if X.ndim == 2 else X[group_mask][:, var_indices, :]
```
eroell: Is this if needed? My first check worked also without it, and just using `X[group_mask][:, var_indices]`

ehrapy/preprocessing/_normalization.py
```diff
@@ -43,53 +52,105 @@ def _scale_func_group(
...
+                    if len(non_var_indices) > 0:
+                        sorted_indices = np.argsort(np.concatenate([non_var_indices, var_indices]))
+                        updated_group_rows = da.concatenate(
+                            [group_rows[:, non_var_indices], normalized_group_data], axis=1
+                        )[:, sorted_indices]
+                    else:
+                        updated_group_rows = normalized_group_data
+
+                    full_updated_rows = result
+                    for i, obs_idx in enumerate(group_mask):
```
eroell: The items over which a for loop is written should never be individual observations: Python for-loops are very slow. If we need loops, we try to use the smaller axis.

I see the complicaton here with dask arrays; My suggestion is to
1. have a loop-less version as well when all features should be used (this is a frequent case I think)
2. loop over the features instead of the observations here since I ussually expect less features than observations. This is not ideal, but I right now can't come up with something better

ehrapy/preprocessing/_normalization.py
```diff
@@ -43,53 +52,105 @@ def _scale_func_group(
...
+                            updated_obs = normalized_group_data[i : i + 1]
+
+                        obs_mask = da.arange(X.shape[0], chunks=X.chunksize[0]) == obs_idx
+                        full_updated_rows = da.where(obs_mask[:, None, None], updated_obs, full_updated_rows)
+
+                result = full_updated_rows
+
+            X = result
+        else:
+            for group in edata.obs[group_key].unique():
```
eroell:
```py
for group in edata.obs[group_key].unique():
    group_idx = edata.obs[group_key] == group
    group_mask = np.where(group_idx)[0]
    X[np.ix_(group_mask, var_indices)] = scale_func(X[np.ix_(group_mask, var_indices)])
```

Does this not work as well, regardless of the dimension?

ehrapy/preprocessing/_normalization.py
```diff
@@ -43,53 +52,105 @@ def _scale_func_group(
...
+    if group_key is None:
+        X[:, var_indices] = scale_func(X[:, var_indices])
     else:
-        for group in edata.obs[group_key].unique():
-            group_idx = edata.obs[group_key] == group
-            var_values[group_idx] = scale_func(var_values[group_idx])
+        import dask.array as da
+
+        if isinstance(X, DaskArray):
+            result = X
```
eroell: I think you don't need to assign a separate variable here: There's `result`, `X`, `full_updated_rows` and I think they might all be cancelled to just use `X`

tests/preprocessing/test_normalization.py
```diff
@@ -701,22 +647,241 @@ def test_norm_record(adata_to_norm):
...
+    layer_after = edata.layers[DEFAULT_TEM_LAYER_NAME]
+    if isinstance(layer_after, da.Array):
+        layer_after = layer_after.compute()
+    if isinstance(layer_before, da.Array):
+        layer_before = layer_before.compute()
+    assert not np.allclose(layer_before, layer_after, equal_nan=True), (
+        "Normalization did not modify the data - check that feature types are set correctly"
+    )
+
+    layer_data = edata.layers[DEFAULT_TEM_LAYER_NAME]
```
eroell: Why do a second compute here, this is just again the `layer_after` variable?

tests/preprocessing/test_normalization.py
eroell: I have a different question: why even check for this? Isn't the group size is fixed, and even with a few nans will never be 0 for a group?

eroell (changes requested):
This has improved now: The function calls seem to do their job, and from what I see internally, dask never computes the full data.

I'll try to stop being picky :) But there's a few things I spotted that should be improved before we can merge this.

ehrapy/preprocessing/_normalization.py
agerardy: you're right. that wasn't even the AI, I was just overcomplicating things

ehrapy/preprocessing/_normalization.py
```diff
@@ -43,53 +52,57 @@ def _scale_func_group(
...
 
-    if layer is None:
-        var_values = edata[:, vars].X.copy()
-    else:
-        var_values = edata[:, vars].layers[layer].copy()
+    var_indices = _get_var_indices(edata, vars)
+    X = edata.X if layer is None else edata.layers[layer]
+
+    if np.issubdtype(X.dtype, np.integer):
+        X = X.astype(np.float64)
```
Zethson: Why are we casting X to a specific type? This is not obvious behavior, is it?

ehrapy/preprocessing/_normalization.py
```diff
@@ -113,15 +131,22 @@ def scale_norm(
...
-        `None` if `copy=False` and modifies the passed edata, else returns an updated AnnData object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
+        `None` if `copy=False` and modifies the passed edata, else returns an updated object. Also stores a record of applied normalizations as a dictionary in edata.uns["normalization"].
 
     Examples:
         >>> import ehrdata as ed
         >>> import ehrapy as ep
-        >>> edata = ed.dt.mimic_2()
-        >>> edata_norm = ep.pp.scale_norm(edata, copy=True)
+        >>> import numpy as np
+        >>> edata = ed.dt.physionet2012(layer="tem_data")
```
Zethson, line 140: @eroell why do we need the layer parameter here?

ehrapy/preprocessing/_normalization.py
```diff
@@ -438,8 +531,43 @@ def power_norm(
     )
 
 
+@singledispatch
+def _log_norm_function(arr, offset: int | float = 1, base: int | float | None = None):
+    _raise_array_type_not_implemented(_log_norm_function, type(arr))
+
+
+@_log_norm_function.register
```
Zethson: Nit: Could we please always add the type that we're registering here as well? See e.g. Öyku's recent PR.

Zethson (commented):
Just a few more nitpicks (sorry). They'll become less and less as you make more PRs

ehrapy/preprocessing/_normalization.py
agerardy: This is so we don't lose float accuracy. the normalization functions return floats but if we assign them to an int array they can get truncated, so we convert to float before that

Zethson: Ah I see. I think it's very unlikely that users will ever have an integer array in the first place but then I guess it's fine. float64 is kind of expensive though and I wonder whether we can getaway with float32 instead?

ehrapy/preprocessing/_normalization.py
agerardy: if we don't pass the layer it will automatically operate on .X only if i recall correctly

Zethson: But we're just getting a dataset here no? Why does this layer not exist from the beginning?

ehrapy/preprocessing/_normalization.py
agerardy: If i comment it out several tests fail so at least our tests have them :D float32 fails only a single test and it is one of the old ones I didn't add. should I modify that test instead?

Zethson: Depends on the failure but I'd say let's aim for float32, please.

ehrapy/preprocessing/_normalization.py
agerardy: physionet2012 doesn't have the layer by default. the layer argument creates the layer. I don't think we have a testing dataset with the layer already there, right?

ehrapy/preprocessing/_normalization.py
agerardy: i made the assert more lenient, now it works

ehrapy/preprocessing/_normalization.py
Zethson: I don't understand why we need this argument and wouldn't populate this layer by default. If that's the case, we should fix this @eroell

eroell: Since its 3D and can't be in .X, we need to specify a layer at least under the hood. without any arguments, the constant we're currently using everywhere as default is used. This has exactly this name used here

instead of doing it under the hood, I'm in favor of being explicit in the demo and pass the name as is currently done. Once .X is working, it can be skipped and data should by default be in .X

Zethson: Hmm okay. Then let's follow up on this. It bugs me.

PR #967, opened 2025-10-24
Longitudinal and new qc_metrics

### LINKED ISSUE #950: Longitudinal qc_metrics

**3D**
If 3D allowed:

If 3D not allowed:
- [x] only_2D decorator
- [x] the test of the function tests for failure if 3D is passe

**array-type**
- [x] function is single-dispatched, with potential not_implemented errors being raised
- [x] test is parametrized to test the different array-types

**Encoding**
If encoded data is required

sueoglu:
**PR Checklist**

**Description of Changes**

This PR fixes #950 and extends the QC metrics functionality by adding new statistical and information-theoretic measures for both observations and variables. The implementation now supports 3D EHRData objects with longitudinal data while maintaining backward compatibility with 2D data. It updated tests accordingly with new metrics.

**Technical Details**
- New observation-level metrics (in  `_compute_obs_metrics`) :
`unique_values_abs`
`unique_values_ratio`
`entropy_of_missingness`

- New variable-level metrics ( in `_compute_var_metrics `):
`unique_values_abs`
`unique_values_ratio`
`entropy_of_missingness`
`coefficient_of_variation`
`is_constant `
`constant_variable_ratio`
`range_ratio`

- 3D support:
Both private functions (_compute_obs_metrics and _compute_var_metrics) now handle 3D data with shape (n_obs, n_var, n_timestamps) Observation-level metrics use the `apply_to_time_axis` decorator to process temporal dimensions. Variable-level metrics compute statistics across the flattened time dimension

**Examples**

```
import ehrapy as ep
import ehrdata as ed

edata = ed.dt.ehrdata_blobs(n_observations=5,  n_variables=3, base_timepoints=3, missing_values=0.3, seasonality=True)
ed.infer_feature_types(edata)
obs_qc, var_qc = ep.pp.qc_metrics(edata, layer="tem_data")
```

Output: 
- obs_qc:

```
   missing_values_abs  missing_values_pct  entropy_of_missingness  \
0                   2           22.222222                0.764205   
1                   3           33.333333                0.918296   
2                   3           33.333333                0.918296   
3                   5           55.555556                0.991076   
4                   1           11.111111                0.503258   

   unique_values_abs  unique_values_ratio  
0                NaN                  NaN  
1                NaN                  NaN  
2                NaN                  NaN  
3                NaN                  NaN  
4                NaN                  NaN  
```

- var_qc:

```
          missing_values_abs  missing_values_pct  entropy_of_missingness  \
feature_0                   4           26.666667                0.836641   
feature_1                   5           33.333333                0.918296   
feature_2                   5           33.333333                0.918296   

           unique_values_abs  unique_values_ratio  coefficient_of_variation  \
feature_0                NaN                  NaN                 -1.950351   
feature_1                NaN                  NaN                 -1.087513   
feature_2                NaN                  NaN                 45.808075   

           is_constant  constant_variable_ratio   range_ratio      mean  \
feature_0          0.0                      0.0   -531.018674 -3.356061   
feature_1          0.0                      0.0   -375.739531 -0.922850   
feature_2          0.0                      0.0  13557.875431  0.131194   

             median  standard_deviation        min       max  iqr_outliers  
feature_0 -0.881172            6.545496 -13.923100  3.898210         False  
feature_1 -0.818503            1.003611  -2.923477  0.544035         False  
feature_2  3.121145            6.009729 -11.144998  6.642075         False

```

### DIFF
--- a/ehrapy/preprocessing/_quality_control.py
+++ b/ehrapy/preprocessing/_quality_control.py
@@ -98,16 +98,16 @@ def qc_metrics(
         )
 
     feature_type = edata.var.get("feature_type", None)
-    if_advanced = True
+    extended = True
     if feature_type is None:
-        if_advanced = False
+        extended = False
 
     mtx = edata.X if layer is None else edata.layers[layer]
 
     _raise_error_when_heterogeneous(mtx)
 
-    var_metrics = _compute_var_metrics(mtx, edata, advanced=if_advanced)
-    obs_metrics = _compute_obs_metrics(mtx, edata, qc_vars=qc_vars, log1p=True, advanced=if_advanced)
+    var_metrics = _compute_var_metrics(mtx, edata, extended=extended)
+    obs_metrics = _compute_obs_metrics(mtx, edata, qc_vars=qc_vars, log1p=True, extended=extended)
 
     edata.var[var_metrics.columns] = var_metrics
     edata.obs[obs_metrics.columns] = obs_metrics
@@ -230,7 +230,7 @@ def _compute_obs_metrics(
     *,
     qc_vars: Collection[str] = (),
     log1p: bool = True,
-    advanced: bool = False,
+    extended: bool = False,
 ):
     """Calculates quality control metrics for observations.
 
@@ -241,7 +241,7 @@ def _compute_obs_metrics(
         edata: Central data object.
         qc_vars: A list of previously calculated QC metrics to calculate summary statistics for.
         log1p: Whether to apply log1p normalization for the QC metrics. Only used with parameter 'qc_vars'.
-        advanced: Whether to calculate further metrics that require feature type information.
+        extended: Whether to calculate further metrics that require feature type information.
 
     Returns:
         A Pandas DataFrame with the calculated metrics.
@@ -276,12 +276,12 @@ def _compute_obs_metrics(
     obs_metrics["missing_values_pct"] = (obs_metrics["missing_values_abs"] / flat_mtx.shape[1]) * 100
     obs_metrics["entropy_of_missingness"] = _compute_entropy_of_missingness(flat_mtx, axis=1)
 
-    if advanced and "feature_type" not in edata.var:
+    if extended and "feature_type" not in edata.var:
         raise ValueError(
             "Extended QC metrics require `edata.var['feature_type']`. Please run `ehrdata.infer_feature_types(edata)` first"
         )
 
-    if advanced:
+    if extended:
         feature_type = edata.var["feature_type"]
         categorical_mask = feature_type == "categorical"
 
@@ -338,14 +338,14 @@ def _compute_obs_metrics(
 def _compute_var_metrics(
     mtx,
     edata: EHRData | AnnData,
-    advanced: bool = False,
+    extended: bool = False,
 ):
     """Compute variable metrics for quality control.
 
     Args:
         mtx: Data array.
         edata: Central data object.
-        advanced: Whether to calculate further metrics that require feature type information.
+        extended: Whether to calculate further metrics that require feature type information.
     """
     categorical_indices = np.ndarray([0], dtype=int)
     var_metrics = pd.DataFrame(index=edata.var_names)
@@ -378,12 +378,12 @@ def _compute_var_metrics(
     var_metrics["missing_values_pct"] = (var_metrics["missing_values_abs"] / mtx.shape[0]) * 100
     var_metrics["entropy_of_missingness"] = _compute_entropy_of_missingness(mtx, axis=0)
 
-    if advanced and "feature_type" not in edata.var:
+    if extended and "feature_type" not in edata.var:
         raise ValueError(
             "Extended QC metrics require `edata.var['feature_type']`. Please run `ehrdata.infer_feature_types(edata)` first"
         )
 
-    if advanced:
+    if extended:
         feature_type = edata.var["feature_type"]
         categorical_mask = feature_type == "categorical"
 
@@ -456,7 +456,7 @@ def _compute_var_metrics(
         # Fill all non_categoricals with False because else we have a dtype object Series which h5py cannot save
         var_metrics["iqr_outliers"] = var_metrics["iqr_outliers"].astype(bool).fillna(False)
 
-        if advanced:
+        if extended:
             feature_type = edata.var["feature_type"]
             numeric_mask = feature_type == "numeric"
 
--- a/tests/preprocessing/test_quality_control.py
+++ b/tests/preprocessing/test_quality_control.py
@@ -183,7 +183,7 @@ def test_qc_metrics_heterogeneous_columns():
     mtx = np.array([[11, "a"], [True, 22]], dtype=object)
 
     edata = ed.EHRData(shape=(2, 2), layers={"tem_data": mtx})
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Mixed or unsupported"):
         ep.pp.qc_metrics(edata, layer="tem_data")
 
 


ehrapy/preprocessing/_quality_control.py
```diff
@@ -180,14 +250,34 @@ def _compute_var_metrics(
     var_metrics["missing_values_abs"] = _compute_missing_values(mtx, axis=0)
     var_metrics["missing_values_pct"] = (var_metrics["missing_values_abs"] / mtx.shape[0]) * 100
 
+    var_metrics["unique_values_abs"] = _compute_unique_values(mtx, axis=0)
```
eroell: Excellent way to put the computation happening on the matrix into a single-dispatch function

ehrapy/preprocessing/_quality_control.py
```diff
@@ -90,6 +104,49 @@ def _(mtx: DaskArray, axis) -> np.ndarray:
...
+
+    return da.map_blocks(nunique_block, mtx, axis=axis, dtype=int).compute()
+
+
+@singledispatch
+def _compute_entropy_of_missingness(mtx, axis):
+    _raise_array_type_not_implemented(_compute_entropy_of_missingness, type(mtx))
+
+
+@_compute_entropy_of_missingness.register
```
eroell: New with [Lukas' comment](https://github.com/theislab/ehrapy/pull/975#discussion_r2504561273) on my PR very recently; register(<array type>) is a very nice way to make this even a bit more readable

eroell:
> in _compute_obs_metrics :
> unique_values_abs
> unique_values_ratio

These only make sense for categorical data, since for floats this will be not very meaningful.
For ehrapy to know about categorical data, `infer_feature_types` must be called, and I think it would be nice to require this for as little functions as possible.

> entropy_of_missingness

Cool

> in _compute_var_metrics :
> unique_values_abs
> unique_values_ratio

Comment on categorical vs numeric above applies

> entropy_of_missingness

Cool from above applies

> coefficient_of_variation
> is_constant 
> constant_variable_ratio
> range_ratio
> skewness
> kurtosis

All of this require knowledge on categorical and numerical variables - comment above applies.

What do you think about having by default a qc metrics which does only compute the things without feature type information needed; and have e.g. an argument for qc metrics to be computed that is a list, and when someone wants the fancy stuff you suggest here that requires numeric/categorical distinction, they'd need to run infer_feature_types first?

ehrapy/preprocessing/_quality_control.py
```diff
@@ -28,6 +29,7 @@ def qc_metrics(
     qc_vars: Collection[str] = (),
     *,
     layer: str | None = None,
+    advanced: bool = False,
```
eroell: could this be made two arguments 
`observation_level` and `variable_level`, which take lists of strings and by default the lists are what the current default is?

It would seem to me a bit more readable than "advanced"

ehrapy/preprocessing/_quality_control.py
```diff
@@ -130,6 +204,35 @@ def _compute_obs_metrics(
 
     obs_metrics["missing_values_abs"] = _compute_missing_values(mtx, axis=1)
     obs_metrics["missing_values_pct"] = (obs_metrics["missing_values_abs"] / mtx.shape[1]) * 100
+    obs_metrics["entropy_of_missingness"] = _compute_entropy_of_missingness(mtx, axis=1)
+
+    if advanced and "feature_type" in edata.var:
```
eroell: Raising a ValueError if "feature_type" is not in var is fine, asking the user to call `infer_feature_types` first if this specific qc metric is wished for

ehrapy/preprocessing/_quality_control.py
```diff
@@ -78,24 +107,68 @@ def _compute_missing_values(mtx, axis):
...
+
+
+@_compute_entropy_of_missingness.register(DaskArray)
+def _(mtx: DaskArray, axis) -> np.ndarray:
+    import dask.array as da
+
+    missing_mask = da.isnull(mtx)
+    p_miss = missing_mask.mean(axis=axis)
+    p = da.clip(p_miss, 1e-10, 1 - 1e-10)  # avoid log(0)
+    return -(p * da.log2(p) + (1 - p) * da.log2(1 - p)).compute()
```
eroell, line 175: This looks like very good use of dask!
Where we try to never fully compute `.X` or the array in `.layers`, and only call `.compute()` on the final result we're interested in 👍

ehrapy/preprocessing/_quality_control.py
Zethson: If possible, I'd try to keep the number of parameters as low as possible. Can we come up with design where these parameters do not exist? Like a scenario where some things would be skipped over unless the feature specs were calculated but it doesn't crash.

ehrapy/preprocessing/_quality_control.py
```diff
@@ -179,15 +284,52 @@ def _compute_var_metrics(
...
+
+            mtx_cat = mtx[:, cat_mask_np]
+
+            unique_val_abs = _compute_unique_values(mtx_cat, axis=0)
+            missing_cat = _compute_missing_values(mtx_cat, axis=0)
+            valid_counts = mtx_cat.shape[0] - missing_cat
+
+            unique_val_ratio = np.where(
+                valid_counts > 0,
+                unique_val_abs / valid_counts * 100,
```
agerardy, line 405: multiplication *100 is in the function here but outside the function in line 227. Doesnt really matter for function i think, just a small thing i noticed

ehrapy/preprocessing/_quality_control.py
```diff
@@ -217,6 +359,43 @@ def _compute_var_metrics(
         )
         # Fill all non_categoricals with False because else we have a dtype object Series which h5py cannot save
         var_metrics["iqr_outliers"] = var_metrics["iqr_outliers"].astype(bool).fillna(False)
+
+        if advanced:
+            feature_type = edata.var["feature_type"]
+            numeric_mask = feature_type == "numeric"
+
+            non_categorical_indices = np.asarray(numeric_mask)
```
agerardy: non_categorical_indices is a variable name already used in line 282, this could lead to issues. Maybe use numeric_indices? or was this intentional?

agerardy (commented):
Didn't find any big issues, looks pretty good to me already. tests seem to cover everything as far as I understand it.

ehrapy/preprocessing/_quality_control.py
eroell: I wonder if the function computes too many things at once where it adds even more complexity computing 8 metrics in some cases and 12 in another without changing the passed argument. What do you think? I don't have a strong opinion here

The way to address your design that I'd see would be to document well what will be computed if no feature types are found, and what will be computed additionally if they are found.

Zethson: Yeah kinda like that. If possible, we should purge all and any parameters. Users rarely read API docs.

Zethson:
Eventually, I'd like to do a final review because there's a few things that I think need to be changed.

Could you please update the PR description?

ehrapy/preprocessing/_quality_control.py
```diff
@@ -128,11 +216,68 @@ def _compute_obs_metrics(
...
+        if np.any(categorical_mask):
+            cat_mask_np = np.asarray(categorical_mask)
+            if original_mtx.ndim == 2:
+                mtx_cat = mtx[:, cat_mask_np]
+
+                unique_val_abs = _compute_unique_values(mtx_cat, axis=1)
+                missing_cat = _compute_missing_values(mtx_cat, axis=1)
+                valid_counts = mtx_cat.shape[1] - missing_cat
+
+            elif original_mtx.ndim == 3:
```
eroell: Can this branching here and the for loop below be replaced by using `_apply_over_time_axis`?

ehrapy/preprocessing/_quality_control.py
```diff
@@ -28,44 +36,77 @@ def qc_metrics(
     qc_vars: Collection[str] = (),
     *,
     layer: str | None = None,
+    advanced: bool = False,
```
eroell: There is still the decorator `@function_2D_only()` present

tests/conftest.py
```diff
@@ -79,6 +95,29 @@ def missing_values_edata(obs_data, var_data):
...
+def missing_values_edata_adv(obs_data, var_data_adv):
+    return ed.EHRData(
+        X=np.array([[0.21, np.nan, 41.42], [np.nan, np.nan, 7.234]], dtype=np.float32),
+        obs=pd.DataFrame(data=obs_data),
+        var=pd.DataFrame(data=var_data_adv, index=["Acetaminophen", "hospital", "crazy"]),
+    )
+
+
+@pytest.fixture
+def missing_values_edata_3d(obs_data, var_data_adv):
```
eroell: Could [this fixture](https://github.com/theislab/ehrapy/blob/a8aff81af145dc73328ffe042aa661d68ee71fd9/tests/conftest.py#L199) be used instead, to not create an additional test dataset?

tests/conftest.py
```diff
@@ -79,6 +95,29 @@ def missing_values_edata(obs_data, var_data):
     )
 
 
+@pytest.fixture
+def missing_values_edata_adv(obs_data, var_data_adv):
```
eroell: This seems to have just copied the `missing_values_edata` fixture from above, I know you're still working on this so just a hint. Would here be nice to not need a new fixture if another one can help for tests

tests/preprocessing/test_quality_control.py
```diff
@@ -40,6 +42,46 @@ def test_qc_metrics_vanilla(array_type, missing_values_edata):
...
+    assert (~var_metrics["iqr_outliers"]).all()
+
+    # check that none of the columns were modified
+    for key in modification_copy.obs.keys():
+        assert np.array_equal(modification_copy.obs[key], adata.obs[key])
+    for key in modification_copy.var.keys():
+        assert np.array_equal(modification_copy.var[key], adata.var[key])
+
+
 @pytest.mark.parametrize("array_type", ARRAY_TYPES_NONNUMERIC)
```
eroell: Most of the time, we care about tests for the public API.
You have added one above, which is great. From what I see, the test above tests all of what `_compute_obs_metrics` returns, and I think this test for the private function `_compute_obs_metrics` can be deleted.

tests/preprocessing/test_quality_control.py
```diff
@@ -48,6 +90,40 @@ def test_obs_qc_metrics(array_type, missing_values_edata):
 
     assert np.array_equal(obs_metrics["missing_values_abs"].values, np.array([1, 2]))
     assert np.allclose(obs_metrics["missing_values_pct"].values, np.array([33.3333, 66.6667]))
+    assert np.allclose(obs_metrics["entropy_of_missingness"].values, np.array([0.9183, 0.9183]))
+
+
+def test_obs_qc_metrics_3D(missing_values_edata_3d):
```
eroell: Following the comment that we'd like to test public API mainly:

It is neater to write a test alike `test_qc_metrics_vanilla_advanced`, for 3D.

With this, this test for `_compute_obs_metrics` becomes obsolete. (You can reuse the things here for this test of the public function)

Here, the test of the private function passes - but it misses that the public function is not 3D enabled because the decorator is accidentally still there ;)

tests/preprocessing/test_quality_control.py
```diff
@@ -48,6 +90,40 @@ def test_obs_qc_metrics(array_type, missing_values_edata):
...
+    mtx = missing_values_edata_3d.layers["layer_1"]
+    obs_metrics = _compute_obs_metrics(mtx, missing_values_edata_3d)
+
+    assert np.array_equal(obs_metrics["missing_values_abs"].values, np.array([2, 3]))
+    assert np.allclose(obs_metrics["missing_values_pct"].values, np.array([33.3333, 50.0]))
+    assert np.allclose(obs_metrics["entropy_of_missingness"].values, np.array([0.9183, 1.0]))
+
+
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_NONNUMERIC)
```
eroell: You already test for advanced above - I think this test for the private function is obsolete, too.

tests/preprocessing/test_quality_control.py
```diff
@@ -48,6 +90,40 @@ def test_obs_qc_metrics(array_type, missing_values_edata):
...
+    obs_metrics = _compute_obs_metrics(mtx, missing_values_edata_adv, advanced=True)
+
+    assert np.array_equal(obs_metrics["missing_values_abs"].values, np.array([1, 2]))
+    assert np.allclose(obs_metrics["missing_values_pct"].values, np.array([33.3333, 66.6667]))
+    assert np.array_equal(obs_metrics["unique_values_abs"].values, np.array([1, 1]))
+    assert np.allclose(obs_metrics["unique_values_ratio"].values, np.array([100.0, 100.0]))
+    assert np.allclose(obs_metrics["entropy_of_missingness"].values, np.array([0.9183, 0.9183]))
+
+
+def test_obs_qc_metrics_advanced_3D(missing_values_edata_3d):
```
eroell: Can be removed as well in favor of the suggested test for the public API

tests/preprocessing/test_quality_control.py
```diff
@@ -48,6 +90,40 @@ def test_obs_qc_metrics(array_type, missing_values_edata):
...
+    obs_metrics = _compute_obs_metrics(mtx, missing_values_edata_3d, advanced=True)
+
+    assert np.array_equal(obs_metrics["missing_values_abs"].values, np.array([2, 3]))
+    assert np.allclose(obs_metrics["missing_values_pct"].values, np.array([33.3333, 50.0]))
+    assert np.array_equal(obs_metrics["unique_values_abs"].values, np.array([1, 2]))
+    assert np.allclose(obs_metrics["unique_values_ratio"].values, np.array([100.0, 100.0]))
+    assert np.allclose(obs_metrics["entropy_of_missingness"].values, np.array([0.9183, 1.0]))
 
 
 @pytest.mark.parametrize("array_type", ARRAY_TYPES_NONNUMERIC)
```
eroell: Everything for _obs_qc_metrics applies to _var_qc_metrics here, too.

eroell (commented):
This is on a good path.

When all the current comments are addressed, the final review that @Zethson suggested can start. :)

ehrapy/preprocessing/_quality_control.py
```diff
@@ -28,44 +36,77 @@ def qc_metrics(
...
+        - `range_ratio`: Relative dispersion of features values respective to their mean.
+
 
     Examples:
             >>> import ehrapy as ep
             >>> edata = ed.dt.mimic_2()
             >>> obs_qc, var_qc = ep.pp.qc_metrics(edata)
             >>> obs_qc["missing_values_pct"].plot(kind="hist", bins=20)
     """
+    feature_type = edata.var.get("feature_type", None)
```
Zethson, line 100: Yeah, let's please use this and kick "advanced".

tests/conftest.py
sueoglu: I tried using this fixture directly, but it fails during EHRData construction so I'll fix it and use it that way

ehrapy/preprocessing/_quality_control.py
```diff
@@ -42,30 +51,53 @@ def qc_metrics(
...
         Two Pandas DataFrames of all calculated QC metrics for `obs` and `var` respectively.
 
-        Observation level metrics include:
+        Basic (default) observation level metrics include:
 
         - `missing_values_abs`: Absolute amount of missing values.
         - `missing_values_pct`: Relative amount of missing values in percent.
+        - `entropy_of_missingness`: Entropy of the missingness pattern for each observation. Higher values indicate a more heterogeneous (less structured) missingness pattern.
+
+        Advanced observation level metrics include (only computed if advanced is set to `True`):
```
Zethson:
```suggestion
        Advanced observation level metrics include:
```

Not anymore, right? You could specify the new conditions.

ehrapy/preprocessing/_quality_control.py
```diff
@@ -42,30 +51,53 @@ def qc_metrics(
...
+        - `entropy_of_missingness`: Entropy of the missingness pattern for each feature. Higher values indicate a more heterogeneous (less structured) missingness pattern.
         - `mean`: Mean value of the features.
         - `median`: Median value of the features.
         - `std`: Standard deviation of the features.
         - `min`: Minimum value of the features.
         - `max`: Maximum value of the features.
+        - `iqr_outliers`: Whether the feature contains outliers based on the interquartile range (IQR) method.
+
+
+        Advanced feature level metrics include (only computed if ehrdata.infer_feature_types is run first):
```
Zethson: Could you please use a sphinx ref for this?

ehrapy/preprocessing/_quality_control.py
```diff
@@ -128,11 +220,65 @@ def _compute_obs_metrics(
...
+    if mtx.ndim == 2:
+        flat_mtx = mtx
+
+    obs_metrics["missing_values_abs"] = _compute_missing_values(flat_mtx, axis=1)
+    obs_metrics["missing_values_pct"] = (obs_metrics["missing_values_abs"] / flat_mtx.shape[1]) * 100
+    obs_metrics["entropy_of_missingness"] = _compute_entropy_of_missingness(flat_mtx, axis=1)
+
+    if advanced and "feature_type" not in edata.var:
+        raise ValueError(
+            "Advanced QC metrics require `edata.var['feature_type']`. Please run `infer_feature_types(edata)` first"
```
Zethson:
```suggestion
            "Advanced QC metrics require `edata.var['feature_type']`. Please run `ed.infer_feature_types(edata)` first"
```

right? or `ehrdata`? Just to make it explicit.

ehrapy/preprocessing/_quality_control.py
```diff
@@ -32,7 +39,9 @@ def qc_metrics(
     """Calculates various quality control metrics.
 
     Uses the original values to calculate the metrics and not the encoded ones.
-    Look at the return type for a more in depth description of the calculated metrics.
+    Look at the return type for a more in depth description of the basic and advanced calculated metrics.
+    If ehrdata.infer_feature_types is run first, then advanced metrics are calculated in addition to basic metrics that require feature type information.
```
eroell: ehrdata.infer_feature_types can be linked nicely with [intersphinx mappings](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html)

ehrapy/preprocessing/_quality_control.py
```diff
@@ -42,30 +51,53 @@ def qc_metrics(
...
+        - `is_constant`: Whether the feature is constant (with near zero variance).
+        - `constant_variable_ratio`: Relative amount of constant features in percent.
+        - `range_ratio`: Relative dispersion of features values respective to their mean.
+
 
     Examples:
             >>> import ehrapy as ep
             >>> edata = ed.dt.mimic_2()
             >>> obs_qc, var_qc = ep.pp.qc_metrics(edata)
             >>> obs_qc["missing_values_pct"].plot(kind="hist", bins=20)
```
eroell: could the example be instead the `obs_qc.head()` and the `var_qc.head()` outputs?

ehrapy/preprocessing/_quality_control.py
```diff
@@ -42,30 +51,53 @@ def qc_metrics(
...
         - `median`: Median value of the features.
         - `std`: Standard deviation of the features.
         - `min`: Minimum value of the features.
         - `max`: Maximum value of the features.
+        - `iqr_outliers`: Whether the feature contains outliers based on the interquartile range (IQR) method.
+
+
+        Advanced feature level metrics include (only computed if ehrdata.infer_feature_types is run first):
+
+        - `unique_values_abs`: Absolute amount of unique values.
```
eroell: There seems to be an issue with indentation on readthedocs

ehrapy/preprocessing/_quality_control.py
```diff
@@ -128,11 +220,65 @@ def _compute_obs_metrics(
...
+            else:  # ndim == 3
+                mtx_cat = original_mtx[:, cat_mask_np, :]  # (n_obs, n_cat_var, n_time)
+
+            unique_arr = _row_unique(mtx_cat, axis=1)
+            valid_arr = _row_valid(mtx_cat, axis=1)
+
+            if unique_arr.ndim == 2:
+                unique_val_abs = unique_arr[:, 0]
+                valid_counts = valid_arr[:, 0]
+
```
Zethson:
```suggestion
```

ehrapy/_compat.py
```diff
@@ -248,3 +249,25 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
         return wrapper
 
     return decorator
+
+
+def asarray(a):
```
Zethson: This shouldn't exist. Use fast-array-utils, please.

ehrapy/_compat.py
```diff
@@ -248,3 +249,25 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+    import numpy as np
+
+    return np.asarray(a)
+
+
+def as_dense_dask_array(a, chunk_size=1000):
```
Zethson, line 254: Typehints. Do we really need a function for that? But I'm okay with it.

ehrapy/_compat.py
```diff
@@ -248,3 +249,25 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+    import dask.array as da
+
+    return da.from_array(a, chunks=chunk_size)
+
+
+ARRAY_TYPES_NUMERIC = (
```
Zethson: Here or in https://github.com/theislab/ehrapy/blob/main/ehrapy/_types.py ?

Zethson:
Yup, this sounds better. Thanks!

ehrapy/_compat.py
sueoglu: I can move these to _types.py  if it's better

Zethson: I'd say yes

PR #969, opened 2025-10-28
Fix/fix use rep when x none

eroell:
**PR Checklist**

**Description of changes**

`ep.pp.neighbors` if `.X` is `None` but using `use_rep` like in

```py
def test_neighbors(edata_blob_small):
    ep.pp.neighbors(edata_blob_small, n_neighbors=5)

    # since use_rep="..." is possible, check edgecase where X is None and layers invalid
    edata_blob_small.obsm["X_pca"] = rng.random((edata_blob_small.n_obs, 5))
    edata_blob_small.X = None
    ep.pp.neighbors(edata_blob_small, use_rep="X_pca", n_neighbors=5)
``` 

generates the following error
```
File ~<...>/ehrapy/ehrapy/_compat.py:214, in function_2D_only.<locals>.decorator.<locals>.wrapper(*args, **kwargs)
    211 if data is not None:
    212     array = data.X if layer is None else data.layers[layer]
--> 214     if array.ndim != 2 and array.shape[2] != 1:
    215         raise ValueError(
    216             f"{func.__name__}() only supports 2D data, got {'data.X' if layer is None else f'data.layers[{layer}]'} with shape {array.shape}"
    217         )
    219 return func(*args, **kwargs)

AttributeError: 'NoneType' object has no attribute 'ndim'
```

This PR fixes this, and adds a test.

### DIFF
--- a/tests/preprocessing/test_scanpy_pp_api.py
+++ b/tests/preprocessing/test_scanpy_pp_api.py
@@ -42,7 +42,6 @@ def test_neighbors(edata_blob_small):
     ep.pp.neighbors(edata_blob_small, n_neighbors=5)
 
     # since use_rep="..." is possible, check edge case where X is None and layers invalid
-    rng = np.random.default_rng(42)
-    edata_blob_small.obsm["X_pca"] = rng.random((edata_blob_small.n_obs, 5))
+    edata_blob_small.obsm["X_pca"] = np.random.default_rng(42).random((edata_blob_small.n_obs, 5))
     edata_blob_small.X = None
     ep.pp.neighbors(edata_blob_small, use_rep="X_pca", n_neighbors=5)


tests/preprocessing/test_scanpy_pp_api.py
```diff
@@ -35,3 +36,13 @@ def test_combat_3D_edata(edata_blob_small):
     ep.pp.combat(edata_blob_small, key="cluster", layer="layer_2")
     with pytest.raises(ValueError, match=r"only supports 2D data"):
         ep.pp.combat(edata_blob_small, key="cluster", layer="R_layer")
+
+
+def test_neighbors(edata_blob_small):
+    ep.pp.neighbors(edata_blob_small, n_neighbors=5)
+
+    # since use_rep="..." is possible, check edgecase where X is None and layers invalid
```
Zethson:
```suggestion
    # since use_rep="..." is possible, check edge case where X is None and layers invalid
```

tests/preprocessing/test_scanpy_pp_api.py
```diff
@@ -35,3 +36,13 @@ def test_combat_3D_edata(edata_blob_small):
...
+    rng = np.random.default_rng(42)
```
Zethson: Nit: If you only use the rng object once you can in-line it.

PR #971, opened 2025-11-04
Move to ehrdata 0.0.10

eroell:
**PR Checklist**

**Description of changes**

ehrdata's [release to 0.0.10](https://github.com/theislab/ehrdata/releases/tag/v0.0.10) gets rid of `.R`, and moves all to layers. Adjust ehrapy here to use this

**Technical details**

**Additional context**

### DIFF
--- a/ehrapy/preprocessing/_neighbors.py
+++ b/ehrapy/preprocessing/_neighbors.py
@@ -66,7 +66,7 @@ def neighbors(
                      In general values should be in the range 2 to 100. If `knn` is `True`, number of nearest neighbors to be searched.
                      If `knn` is `False`, a Gaussian kernel width is set to the distance of the `n_neighbors` neighbor.
         n_pcs: Use this many PCs. If `n_pcs==0` use `.X` if `use_rep is None`.
-        use_rep: Use the indicated representation. `'X'` or `.obsm` is valid.
+        use_rep: Use the indicated representation. `'X'` or any key for `.obsm` is valid. For time series data (`metric='dtw'`, `'soft_dtw'`, `'gak'`), the key must be a 3D array with shape (n_obs, n_vars, n_timepoints).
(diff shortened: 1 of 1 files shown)


ehrapy/preprocessing/_filter.py
```diff
@@ -193,107 +184,72 @@ def filter_observations(
...
                 msg += f" in < {prop * 100:.1f}% of time points"
             else:
                 msg += f" in {time_mode} time points"
 
         logger.info(msg)
 
-    label = "n_vars" if is_2d_ref else "n_vars_over_time"
-    data.obs[label] = first_number_per_obs
-    data._inplace_subset_obs(final_obs_mask)
+    label = "n_observations" if is_2d else "n_observations_over_time"
```
sueoglu: why is the label `n_observations`? since it filters the observations out I think it should have been `n_vars` for each observation that is kept

sueoglu: also` n_vars_over_time` if not 2d

eroell: What is your take alternatively on 
`label = "n_nonmissing" if is_2d else "n_nonmissing_over_time"`?

Mentioning `nonmissing` helps to understand what's in the column.

this is what I intended to express with `n_observation`, but I agree this is not a great idea with the overloading of the word observation here

eroell: in `use constants for missingness & test them` added what it would look like with `n_nomissing`, also fixed something with 2d and 3d

sueoglu: I think with `n_nonmissing `its and `n_nonmissing_over_time ` its more straightforward what you mean so I would agree to have it as this

eroell: OK! Thanks for the comment

PR #975, opened 2025-11-07
Simple Impute for timeseries

### LINKED ISSUE #945: Longitudinal simple_impute

**3D**
If 3D allowed:
- [x] function handles 3D data
- [x] the test of the function tests both the expected 2D and 3D implementation

If 3D not allowed:
- [ ] only_2D decorator

**array-type**
- [x] function is single-dispatched, with potential not_implemented errors being raised
- [x] test is parametrized to test the different array-types

**Encoding**
If encoded data is required

eroell:
**PR Checklist**

**Description of changes**

Fixes #945, see there for a checklist

`ehrapy.pp.simple_impute` for timeseries.

**Technical details**

**Additional context**

### DIFF
--- a/ehrapy/preprocessing/_imputation.py
+++ b/ehrapy/preprocessing/_imputation.py
@@ -200,7 +200,7 @@ def simple_impute(
     Args:
         edata: Central data object.
         var_names: A list of column names to apply imputation on (if None, impute all columns).
-        strategy: Imputation strategy to use. One of {'mean', 'median', 'most_frequent'}. If data is a :class:`dask.array.Array`, only 'mean' is supported.
+        strategy: Imputation strategy to use. One of {'mean', 'median', 'most_frequent'}. If data is a `dask.array.Array`, only 'mean' is supported.
         warning_threshold: Display a warning message if percentage of missing values exceeds this threshold.
         layer: The layer to impute.
         copy: Whether to return a copy of `edata` or modify it inplace.


ehrapy/_compat.py
```diff
@@ -165,6 +165,31 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+            return f(arr, *args, **kwargs)
+
+        elif arr.ndim == 3:
+            n_obs, n_vars, n_time = arr.shape
+            arr_2d = arr.transpose(0, 2, 1).reshape(-1, n_vars)
+            arr_modified_2d = f(arr_2d, *args, **kwargs)
+            return arr_modified_2d.reshape(n_obs, n_time, n_vars).transpose(0, 2, 1)
+
+        else:
+            raise ValueError(f"Unsupported array dimensionality: {arr.ndim}")
```
Zethson: Always suggest the resolution -> reshape to 3d or 2d

ehrapy/_compat.py
```diff
@@ -165,6 +165,31 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+    """Decorator to allow functions to handle both 2D and 3D arrays.
+
+    - If the input is 2D: pass it through unchanged.
+    - If the input is 3D: reshape to 2D before calling the function,
+      then reshape the result back to 3D.
+    """
+
+    @wraps(f)
+    def wrapper(arr, *args, **kwargs):
+        if arr.ndim == 2:
```
Zethson, line 179: You're only using the array API here, right? So this should work for all array types I hope. Maybe double check...

ehrapy/_compat.py
```diff
@@ -165,6 +165,31 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
     return decorator
 
 
+def _support_3d(f: Callable) -> Callable:
+    """Decorator to allow functions to handle both 2D and 3D arrays.
+
+    - If the input is 2D: pass it through unchanged.
+    - If the input is 3D: reshape to 2D before calling the function,
+      then reshape the result back to 3D.
```
Zethson: This can go into the line just above. Your autoformatter makes it a bit ugly.

ehrapy/preprocessing/_imputation.py
```diff
@@ -148,8 +146,29 @@ def _extract_impute_value(replacement: dict[str, str | int], column_name: str) -
         return None
 
 
+@singledispatch
+def _simple_impute_function(arr, strategy: Literal["mean", "median", "most_frequent"]) -> None:
+    _raise_array_type_not_implemented(_simple_impute_function, type(arr))
+
+
+@_simple_impute_function.register
```
Zethson: It's also possible to register the type very explicitly here like e.g. in https://github.com/scverse/pertpy/blob/main/pertpy/preprocessing/_guide_rna.py#L79 which makes it more readable in my humble opinion.

ehrapy/preprocessing/_imputation.py
```diff
@@ -148,8 +146,29 @@ def _extract_impute_value(replacement: dict[str, str | int], column_name: str) -
         return None
 
 
+@singledispatch
```
Zethson, line 156: Do you already want to add the scipy.sparse stub here because you want this to serve as a role model? I know that scipy hasn't made it's release yet.

ehrapy/preprocessing/_imputation.py
```diff
@@ -148,8 +146,29 @@ def _extract_impute_value(replacement: dict[str, str | int], column_name: str) -
...
+@_support_3d
+def _(arr: np.ndarray, strategy: Literal["mean", "median", "most_frequent"]) -> np.ndarray:
+    import sklearn
+
+    return sklearn.impute.SimpleImputer(strategy=strategy).fit_transform(arr)
+
+
 @use_ehrdata(deprecated_after="1.0.0")
-@function_2D_only()
 @spinner("Performing simple impute")
```
Zethson:
```suggestion
```

This can be removed by now for such trivial functions.

ehrapy/_compat.py
```diff
@@ -165,6 +165,31 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
     return decorator
 
 
+def _support_3d(f: Callable) -> Callable:
```
Zethson: The name is pretty general. Can we be more specific?

tests/preprocessing/test_imputation.py
```diff
@@ -147,10 +149,34 @@ def test_base_check_imputation_change_detected_in_imputed_column(impute_num_edat
         _base_check_imputation(impute_num_edata, edata_imputed)
 
 
+@pytest.mark.parametrize(
+    "array_type,expected_error",
+    [
+        (np.array, None),
+        (da.array, None),
+        (sparse.csr_matrix, NotImplementedError),
```
Zethson: We'll likely with work scipy.sparse.coo . But in any case, certainly not with the old matrix classes. array API ftw

ehrapy/_compat.py
agerardy: maybe support_3d_layers, since it also doesn't work with .X yet

tests/conftest.py
```diff
@@ -142,7 +142,11 @@ def mcar_edata(rng) -> ed.EHRData:
     missing_indices = rng.choice(a=[False, True], size=data.shape, p=[1 - 0.1, 0.1])
     data[missing_indices] = np.nan
 
-    return ed.EHRData(data)
+    data_3d = rng.random((100, 10, 3))
```
agerardy, line 145: so mcar_edata will be the new standard 3d test fixture? or is there another reason why this one was modified?

ehrapy/_compat.py
```diff
@@ -165,6 +165,31 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+    @wraps(f)
+    def wrapper(arr, *args, **kwargs):
+        if arr.ndim == 2:
+            return f(arr, *args, **kwargs)
+
+        elif arr.ndim == 3:
+            n_obs, n_vars, n_time = arr.shape
+            arr_2d = arr.transpose(0, 2, 1).reshape(-1, n_vars)
+            arr_modified_2d = f(arr_2d, *args, **kwargs)
+            return arr_modified_2d.reshape(n_obs, n_time, n_vars).transpose(0, 2, 1)
```
agerardy: so the arrays are flattened, the function this wrapper is placed on is called on each timestep? did i get that right?

ehrapy/_compat.py
agerardy: we can just reshape higher dimensional data?

agerardy (commented):
I think I understood it :)

ehrapy/_compat.py
sueoglu: only 3d and 2d are supported by this decorator right?

tests/conftest.py
```diff
@@ -142,7 +142,11 @@ def mcar_edata(rng) -> ed.EHRData:
...
+    missing_indices = rng.choice(a=[False, True], size=data_3d.shape, p=[1 - 0.1, 0.1])
```
sueoglu, line 146: this is for the missing data points in the 3d space right? if no what is the difference between this one (line 146) and the first one (line 142)

ehrapy/_compat.py
eroell: I'll indicate 2D/3D.

Reshaping different dimensions would need choices how to do this, we'll stick to 2D and 3D for this reason..

Only 2D and 3D are supported, yes.

ehrapy/preprocessing/_imputation.py
eroell: `sklearn.impute.SimpleImputer`even works with scipy's `csr_array` and `csc_array`, I added these two to the existing function.

I added a specific stub now for `coo_array` which we'd want to support too. Is this what you meant here?

ehrapy/_compat.py
eroell: changed to `_apply_over_time_axis`, which is more descriptive of what this is doing.
Thank you for the suggestion @agerardy! I omitted putting layers into its name, since this function only receives an array and doesn't (need to) know about where the array comes from :)

tests/conftest.py
eroell: The reason that this was modified is that before, we didn't have a fixture with a 3D array that has missing values.

And as you've seen right, `mcar_edata` can be used as the fixture for the minimal 3D test case, which is called`test_simple_impute_3D_edata` here.

ehrapy/_compat.py
eroell: The arrays are flattened, yes; the function is then applied on a 2D array that stacks all timesteps

tests/conftest.py
eroell: Yes, exactly this is for missing data points in a 3D array

ehrapy/_compat.py
eroell: I added some checks in `test_compat` that run this function with our arrays of interest, which actually prompted me to use `np.moveaxis()` instead of `.transpose()`.

Sidenote:
`.transpose()` caused an error apparently with a specific scipy version, which I did not drill down to since I switch to `np.moveaxis()`. It would have seemed to work with the latest version though, but I do not trust its stability because of [their documentation of the axis parameter in coo_array.tranpose()](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_array.transpose.html#scipy.sparse.coo_array.transpose)

ehrapy/preprocessing/_imputation.py
Zethson: Yes.

tests/preprocessing/test_imputation.py
```diff
@@ -147,10 +149,36 @@ def test_base_check_imputation_change_detected_in_imputed_column(impute_num_edat
...
+    "array_type,expected_error",
+    [
+        (np.array, None),
+        (da.array, None),
+        (sparse.csr_array, None),
+        (sparse.csc_array, None),
+        # (sparse.coo_array, None) # not yet supported by AnnData
+    ],
+)
+def test_simple_impute_array_types(impute_num_edata, array_type, expected_error):
```
eroell, line 162: Explanation: This first simple test should check that the function executes for array types it supports with a simple `edata`, and default function arguments; If an array type is not supported, a proper NotImplementedError should be raised by the function call.

tests/preprocessing/test_imputation.py
```diff
@@ -147,10 +149,36 @@ def test_base_check_imputation_change_detected_in_imputed_column(impute_num_edat
...
+        with pytest.raises(expected_error):
+            simple_impute(impute_num_edata, strategy="mean")
+
+
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_NUMERIC)
 @pytest.mark.parametrize("strategy", ["mean", "median", "most_frequent"])
-def test_simple_impute_basic(impute_num_edata, strategy):
-    edata_imputed = simple_impute(impute_num_edata, strategy=strategy, copy=True)
-    _base_check_imputation(impute_num_edata, edata_imputed)
+def test_simple_impute_basic(impute_num_edata, array_type, strategy):
```
eroell, line 172: Explanation: This test and the following one are specific for the function, which is about imputation here

tests/preprocessing/test_imputation.py
```diff
@@ -161,19 +189,81 @@ def test_simple_impute_copy(impute_num_edata, strategy):
...
+            assert edata_imputed.X[0, 1] == 2.0  # if multiple equally frequent values, return minimum
 
-def test_simple_impute_3D_edata(edata_blob_small):
-    simple_impute(edata_blob_small, layer="layer_2")
-    with pytest.raises(ValueError, match=r"only supports 2D data"):
-        simple_impute(edata_blob_small, layer=DEFAULT_TEM_LAYER_NAME)
+
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_NUMERIC_3D_ABLE)
+@pytest.mark.parametrize("strategy", ["mean", "median", "most_frequent"])
+def test_simple_impute_3D_edata(mcar_edata, array_type, strategy):
```
eroell, line 215: Explanation: this test checks that for numeric data, the method can handle 3D data.

There might be combinations of arguments that are not supported, which should be mentioned in the documentation if not obvious.

tests/preprocessing/test_imputation.py
```diff
@@ -161,19 +189,81 @@ def test_simple_impute_copy(impute_num_edata, strategy):
...
+                reference_value = np.nanmean(mcar_edata[:, 0, :].layers[DEFAULT_TEM_LAYER_NAME])
+            elif strategy == "median":
+                reference_value = np.nanmedian(mcar_edata[:, 0, :].layers[DEFAULT_TEM_LAYER_NAME])
+
+            assert np.isclose(element, reference_value)
+
+
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_NONNUMERIC)
+@pytest.mark.parametrize("strategy", ["mean", "median", "most_frequent"])
+def test_simple_impute_3D_edata_nonnumeric(edata_mini_3D_missing_values, array_type, strategy):
```
eroell, line 245: Explanation: this checks whether the function can also handle non-numeric 3D data.

There might be combinations of arguments that are not supported, which should be mentioned in the documentation if not obvious.

tests/conftest.py
```diff
@@ -415,4 +434,11 @@ def as_dense_dask_array(a, chunk_size=1000):
     return da.from_array(a, chunks=chunk_size)
 
 
-ARRAY_TYPES = (asarray, as_dense_dask_array)
+ARRAY_TYPES_NUMERIC = (
```
eroell, line 437: I suggest these collections of array types for testing, which I hope reduces some repetitiveness

tests/preprocessing/test_imputation.py
```diff
@@ -161,19 +189,81 @@ def test_simple_impute_copy(impute_num_edata, strategy):
...
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_NONNUMERIC)
 @pytest.mark.parametrize("strategy", ["mean", "median", "most_frequent"])
-def test_simple_impute_subset(impute_edata, strategy):
+def test_simple_impute_subset(impute_edata, array_type, strategy):
+    impute_edata.X = array_type(impute_edata.X)
     var_names = ("intcol", "indexcol")
-    edata_imputed = simple_impute(impute_edata, var_names=var_names, copy=True)
+    if isinstance(impute_edata.X, da.Array) and strategy != "mean":
+        with pytest.raises(ValueError):
+            edata_imputed = simple_impute(impute_edata, var_names=var_names, strategy=strategy, copy=True)
```
agerardy, line 199: so the function is still called normally when it is a dask array and strategy is not mean? shouldnt it just raise the valueerror?

eroell: The test asserts that a ValueError is raise if the function is called like this; and this will also finish the test for this function, so all the below rows are not executed. Does this answer your question? :)

ehrapy/_compat.py
Zethson: Does the numpy function work with numpy, scipy sparse, and Dask?

tests/conftest.py
Zethson: I wonder whether we should have this here or even in our compat code because we might also use this for the implementations?

ehrapy/_compat.py
eroell: Yes, I added a test for numpy, scipy coo, and dask 3D arrays that this is properly passed through. For scipy csr/csc array, the numpy function is not required since they are 2D only.

tests/conftest.py
eroell: Agree actually, didn't occur to me before. -> asked @sueoglu to move in #967

PR #989, opened 2025-11-26
Add Sankey diagram visualization functions 

sueoglu:
fixes #232 
two new visualization functions sankey_diagram and sankey_diagram_time for creating Sankey diagrams to analyze patient flows and state transitions in EHR data, added comprehensive tests

sankey_diagram : relationships across observation columns
sankey_diagram_time: observation state transitions over time

**Examples**

- ep.pl.sankey_diagram()

```
import ehrapy as ep
import ehrdata as ed

edata = diabetes_130_fairlearn_sample_100()
ep.pl.sankey_diagram(edata, columns=["gender", "race"])
```
Output: 

image

- ep.pl.sankey_diagram_time()

```
import ehrapy as ep
import ehrdata as ed

edata = ed.dt.ehrdata_blobs(base_timepoints=5, n_variables=1, n_observations=5, random_state=59)
edata.layers["tem_data] = edata.layers["tem_data].astype(int)
sankey_time = ep.pl.sankey_diagram_time(edata, var_name="feature_0", layer="tem_data", state_labels={-2: "no", -3: "mild", -4: "moderate", -5: "severe", -6: "critical"})
```
Output: 
image

PR #994, opened 2025-12-10
Add `ep.pl.timeseries()` to visualize variables over time

sueoglu:
This PR introduces plot_timeseries(), a new plotting function for timeseries data from 3D EHRData layers. This function enables users to visualise how given features/variables change over time for individual patients or across multiple patients. 

**Examples**

```python
import ehrapy as ep
import ehrdata as ed
edata = ed.dt.ehrdata_blobs(n_variables=2, n_observations=10, base_timepoints=100, layer="tem_name")
edata.var.index = ["heart_rate", "oxygen_sat"]
ep.pl.timeseries(edata, obs_names="2", var_names=["heart_rate", "oxygen_sat"],tem_names=slice(0,20), layer="tem_name")

```
image

### DIFF
--- a/ehrapy/plot/_timeseries.py
+++ b/ehrapy/plot/_timeseries.py
@@ -49,13 +49,12 @@ def timeseries(
         HoloViews Overlay (if overlay=True) or Layout (if overlay=False) object representing the time series plot(s).
 
     Examples:
-    >>> import ehrapy as ep
-    >>> import ehrdata as ed
-    >>> edata = ed.dt.ehrdata_blobs(n_variables=10, n_observations=5, base_timepoints=100)
-    >>> ep.pl.timeseries(edata, obs_names="1", var_names=["feature_1", "feature_2"], tem_names=slice(0, 10))
+        >>> import ehrapy as ep
+        >>> import ehrdata as ed
+        >>> edata = ed.dt.ehrdata_blobs(n_variables=10, n_observations=5, base_timepoints=100)
+        >>> ep.pl.timeseries(edata, obs_names="1", var_names=["feature_1", "feature_2"], tem_names=slice(0, 10))
 
-    .. image:: /_static/docstring_previews/timeseries_plot.png
-
+        .. image:: /_static/docstring_previews/timeseries_plot.png
     """
     opts_dict: dict[str, Any] = {}
     if width is not None:


ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+    ylabel: str | None = None,
+    title: str | None = None,
+    show: bool = True,
+) -> Axes | Sequence[Axes] | None:
+    """Plot variable time series either for an observation or multiple observations from a 3D EHRData layer.
+
+    Selection logic:
+        - If obs_id is an int in [0, n_obs), use it as row index.
+        - Otherwise, obs_id_key must be a column name in edata.obs, and
+          obs_id is matched against that column.
```
Zethson: Nit: Think we can move this up to the line above.

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+    from collections.abc import Sequence
+
+    from ehrdata import EHRData
+    from matplotlib.axes import Axes
+
+
+def plot_timeseries(
+    edata: EHRData,
+    obs_id: str | int | Sequence[str | int],
+    keys: str | Sequence[str],
```
Zethson: Maybe we should call it `var_key`? keys could be absolutely anything

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+        - If obs_id is an int in [0, n_obs), use it as row index.
+        - Otherwise, obs_id_key must be a column name in edata.obs, and
+          obs_id is matched against that column.
+
+    Args:
+        edata: Central data object.
+        obs_id: row index or observation identifier(s) to plot.
+        keys: Variable name or list of variable names in edata.var_names to plot.
+        layer: layer to use for time series data.
+        obs_id_key: Column in edata.obs to match obs_id against (if obs_id is not given as row index).
```
Zethson:
```suggestion
        obs_id_key: Column in `edata.obs` to match obs_id against (if obs_id is not given as row index).
```

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+        tem_time_key: Key in edata.tem to use as timepoints. If None, use edata.tem as 1D array.
```
Zethson:
```suggestion
        tem_time_key: Key in `edata.tem` to use as timepoints. If None, use `edata.tem` as 1D array.
```

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+        overlay: Whether to overlay multiple observations in a single plot (True) or create subplots (False).
+        xlabel: The x-axis label text.
+        ylabel: The y-axis label text.
+        title: Set the title of the plot.
+        show: Show the plot, do not return axis.
```
Zethson: Tricky because I think holoviews displays by default. I don't have this for the survival analysis plots anymore.

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Returns:
+        Axes object or Sequence of Axes or None
```
Zethson: Hmm, I think we don't return this for bokeh plots do we?

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Examples:
+        >>> edata = ed.dt.ehrdata_blobs(
+        ...     n_variables=4,
+        ...     n_observations=10,
+        ...     base_timepoints=100,
+        ...     layer=DEFAULT_TEM_LAYER_NAME,
+        ... )
+        >>> edata.var.index = ["feature1", "feature2", "feature3", "feature4"]
```
Zethson: Is there nothing in the blobs function that we can use? Ideally, we don't need this line and have something in blobs. If we don't then we should probably change the blobs function to populate it with some crap

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from ehrdata import EHRData
+    from matplotlib.axes import Axes
+
+
+def plot_timeseries(
+    edata: EHRData,
+    obs_id: str | int | Sequence[str | int],
```
Zethson: I think it's rather weird to use indices. If anything, maybe a range of indices? @eroell wdyt?

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+    else:
+        obs_ids = list(obs_id)
+
+    mtx = np.asarray(edata.layers[layer])
+    if mtx.ndim != 3:
+        raise ValueError(f"Layer {layer!r} must be 3D (n_obs, n_vars, n_time), got shape {mtx.shape}.")
+    n_obs, _, n_time = mtx.shape
+
+    if tem_time_key is None:
+        warnings.warn(
```
Zethson: I don't think that we should emit this warning and just roll with it.

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+        var_idx_list.append(int(matches[0]))
+
+    if overlay:
+        if len(key_list) != 1:
+            raise ValueError("When overlay=True, only a single key can be plotted at a time.")
+        n_panels = 1
+    else:
+        n_panels = len(obs_ids)
+
+    fig, axes = plt.subplots(
```
Zethson: Oh, this is not a holoviews plot. Could we implement this with holoviews?

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+    if show:
+        plt.show()
+        return None
+    else:
+        if n_panels == 1:
+            return axes[0]
+        return axes
+
+
+def _resolve_obs(obs, obs_id_key, n_obs, edata) -> tuple[int, str]:
```
Zethson: Typehints, please. Usually edata is the first argument

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+    """Resolve obs identifier to (row index, obs_id_info) tuple.
+
+    Args:
+        obs: Integer row index or label to match in `edata.obs[obs_id_key]`.
+        obs_id_key: Column used when `obs` is not a valid row index.
+        n_obs: Number of observations for index validation.
+        edata: EHR data object providing the `.obs` table.
```
Zethson: Not an EHR data object. It's something very specific :)

ehrapy/plot/_timeseries.py
eroell: I suggest to put only edata as mandatory, positional argument.

If nothing else is specified, it could just plot all patients, features, and timepoints

ehrapy/plot/_timeseries.py
eroell: scanpy loves [var_names](https://scanpy.readthedocs.io/en/stable/generated/scanpy.pl.dotplot.html)

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+    """Plot variable time series either for an observation or multiple observations from a 3D EHRData layer.
+
+    Selection logic:
+        - If obs_id is an int in [0, n_obs), use it as row index.
+        - Otherwise, obs_id_key must be a column name in edata.obs, and
+          obs_id is matched against that column.
+
+    Args:
+        edata: Central data object.
+        obs_id: row index or observation identifier(s) to plot.
```
eroell: What is your take on just asking users to specify `obs_names` (similar as to var_names above)? That would allow to drop the `obs_id_key` argument
This should also be unique, which matches this function's intention

ehrapy/plot/_timeseries.py
eroell: ehrdata doesn't have `tem_names` analog to `obs_names` and `var_names`.

My suggestion is to add an argument `tem_names`, which uses the index of `ehrdata.tem`. This is also what `.var_names` and `.obs_names` basically is.

It would be more a repetitive pattern for users. wdyt @sueoglu ?

ehrapy/plot/_timeseries.py
eroell: It should be usable without any arguments here. 
`ep.pl.plot_time_series(..., layer="tem_data")` is used below.
Since users wonder "where is tem_data coming from?", I tend to recommend that the single argument that should be specified is `layer="tem_data"` in `ehrdata_blobs`

tests/plot/test_timeseries.py
```diff
@@ -0,0 +1,153 @@
...
+import ehrapy as ep
+
+matplotlib.use("Agg")
+
+CURRENT_DIR = Path(__file__).parent
+_TEST_IMAGE_PATH = f"{CURRENT_DIR}/_images"
+
+
+@pytest.fixture
+def edata_blob_timeseries():
```
eroell: is there a fixture in `conftest.py` that can be used instead of creating a new one here?

tests/plot/test_timeseries.py
```diff
@@ -0,0 +1,153 @@
...
+        obs_id=2,
+        keys=["feature1", "feature2", "feature3"],
+        layer=DEFAULT_TEM_LAYER_NAME,
+        tem_time_key="timepoint",
+        show=False,
+    )
+
+    fig = ax.figure
+
+    check_same_image(
```
eroell: amazing, the plot test expert flexing her testing skills on top of implementing a new feature

eroell (commented):
This is a very cool first shot. Throwing in a few viewpoints additionally to @Zethson .

ehrapy/plot/_timeseries.py
Zethson: Would likely lead to a very crowded plot but I totally agree.

tests/plot/test_timeseries.py
```diff
@@ -0,0 +1,153 @@
...
+
+import ehrdata as ed
+import matplotlib
+import matplotlib.pyplot as plt
+import pytest
+from ehrdata.core.constants import DEFAULT_TEM_LAYER_NAME
+
+import ehrapy as ep
+
+matplotlib.use("Agg")
```
Zethson: I wonder whether we should put this into the main conftest.py and apply it across all plots

ehrapy/plot/_timeseries.py
sueoglu: hmm then in that case the users should be able to either enter the row index of the observation or the obs name (like the subject id etc) that they want to plot, right?

ehrapy/plot/_timeseries.py
sueoglu: Sure, I added that as a key to specify because ehrdata_blobs creates the tem data frame with columns` timepoint` and `time_value `

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,186 @@
...
+    xlabel: str | None = None,
+    ylabel: str | None = None,
+    width: int | None = 600,
+    height: int | None = 400,
+    title: str | None = None,
+) -> hv.Overlay | hv.Layout:
+    """Plot variable time series either for an observation or for multiple observations from a 3D EHRData layer.
+
+    Selection logic:
+        - If obs_names is an int in [0, n_obs), use it as row index.
```
eroell: I suggest to use the same logic for all 3 axes, and no special case for int/row index for obs

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,186 @@
...
+import numpy as np
+import pandas as pd
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from ehrdata import EHRData
+
+
+def plot_timeseries(
```
eroell: See comment [here](https://github.com/theislab/ehrapy/pull/989#discussion_r2622243321), no `plot.plot_...`

just `timeseries`?

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,186 @@
...
+    tem_names: int | Sequence[int] | None = None,
+    layer: str = "tem_data",
+    overlay: bool = False,
+    xlabel: str | None = None,
+    ylabel: str | None = None,
+    width: int | None = 600,
+    height: int | None = 400,
+    title: str | None = None,
+) -> hv.Overlay | hv.Layout:
+    """Plot variable time series either for an observation or for multiple observations from a 3D EHRData layer.
```
eroell:
```suggestion
    """Plot time series from a 3D EHRData object.
```

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,186 @@
...
+        xlabel: The x-axis label text.
+        ylabel: The y-axis label text.
+        width: Plot width in pixels.
+        height: Plot height in pixels.
+        title: Set the title of the plot.
+
+    Returns:
+        HoloViews Overlay (if overlay=True) or Layout (if overlay=False) object representing the time series plot(s).
+
+    Examples:
```
eroell, line 51: examples should be copy-paste ready code: please add `import ehrapy as ep` and `import ehrdata as ed`

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,186 @@
...
+    >>> edata = ed.dt.ehrdata_blobs(n_variables=10, n_observations=5, base_timepoints=100)
+    >>> ep.pl.plot_timeseries(edata, obs_names=1, var_names=["feature_1", "feature_2], tem_names=range(50))
```
eroell: Missing `"` after `feature_2`

tests/_scripts/timeseries_plot_multiple_obs_expected .ipynb
```diff
@@ -0,0 +1,126 @@
+{
```
eroell: Are these scripts, and the images, still used for testing, or can they be discarded?

eroell (commented):
Mainly last cleanup-comments: I think this is a valid implementation improving ehrapy by now!

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,186 @@
...
+    ylabel: str | None = None,
+    width: int | None = 600,
+    height: int | None = 400,
+    title: str | None = None,
+) -> hv.Overlay | hv.Layout:
+    """Plot variable time series either for an observation or for multiple observations from a 3D EHRData layer.
+
+    Selection logic:
+        - If obs_names is an int in [0, n_obs), use it as row index.
+        - Otherwise, it should match a row name in edata.obs_names.
```
Zethson:
```suggestion
        - Otherwise, it should match a row name in `edata.obs_names`.
```

or a sphinx ref

ehrapy/plot/_timeseries.py
```diff
@@ -0,0 +1,186 @@
...
+    obs_ids: list[str | int]
+    if obs_names is None:
+        obs_ids = list(range(n_obs))
+    elif isinstance(obs_names, (str, int)):
+        obs_ids = [obs_names]
+    else:
+        obs_ids = list(obs_names)
+
+    if not obs_ids:
+        raise ValueError("obs_names is empty")
```
Zethson: Please ensure that you end all errors with a full stop.

ehrapy/plot/_timeseries.py
sueoglu: Got it! I’ll resolve all of the axes via their indices

Zethson (approved):
Thank you!

eroell (commented):
formating of code example in the documentaiton is off - when this is fixed, you can hit the merge button! 🥇 

image

Issue #892, opened 2025-04-27
New documentation and tutorials structure


Zethson:
### Description of feature

Our tutorials suffer from at least the following issues:

1. They are too long and verbose.
2. They do not suggest best practices or which of the many options to use.
3. They do not explain how to work with time series data
4. They are hard to navigate because of a weird sidebar structure and the sphinx gallery.

Therefore, I suggest that we:
1. Remove the sphinx gallery and just use text links. A very smart friend of mine told me years ago that this is the way to go.
2. Use lots of dropdowns to explain details but not make the tutorials seem too long.
3. Use lots of `see also` sections to refer to other parts of the documentation and especially the API. Instead of showing 5 survival analysis functions in one big tutorial, we should show off one and then refer to the API for the other options.
4. Use a more clear structure for the tutorials like:

What we really need is a complete ehrdata example dataset which has every slot populated. No matter how, even if we just add some random stuff to the physionet one. @eroell 

-10 minutes to ehrapy (anndata, ehrdata, scanpy, ehrapy) which replaces and merges the current introduction to ehrapy and MIMIC-II IAC if possible. Also introduces the readers and file formats.
- Working with time series data (lots of work needed to get this ready)
- Working with text data (medcat and ontology mapping)
- Survival analysis
- Trajectory inference
- Causal analysis (causal discovery (NA) & causal inference)
- Bias detection (merges with cohort tracking)
- Machine learning (simple model, deep learning & time series, dataloaders & sampling)
- Using ehrapy with large datasets

Issue #889, opened 2025-04-21
A pipeline for time series data


Zethson:
### Description of feature

There are several time series issues here already and this is yet another one. I'll try to consolidate them later. I'll edit this issue when I'm back with more details.

IMO it is important that we have a pipeline in mind that we implement. Ideally, we draft it using LLM code with example data to ensure that it gives some useful results. Therefore, I would kindly ask @eroell to prepare an artificial ehrdata object (sparse and dense) with random values that we can use to test all of our implementations.

Here are my suggestions:

### 1. **Preprocessing**
- **Data harmonization**: Standardize units, map codes (e.g., ICD, LOINC) to consistent ontology. ✅ 
- **Temporal alignment**: Resample or align all patients on common time axis (e.g., relative to admission, diagnosis). @eroell 
- **Missing data imputation**: Apply time-aware methods (e.g., interpolation, Kalman filter, BRITS). @eroell 
- **Normalization**: Per-feature z-scoring or quantile normalization, possibly per patient. @eroell 
- **Dimensionality reduction (optional)**: Aggregate or collapse rare events, reduce sparsity. We'll need to look into this. @eroell 

---

### 2. **Feature engineering**
- **Time-window aggregation**: Aggregate features into fixed-length windows (mean, slope, min, etc.). @eroell 
- **Trajectory encoding**: Create patient trajectories (sequences) or use sequence embeddings. This is LLM speak and I'll need to think about it. @Zethson 
- **Event encoding**: Binary or frequency counts for diagnoses/procedures per window. This is LLM speak and I'll need to think about it. @Zethson 
- **Derived features**: Compute derived metrics (e.g., delta from baseline, moving averages). @eroell 

---

### 3. **Quality control**
- **Patient filtering**: Remove patients with too much missing data or insufficient follow-up.
- **Feature filtering**: Drop features with low variance, extreme missingness, or redundancy.
- **Outlier detection**: Detect patients or time points with extreme values or patterns.

---

### 4. **Embedding / Dimensionality Reduction**
- **Per-patient summary embedding**: Aggregate each patient into a vector, then apply PCA, UMAP, or t-SNE. We already have some of that but we need to ensure that there's simple aggregation. @Zethson 
- **Time-aware embedding**: Use TSNE/UMAP on time-windowed representations or temporal VAEs/RNNs. @Zethson 
- **Manifold learning**: Build pseudotime or trajectory inference if clinically meaningful. @Zethson 

---

[...]

PR #1007, opened 2026-01-27
Longitudinal Notebook

### LINKED ISSUE #806: Tutorial featuring PyPTOS
### Description of feature

Draft good integration with https://github.com/WenjieDu/PyPOTS

TODOs
- [x] Copy [PyPOTS tutorial](https://ehrdata.readthedocs.io/en/latest/tutorials/tutorial_time_series_with_pypots.html) from ehrdata to ehrapy 
- [x] Expand the tutorial with a better introduction to PyPOTS, the used method, and a more interesting analysis of representation learning/classification

eroell:
**PR Checklist**

- [x] Referenced issue is linked. Fixes #806. Also a step regarding #889, #892

**Description of changes**
Adding a longer longitudinal notebook with ehrdata and ehrapy features and deep learning showcase

**Technical details**

**Additional context**


Zethson (commented):
This is nice! Lots of good stuff in there but there's many things that need improving.

 
1. Image 
the icon doesn't seem to show for me.
2. The title is too long. It doesn't look that great in the preview.
3. Image
what happened here? Are these progressbars that don't render correctly?
4. Image

this copy seems a bit weird?
5. Image

the message renders wrongly (spaces). Should be a quick fix.
6. Image

why is this outputting the axes? show=False? why? This also applies to other plots.
7. Everything here https://ehrapy--1007.org.readthedocs.build/en/1007/tutorials/notebooks/longitudinal_with_ehrapy.html#summary-and-next-steps and below is what you've kinda done on the EHRData repository in your new guides but we haven't done that in ehrapy before. We must keep the tutorials consistent. I think that most of this is noise and should not be there. Also, any next steps should properly link to the other tutorials so that they can easily be accessed. Not just raw links or named tutorials
8. Use `hide-output` cell tags for noisy cells whose output does not matter. e.g. random training output and other crap.
9. Input format: (X \in \mathbb{R}^{N \times T \times D}) with missing values allowed -> does not render
10. Image
this is pure noise and nothing of that sort should be shown here
11. The overview is way too long. 1-2 sentences at most and the TOC should show the rest. I'd just get rid of this completely because you already have an intro.
12. Can we use DWT instead of picking an arbitrary hour?

The most important feedback I have tho is that this is NOT a tutorial. This is more like a use case. It's too long for a tutorial although I think we can significantly shorten it. It partially also overlaps with existing tutorials. I'm 200% for a 3D tutorial but we need to discuss the overall tutorial design to ensure that all concepts flow into each other and we don't repeat things.

PR #1020, opened 2026-02-19
Last observed value carried forward (LOCF) Imputation

### LINKED ISSUE #481: Implement forward imputation for single patients
### Description of feature

By Karsten:

We should ensure that we only impute into the future with past values. Think that this must be on patient by patient case

agerardy:
**PR Checklist**

**Description of changes**

Implements Last Observation Carried Forward (LOCF) imputation for longitudinal (3D) EHR data, fixes #481.

For each patient and feature, missing values are replaced with the most recent non-missing value along the time axis. Missing values that occur before any observation for a given patient/feature are filled using a configurable fallback method (`"mean"`, `"median"`, `"most_frequent"`, `"bfill"`, or `None`).

**Technical details**

**Usage**
```py
import numpy as np
import ehrdata as ed
import ehrapy as ep
data = np.array(
  [
       [[1.0, np.nan, 3.0, np.nan], [np.nan, 2.0, np.nan, 4.0], [5.0, 6.0, 7.0, 8.0]],
       [[np.nan, np.nan, 3.0, np.nan], [1.0, np.nan, np.nan, np.nan], [np.nan, 2.0, np.nan, 4.0]],
  ]
)
edata = ed.EHRData(shape=(2, 3), layers={"tem_data": data})
ep.pp.locf_impute(edata, layer="tem_data")
edata.layers["tem_data"]
```

```
  array([[[1.        , 1.        , 3.        , 3.        ],
          [2.33, 2.        , 2.        , 4.        ],
          [5.        , 6.        , 7.        , 8.        ]],
         [[2.33, 2.33, 3.        , 3.        ],
          [1.        , 1.        , 1.        , 1.        ],
          [5.33, 2.        , 2.        , 4.        ]]])
```

### DIFF
--- a/.github/workflows/run_notebooks.yml
+++ b/.github/workflows/run_notebooks.yml
@@ -38,7 +38,7 @@ jobs:
               run: pip install uv
 
             - name: Install ehrapy and additional dependencies
-              run: uv pip install --system . cellrank nbconvert ipykernel graphviz pypots
+              run: uv pip install --system . 'cellrank[scvelo]' nbconvert ipykernel graphviz pypots
 
             - name: Run ${{ matrix.notebook }} Notebook
               run: jupyter nbconvert --to notebook --execute ${{ matrix.notebook }}


ehrapy/preprocessing/_imputation.py
```diff
@@ -637,6 +637,84 @@ def _warn_imputation_threshold(
     return var_name_to_pct
 
 
+@use_ehrdata(deprecated_after="1.0.0")
+@spinner("Performing LOCF impute")
+def locf_impute(
+    edata: EHRData | AnnData,
```
Zethson: I think this should just be EHRData only.

ehrapy/preprocessing/_imputation.py
```diff
@@ -637,6 +637,84 @@ def _warn_imputation_threshold(
...
+    var_names: Iterable[str] | None = None,
+    *,
+    layer: str | None = None,
+    fallback_method: Literal["mean"] = "mean",
```
Zethson: It should be trivial to also add median, no?

agerardy: sure, that's easy

ehrapy/preprocessing/_imputation.py
agerardy: I don't need the @use_ehrdata decorator then anymore, right?

Zethson: Yup! We can remove it.

agerardy:
I put in some comments explaining the new functions, not sure if they are necessary. maybe one comment explaining the purpose of the new file would be better?

ehrapy/preprocessing/_imputation.py
```diff
@@ -637,6 +639,84 @@ def _warn_imputation_threshold(
     return var_name_to_pct
 
 
+@spinner("Performing LOCF impute")
```
eroell: This is a fast imputation method, and does not require a spinner

ehrapy/preprocessing/_imputation.py
```diff
@@ -637,6 +639,84 @@ def _warn_imputation_threshold(
...
+    For each patient and feature, missing values are replaced with the most recent
+    non-missing value. Missing values that occur before any observation for a given
+    patient are filled using a fallback method.
+
+    Args:
+        edata: Central data object.
+        var_names: A list of column names to apply imputation on (if ``None``, impute all columns).
+        layer: The layer to impute. Must contain 3D data of shape ``(n_obs, n_vars, n_time)``.
+        fallback_method: Method for imputing values before the first observation per patient.
+                         ``'mean'`` fills with the per-feature mean, ``'median'`` fills with the
```
eroell, line 694: Lets also provide backwards fill as a fallback method

ehrapy/preprocessing/_imputation.py
```diff
@@ -637,6 +639,84 @@ def _warn_imputation_threshold(
...
+        If copy is True, a modified copy of the original data object with imputed data.
+        If copy is False, the original data object is modified in place, and None is returned.
+
+    Raises:
+        ValueError: If the data is not 3D or if an unsupported ``fallback_method`` is specified.
+
+    Examples:
+        >>> import ehrdata as ed
+        >>> import ehrapy as ep
+        >>> edata = ed.dt.mimic_2()
```
eroell: isn't mimic_2 2D?

ehrapy/preprocessing/_imputation.py
```diff
@@ -637,6 +639,84 @@ def _warn_imputation_threshold(
...
+        >>> ep.pp.locf_impute(edata, layer="tem_data")
```
eroell, line 720: this example doesn't work, please test an example (e.g. with physionet2012 challenge) and show a small output here, such as missing value count before and after imputation

ehrapy/_array_api.py
```diff
@@ -0,0 +1,105 @@
...
+        pad_shape = list(result.shape)
+        pad_shape[axis] = shift
+        pad = xp.full(pad_shape, float("nan"), dtype=result.dtype)
+        shifted = xp.concat([pad, result[_slice_along(axis, 0, n - shift)]], axis=axis)
+        result = xp.where(xp.isnan(result), shifted, result)
+
+    return result
+
+
+def nanmean_api(xp: ArrayNamespace, arr: ArrayAPICompliant, axes: int | tuple[int, ...]) -> ArrayAPICompliant:
```
eroell: why not just `array.nanmean()` or so in the main function?

ehrapy/_array_api.py
```diff
@@ -0,0 +1,105 @@
...
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ehrapy._types import ArrayAPICompliant, ArrayNamespace
+
+
+def ffill_along_axis(xp: ArrayNamespace, arr: ArrayAPICompliant, axis: int = -1) -> ArrayAPICompliant:
```
eroell: I don't see why we need a new `_array_api.py` file; this is just the helper that should be in `_imputation.py` in my opinion.

ehrapy/_array_api.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Args:
+        xp: Array namespace (e.g. from ``array_api_compat.array_namespace()``).
+        arr: Input array.
+        axis: Axis along which to forward-fill (default: last axis).
+
+    Returns:
+        Array with NaNs forward-filled along the specified axis.
+    """
+    import math
```
eroell: this looks complex, and the for loop is probably making this fairly slow

How about something like this?

```py
def xp_ffill_3d(x, xp):
    mask = xp.isnan(x)

    time_steps = xp.arange(x.shape[2], device=x.device)

    full_idx = xp.broadcast_to(time_steps[None, None, :], x.shape)
    

    valid_idx = xp.where(~mask, full_idx, xp.asarray(0, device=x.device))
    

    carried_idx = xp.cumulative_sum(valid_idx, axis=2) # Note: Logic varies by backend
```

You don't need to copy-paste, but there should be only `xp` operations, and I'd be surprised if extra `math` operations or slow `for-loops` are really required :)

ehrapy/_array_api.py
eroell: and there, be private, so `_ffill_along_axis`

ehrapy/_array_api.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Returns:
+        Array of shape ``(n_vars,)`` with the median per feature. Returns NaN
+        for features where all values are NaN.
+    """
+    n_obs, n_vars, n_time = arr.shape
+    arr_flat = xp.reshape(xp.permute_dims(arr, (1, 0, 2)), (n_vars, -1))
+
+    medians = []
+    for i in range(n_vars):
```
eroell: for loop really required? this function looks very complex for just a nanmedian..

ehrapy/preprocessing/_imputation.py
agerardy: i think so yes. forgot to test the final example yesterday, sry

ehrapy/_array_api.py
agerardy: I've tried several other methods but found no way without for loops. your suggestion would work with a cumulative_max, if I understand it correctly, but that doesnt exist in the API

agerardy: "The fundamental issue: forward fill is a scan operation where each output depends on the previous output. The only scan primitive in the Array API is cumulative_sum (additive), but forward fill needs a copy-last-valid scan (equivalent to cumulative max on indices). Without that primitive, a loop over the time axis is unavoidable."

ehrapy/_array_api.py
agerardy: it is very complex, yes. unfortunately I havent found an alternative within the array api without using a for loop, this is the vectorized version. the fact we need to handle the data differently for odd vs even counts, zero-count rows etc makes it kinda complex when you can't just use nanmedian

ehrapy/preprocessing/_imputation.py
```diff
@@ -637,6 +640,167 @@ def _warn_imputation_threshold(
...
+    xp = array_api_compat.array_namespace(X)
+    arr = xp.astype(X[:, var_indices, :], xp.float64)
+
+    arr = _ffill_along_time(xp, arr)
+
+    if fallback_method == "bfill":
+        arr = _bfill_along_time(xp, arr)
+    else:
+        n_vars = arr.shape[1]
+        if fallback_method == "mean":
```
eroell: We already have an impute function for mean and median in ehrapy, shouldn't we use these? 
If they need to be updated with array api usage, we should do it there (later), but not implement parallel versions here

ehrapy/_types.py
```diff
@@ -1,26 +1,65 @@
...
+    cls = type(name, (), {})
+    cls.__module__ = module
+    return cls
+
+
+if TYPE_CHECKING:
+    from cupy import ndarray as CupyArray
+    from dask.array.core import Array as DaskArray
+    from jax import Array as JaxArray
+    from ndonnx import Array as NdonnxArray
```
eroell: Where does ndonnx come from? It might be good to have, its just the first time I see it being used with ehrdata/ehrapy/scanpy/anndata which makes me curious

ehrapy/preprocessing/_imputation.py
```diff
@@ -637,6 +640,167 @@ def _warn_imputation_threshold(
     return var_name_to_pct
 
 
+def _ffill_along_time(xp: ArrayNamespace, arr: ArrayAPICompliant) -> ArrayAPICompliant:
```
eroell: I am in favor of using https://docs.xarray.dev/en/stable/generated/xarray.DataArray.ffill.html instead of a custom implementation for simplicit and performance; this for-loop's disadvantage and the many lines of code outweigh the array api benefits imo

PR #1026, opened 2026-03-04
fix example rendering of `explicit_impute`

sueoglu:

**Description of changes**
- docstring for `ep.pp.explicit_impute()` rendering should be rendering correctly right now
- `replacement` argument accepts float & generalized types


Zethson:
Thank you for the quick followup!

PR #1041, opened 2026-03-27
Add 3D support for knn_impute

### LINKED ISSUE #947: Longitudinal knn_impute

**3D**
If 3D allowed:

If 3D not allowed:
- [ ] only_2D decorator

**array-type**

**Encoding**
If encoded data is required

sueoglu:
fixes #947 
`ep.pp.knn_impute()` is extended to support 3D EHRdata.

**Description of changes**

- Updated _knn_impute to handle 3D array by first flattening it along the axis 0 and then passing it to the imputer 
- Added test for 3D numerical imputation using the existing edata_mini_3D_missing_values fixture

**Technical details**
For 3D array with shape n_obs, n_var, n_t:

1. The array is sliced to the requested variable and reshaped to (n_obs * n_t, n_var), each patient-timepoint becomes a row
2. The existing complete column logic runs on the flattened 2D array
3. After imputation the result is reshaped back to (n_obs, n_t, n_var) and transposed to (n_obs, n_var, n_t), then written back to only the requested variable positions in the layer


sueoglu:
Small note:

After our discussion, I used vertical stacking as reshaping strategy. I decided not to use the ehrapy decorator `_apply_over_time_axis` and to implement the reshaping logic inside `_knn_impute` because:
* its designed for functions that accept an array as input where `_knn_impute` accepts an EHRData object.
  
* for 3D input, `var_indices` and `numerical_indices` need to be remapped to the flattened array and the decorator has no mechanism to handle this.
  
* anchor column logic appends extra complete columns to `imputer_data_indices`, so `X_imputed` has more columns than the input `X`. With the decorator the reshape step would break since the decorator assumes that the shape of input is preserved.
  
Also here is an evaluation script comparing the two 3D stacking strategies (horizontal vs vertical), where RMSE and MAE of imputed positions in a dataset with 40% missingness are computed against the corresponding positions in the complete dataset to assess imputation quality.

```python3
import numpy as np
import ehrdata as ed
import ehrapy as ep
from typing import Iterable
from ehrdata.core.constants import FEATURE_TYPE_KEY, NUMERIC_TAG


def _knn_impute_with_mode(edata, var_names, n_neighbors, layer, temporal_mode):
    edata = edata.copy()

    from fknni import FastKNNImputer

    imputer = FastKNNImputer(n_neighbors=n_neighbors)

    if var_names is None:
        var_names = edata.var_names
    var_indices = edata.var_names.get_indexer(var_names).tolist()

    numerical_var_names = edata.var_names[edata.var[FEATURE_TYPE_KEY] == NUMERIC_TAG]
    numerical_indices = edata.var_names.get_indexer(numerical_var_names).tolist()

    X = edata.X if layer is None else edata.layers[layer]
    var_indices_original = var_indices
    is_3d = False

    if X.ndim == 3:
        is_3d = True
        n_obs, n_vars, n_t = X.shape
        if temporal_mode == "vertical":
            X = X[:, var_indices, :].astype("float64").transpose(0, 2, 1).reshape(n_obs * n_t, len(var_indices))
        else:  # horizontal
            X = X[:, var_indices, :].astype("float64").reshape(n_obs, len(var_indices) * n_t)
        numerical_indices = list(range(X.shape[1]))
        var_indices = numerical_indices

    complete_numerical_columns = np.array(numerical_indices)[
        ~np.isnan(X[:, numerical_indices]).any(axis=0)
    ].tolist()
    imputer_data_indices = var_indices + [i for i in complete_numerical_columns if i not in var_indices]
    imputer_x = X[:, imputer_data_indices].astype("float64")
    X_imputed = imputer.fit_transform(imputer_x)

    if is_3d:
        if temporal_mode == "vertical":
            X_imputed = X_imputed[:, :len(var_indices_original)].reshape(n_obs, n_t, len(var_indices_original)).transpose(0, 2, 1)
        else:
            X_imputed = X_imputed[:, :len(var_indices_original) * n_t].reshape(n_obs, len(var_indices_original), n_t)
        edata.layers[layer][:, var_indices_original, :] = X_imputed
    else:
        if layer is None:
            edata.X[:, imputer_data_indices] = X_imputed
        else:
            edata.layers[layer][:, imputer_data_indices] = X_imputed

    return edata


edata_complete = ed.dt.ehrdata_blobs(
    n_variables=10, missing_values=0.0, n_observations=100,
    base_timepoints=10, random_state=42, seasonality=True
)
truth = edata_complete.layers["tem_data"]

edata = ed.dt.ehrdata_blobs(
    n_variables=10, missing_values=0.4, n_observations=100,
    base_timepoints=10, random_state=42, seasonality=True
)
ed.infer_feature_types(edata) 
mask = np.isnan(edata.layers["tem_data"])

def evaluate(imputed, truth, mask, label):
    diff  = imputed[mask] - truth[mask]
    rmse  = np.sqrt(np.mean(diff**2))
    mae   = np.mean(np.abs(diff))
    scale = truth[mask].max() - truth[mask].min()
    nrmse = rmse / scale
    nmae  = mae  / scale
    print(f"{label:12s} RMSE: {rmse:.4f}, MAE: {mae:.4f}, NRMSE: {nrmse:.4f}, NMAE: {nmae:.4f}")


for mode in ["vertical", "horizontal"]:
    edata_imputed = _knn_impute_with_mode(edata, var_names=None, n_neighbors=5, layer="tem_data", temporal_mode=mode)
    evaluate(edata_imputed.layers["tem_data"], truth, mask, mode)
```

Output:
```

vertical     RMSE: 3.7324, MAE: 2.6651, NRMSE: 0.0949, NMAE: 0.0677
horizontal   RMSE: 2.1849, MAE: 1.5606, NRMSE: 0.0555, NMAE: 0.0397
```

tests/preprocessing/test_imputation.py
```diff
@@ -278,10 +278,43 @@ def test_simple_impute_invalid_strategy(impute_edata):
...
+    _base_check_imputation(
+        edata,
+        edata_imputed,
+        before_imputation_layer=DEFAULT_TEM_LAYER_NAME,
+        after_imputation_layer=DEFAULT_TEM_LAYER_NAME,
+    )
+
+
+def test_knn_impute_3d_var_names_subset(edata_mini_3D_missing_values):
+    edata = edata_mini_3D_missing_values.copy()
```
Zethson, line 306: Why are we creating a copy here?

ehrapy/preprocessing/_imputation.py
```diff
@@ -388,14 +392,37 @@ def _knn_impute(
             "var_names parameter or perform an encoding of your data."
         )
     X = edata.X if layer is None else edata.layers[layer]
+    var_indices_original = var_indices
+    is_3d = False
```
Zethson, line 409: Nit: This variable is only used once so we don't really need it, right? We can just check the dim of `X` again. 

I suggest that we rename `X` to `mtx` here because `X` is overloaded in our context.

ehrapy/preprocessing/_imputation.py
```diff
@@ -388,14 +392,37 @@ def _knn_impute(
...
+
+    # if input data is 3D flatten along axis 0 before giving it to the imputer: each timepoint becomes a row
```
Zethson:
```suggestion
    # if input data is 3D, flatten along axis 0 before giving it to the imputer: each timepoint becomes a row
```

just for readability.

ehrapy/preprocessing/_imputation.py
```diff
@@ -388,14 +392,37 @@ def _knn_impute(
...
+    if X.ndim == 3:
+        is_3d = True
+        n_obs, n_vars, n_t = X.shape
+        X = X[:, var_indices, :].astype("float64").transpose(0, 2, 1).reshape(n_obs * n_t, len(var_indices))
```
Zethson: Do we need a copy here? Like does this affect the original input?

ehrapy/preprocessing/_imputation.py
```diff
@@ -388,14 +392,37 @@ def _knn_impute(
...
+
+    # complete columns to be used as anchors
     complete_numerical_columns = np.array(numerical_indices)[~np.isnan(X[:, numerical_indices]).any(axis=0)].tolist()
-    imputer_data_indices = var_indices + [i for i in complete_numerical_columns if i not in var_indices]
-    imputer_x = X[::, imputer_data_indices].astype("float64")
 
-    if layer is None:
-        edata.X[::, imputer_data_indices] = imputer.fit_transform(imputer_x)
+    imputer_data_indices = var_indices + [
+        i for i in complete_numerical_columns if i not in var_indices
```
Zethson: Not a fan of `i` as variable name. Please use something more specific.

ehrapy/preprocessing/_imputation.py
```diff
@@ -388,14 +392,37 @@ def _knn_impute(
...
+    ]  # columns to impute
+    imputer_x = X[:, imputer_data_indices].astype("float64")
```
Zethson: Can we reuse the input dtype instead of casting everything randomly to float64? What if the input was float32?

tests/preprocessing/test_imputation.py
```diff
@@ -278,10 +278,43 @@ def test_simple_impute_invalid_strategy(impute_edata):
         simple_impute(impute_edata, strategy="invalid_strategy", copy=True)  # type: ignore
 
 
-def test_knn_impute_3D_edata(edata_blob_small):
-    knn_impute(edata_blob_small, layer="layer_2")
-    with pytest.raises(ValueError, match=r"only supports 2D data"):
-        knn_impute(edata_blob_small, layer=DEFAULT_TEM_LAYER_NAME)
+def test_knn_impute_3d_numerical(edata_mini_3D_missing_values):
+    edata = edata_mini_3D_missing_values[:, :4].copy()
```
Zethson: The `4` is pretty magic here. Why is it needed and is there a better way?

tests/preprocessing/test_imputation.py
```diff
@@ -278,10 +278,43 @@ def test_simple_impute_invalid_strategy(impute_edata):
...
+    _base_check_imputation(
+        edata,
+        edata_imputed,
+        before_imputation_layer=DEFAULT_TEM_LAYER_NAME,
+        after_imputation_layer=DEFAULT_TEM_LAYER_NAME,
+    )
+
+
+def test_knn_impute_3d_scikit_backend(edata_mini_3D_missing_values):
+    edata = edata_mini_3D_missing_values[:, :4].copy()
```
Zethson: See above.

Zethson (commented):
Thank you!
I think it looks good already but I have some nitpicky comments.

tests/preprocessing/test_imputation.py
sueoglu: because the first 4 variables in that fixture (`edata_mini_3D_missing_values`) are numerical and the last two are categorical

Zethson: Nit: Right but how about we find a way to only generate this fixture with numerical? Like with an explicit parameter that humans can read instead of  a magical number `4`.

sueoglu: oh yes sure, I didnt want to touch the fixture as there are other tests using that but I can modify it

ehrapy/preprocessing/_imputation.py
```diff
@@ -324,6 +323,11 @@ def knn_impute(
     if copy:
         edata = edata.copy()
 
+    if edata.X is None and layer is None:  # if edata is 3D
```
eroell, line 339: I wonder what the failure case is which makes this error necessary?

We don't have such a check in the other imputation methods, either

ehrapy/preprocessing/_imputation.py
sueoglu: yes but the dimension of mtx is changed as it gets flattened. another option could be saving the original dimension of mtx at the beginning and then checking if its 3

Zethson: I see or we just keep it as you have it then. No big deal.

ehrapy/preprocessing/_imputation.py
sueoglu: I think no copy is needed because with .astype(), mtx gets reassigned to a fresh array and the original layer is not modified

Zethson: Potentially unless, we don't use `astype()` - see other comment. Thank you though!

eroell: The test `_base_check_imputation` should check that the original data is not affected :)

eroell (commented):
I have not much more add, all the earlier discussion points are well described in your PR comments and I think this is a nice cleanup of this imputation function

ehrapy/preprocessing/_imputation.py
sueoglu: this catches the case where someone passes a 3D EHRData without X without specifying a layer. Wouldnt it throw a messy error if we dont catch this failure case?

ehrapy/preprocessing/_imputation.py
sueoglu: @eroell you mean a check between original layer before imputation and original layer after imputation right? I think _base_check_imputation currently checks that 

```py3
   if not _are_ndarrays_equal(layer_before[imputed_non_nan_mask], layer_after[imputed_non_nan_mask]):
        raise AssertionError("Non-NaN values in imputed columns were modified.")
```

ehrapy/preprocessing/_imputation.py
eroell: It's cleaner, I see the point of having it

eroell (approved):
Cool!

Can you in the documentation

- A quick preview where dataset with missing values, then the imputation, and then the missing values gone are shown? Either ehrdata blobs or one of the pre-loaded datasets ideally not requiring any other import than ehrdata and ehrapy
- Add 1-2 sentences on how this works in the 2D vs 3D case?

No need to re-request review afterwards, when this is done you can merge!

PR #1045, opened 2026-04-05
feat(preprocessing): add missing_data_mask function

### LINKED ISSUE #900: Feature request: data missingess indicators
### Description of feature

Quite straightforward but nice to have as API call as this is a high-volume operation, e.g.
`ed.dt.missingness_indicator(edata)` with e.g. default new layer `missingness_indicator` or `mask` or `missingness_mask` or so being created.

haoyu-haoyu:
## Summary

Add `ep.pp.missing_data_mask(edata)` to create boolean masks indicating missing values in EHR data matrices.

## Implementation

```python
ep.pp.missing_data_mask(
    edata,
    layer=None,           # Use edata.X or a specific layer
    mask_values=None,     # Additional sentinel values to treat as missing
    key_added="missing_data_mask",
    copy=False,
)
```

**Stores result in** `edata.layers[key_added]` as a dense boolean numpy array.

### Supported array types

| Type | Handler |
|------|---------|
| `np.ndarray` | `np.isnan()` + `np.isin()` |
| `sp.csr_array` / `sp.csc_array` | `.toarray()` → numpy path |
| `DaskArray` | `da.isnan().compute()` + `da.isin().compute()` |

### Files

| File | Change |
|------|--------|
| `ehrapy/preprocessing/_missing_data.py` | New: function + singledispatch handlers |
| `ehrapy/preprocessing/__init__.py` | Export (alphabetically after `_imputation`) |
| `tests/preprocessing/test_missing_data.py` | 10 tests covering all array types + edge cases |

### Pattern compliance

- `@use_ehrdata(deprecated_after="1.0.0")` decorator
- `singledispatch` for array type dispatch (dispatches on `mtx`, not `mask`)
- Keyword-only parameters after `edata`
- `copy` returns modified copy or `None`
- Google-style `Args:` docstrings

Closes #900

### DIFF
--- a/ehrapy/preprocessing/_missing_data.py
+++ b/ehrapy/preprocessing/_missing_data.py
@@ -1,0 +1,101 @@
+from __future__ import annotations
+
+from collections.abc import Iterable
+from functools import singledispatch
+from typing import TYPE_CHECKING
+
+import numpy as np
+import scipy.sparse as sp
+
+from ehrapy._compat import (
+    DaskArray,
+    _raise_array_type_not_implemented,
+    use_ehrdata,
+)
+
+if TYPE_CHECKING:
+    from anndata import AnnData
+    from ehrdata import EHRData
+
+
+@use_ehrdata(deprecated_after="1.0.0")
+def missing_data_mask(
+    edata: EHRData | AnnData,
+    *,
+    layer: str | None = None,
+    mask_values: Iterable[float | int] | None = None,
+    key_added: str = "missing_data_mask",
+    copy: bool = False,
+) -> EHRData | AnnData | None:
+    """Create a boolean mask indicating missing values in the data matrix.
+
+    By default marks ``NaN`` values as missing.  Optionally also marks
+    user-specified sentinel values (e.g. ``-1``, ``0``, ``999``) as missing.
+
+    The result is stored in ``edata.layers[key_added]``.
+
+    Args:
+        edata: Central data object.
+        layer: Layer to use instead of ``edata.X``.
+        mask_values: Additional values to treat as missing besides ``NaN``.
+        key_added: Key under which the boolean mask is stored in
+            ``edata.layers``.
+        copy: If ``True``, return a modified copy; otherwise modify in place.
+
+    Returns:
+        ``None`` if ``copy=False``, otherwise the updated data object.
+
+    Examples:
+        >>> import ehrdata as ed
+        >>> import ehrapy as ep
+        >>> edata = ed.dt.mimic_2()
+        >>> ep.pp.missing_data_mask(edata)
+    """
+    if copy:
+        edata = edata.copy()
+
+    X = edata.X if layer is None else edata.layers[layer]
+
+    mask = _compute_nan_mask(X)
+
+    if mask_values is not None:
+        values = list(mask_values)
+        mask = _apply_sentinel_mask(mask, X, values)
+
+    edata.layers[key_added] = mask
+
+    return edata if copy else None
+
+
+@singledispatch
+def _compute_nan_mask(mtx):
+    _raise_array_type_not_implemented(_compute_nan_mask, type(mtx))
+
+
+@_compute_nan_mask.register(np.ndarray)
+def _(mtx: np.ndarray) -> np.ndarray:
+    return np.isnan(mtx)
+
+
+@_compute_nan_mask.register(DaskArray)
+def _(mtx: DaskArray) -> np.ndarray:
+    import dask.array as da
+
+    return da.isnan(mtx).compute()
+
+
+@singledispatch
+def _apply_sentinel_mask(mask, mtx, values):
+    _raise_array_type_not_implemented(_apply_sentinel_mask, type(mtx))
+
+
+@_apply_sentinel_mask.register(np.ndarray)
+def _(mask: np.ndarray, mtx: np.ndarray, values: list) -> np.ndarray:
+    return mask | np.isin(mtx, values)
+
+
+@_apply_sentinel_mask.register(DaskArray)
+def _(mask: np.ndarray, mtx: DaskArray, values: list) -> np.ndarray:
+    import dask.array as da
+
+    return mask | da.isin(mtx, values).compute()
--- a/ehrapy/preprocessing/_missing_data.py
+++ b/ehrapy/preprocessing/_missing_data.py
@@ -54,6 +54,10 @@ def missing_data_mask(
         >>> edata = ed.dt.mimic_2()
         >>> ep.pp.missing_data_mask(edata)
         >>> edata
+        EHRData object with n_obs × n_vars × n_t = 1776 × 46 × 1
+            layers: 'missing_data_mask'
+            shape of .X: (1776, 46)
+            shape of .missing_data_mask: (1776, 46)
     """
     if copy:
         edata = edata.copy()
--- a/ehrapy/preprocessing/_missing_data.py
+++ b/ehrapy/preprocessing/_missing_data.py
@@ -79,6 +79,13 @@ def _compute_nan_mask(mtx):
 
 @_compute_nan_mask.register(np.ndarray)
 def _(mtx: np.ndarray) -> np.ndarray:
+    if mtx.dtype.kind == "O":
+        # `np.isnan` only accepts numeric dtypes, so an object-typed
+        # array (e.g. EHR data with categorical and numeric columns
+        # mixed) would raise. `pd.isna` is the dtype-agnostic check.
+        import pandas as pd
+
+        return pd.isna(mtx)
     return np.isnan(mtx)
 
 
--- a/ehrapy/preprocessing/_missing_data.py
+++ b/ehrapy/preprocessing/_missing_data.py
@@ -33,12 +33,15 @@ def missing_data_mask(
     By default marks ``NaN`` values as missing.
     Optionally also marks user-specified sentinel values (e.g. ``-1``, ``0``, ``999``) as missing.
 
-    The result is stored in ``edata.layers[key_added]``.
+    The result is stored in ``edata.layers[key_added]`` and preserves the
+    array backend of the source matrix: dense in / dense out, sparse in /
+    sparse out, dask in / dask out.
 
     Args:
         edata: Central data object.
         layer: Layer to use instead of ``edata.X``.
         mask_values: Additional values to treat as missing besides ``NaN``.
+            Not supported on sparse arrays — densify first or use a dense layer.
         key_added: Key under which the boolean mask is stored in ``edata.layers``.
         copy: If ``True``, return a modified copy; otherwise modify in place.
 
@@ -50,6 +53,7 @@ def missing_data_mask(
         >>> import ehrapy as ep
         >>> edata = ed.dt.mimic_2()
         >>> ep.pp.missing_data_mask(edata)
+        >>> edata
     """
     if copy:
         edata = edata.copy()
@@ -60,7 +64,8 @@ def missing_data_mask(
 
     if mask_values is not None:
         values = list(mask_values)
-        mask = _apply_sentinel_mask(X, mask, values)
+        if values:
+            mask = _apply_sentinel_mask(X, mask, values)
 
     edata.layers[key_added] = mask
 
@@ -79,19 +84,22 @@ def _(mtx: np.ndarray) -> np.ndarray:
 
 @_compute_nan_mask.register(sp.csr_array)
 @_compute_nan_mask.register(sp.csc_array)
-def _(mtx: sp.csr_array | sp.csc_array) -> np.ndarray:
-    return np.isnan(mtx.toarray())
+def _(mtx: sp.csr_array | sp.csc_array) -> sp.csr_array | sp.csc_array:
+    # Sparse formats use implicit zeros, so a NaN can only appear among the
+    # explicitly stored data entries.  Reuse the input's indices/indptr and
+    # replace ``data`` with a boolean isnan view, so the result preserves
+    # the sparse backend without densifying.
+    nan_data = np.isnan(mtx.data)
+    return type(mtx)((nan_data, mtx.indices.copy(), mtx.indptr.copy()), shape=mtx.shape)
 
 
 @_compute_nan_mask.register(DaskArray)
-def _(mtx: DaskArray) -> np.ndarray:
+def _(mtx: DaskArray) -> DaskArray:
     import dask.array as da
 
-    return da.isnan(mtx).compute()
+    return da.isnan(mtx)
 
 
-# singledispatch dispatches on the FIRST argument, so dispatch on mtx (the data matrix), not mask.
-# This ensures sparse and Dask paths are reached.
 @singledispatch
 def _apply_sentinel_mask(mtx, mask, values):
     _raise_array_type_not_implemented(_apply_sentinel_mask, type(mtx))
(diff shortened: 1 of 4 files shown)


ehrapy/preprocessing/_missing_data.py
```diff
@@ -0,0 +1,115 @@
...
+
+@_compute_nan_mask.register(np.ndarray)
+def _(mtx: np.ndarray) -> np.ndarray:
+    return np.isnan(mtx)
+
+
+@_compute_nan_mask.register(sp.csr_array)
+@_compute_nan_mask.register(sp.csc_array)
+def _(mtx: sp.csr_array | sp.csc_array) -> np.ndarray:
+    return np.isnan(mtx.toarray())
```
Zethson: This is bad. This just densifies. It also doesn't use fast-array-utils. Can you come up with a pure sparse implementation?

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+import pandas as pd
+import pytest
+
+import ehrapy as ep
+from ehrapy._types import ARRAY_TYPES_NONNUMERIC
+from tests.conftest import as_dense_dask_array
+
+
+@pytest.fixture
+def edata_with_nan():
```
Zethson: Could you please use blobs or something instead of generating another fixture?

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+    X = np.array([[1.0, np.nan, 3.0], [np.nan, 5.0, np.nan]], dtype=np.float64)
+    return ed.EHRData(
+        X=X,
+        obs=pd.DataFrame({"id": ["a", "b"]}),
+        var=pd.DataFrame(index=["v1", "v2", "v3"]),
+    )
+
+
+class TestMissingDataMaskDefault:
+    """Test default NaN masking."""
```
Zethson: Useless comment
```suggestion
```

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+            var=pd.DataFrame(index=["v1", "v2"]),
+        )
+
+        ep.pp.missing_data_mask(edata)
+
+        assert np.all(~edata.layers["missing_data_mask"])
+
+
+class TestMissingDataMaskCustomValues:
+    """Test custom mask_values parameter."""
```
Zethson:
```suggestion
```

Useless

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+
+        ep.pp.missing_data_mask(edata, mask_values=[999, -1])
+
+        expected = np.array([[True, False], [True, False]])
+        assert np.array_equal(edata.layers["missing_data_mask"], expected)
+
+
+class TestMissingDataMaskLayer:
+    """Test layer parameter."""
```
Zethson:
```suggestion
```

Just noise

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+
+        ep.pp.missing_data_mask(edata, layer="raw")
+
+        expected = np.array([[True, False], [False, True]])
+        assert np.array_equal(edata.layers["missing_data_mask"], expected)
+
+
+class TestMissingDataMaskCopy:
+    """Test copy parameter."""
```
Zethson:
```suggestion
```

Just noise

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+        result = ep.pp.missing_data_mask(edata_with_nan, copy=True)
+
+        assert result is not None
+        assert result is not edata_with_nan
+        assert "missing_data_mask" in result.layers
+        assert "missing_data_mask" not in edata_with_nan.layers
+
+
+class TestMissingDataMaskKeyAdded:
+    """Test key_added parameter."""
```
Zethson:
```suggestion
```

Just noise

Zethson (changes requested):
Thanks! Looks a bit AI generated... Could you please disclose that?

1. I wonder whether the complexity of this is too low and doesn't warrant its own public function. Can you justify why this should be its own function? @eroell WDYT?
2. This function must also be exposed in the docs.
3. We're not really using test classes. Could you please remove them?

haoyu-haoyu (commented):
Thanks for taking a look!

And yeah fair enough, I did use an AI assistant to help with the initial drafting. Should've mentioned that upfront, my bad.

Re the complexity question, I hear you. It's not a huge amount of logic on its own, but my reasoning was that EHR datasets often encode missingness as sentinel values (-1, 0, 999 etc.) rather than NaN, so this handles both cases and stores a consistent mask in `.layers` that downstream imputation can reference. It also dispatches across dense/sparse/dask. But totally understand if you and @eroell think it fits better as an internal helper or part of an existing function. Happy either way, just let me know the preferred direction.

For the rest: the sparse `.toarray()` call is bad, will rework with `fast-array-utils` to keep it sparse. Will also add docs, drop the test classes in favor of plain functions, switch to existing fixtures/blobs, and clean up the noisy comments.

I'll hold off on pushing changes until you guys decide on the scope question so I'm not doing work twice.

Zethson:
> but my reasoning was that EHR datasets often encode missingness as sentinel values (-1, 0, 999 etc.) rather than NaN, so this handles both cases and stores a consistent mask in .layers that downstream imputation can reference. It also dispatches across dense/sparse/dask. 

These are good reasons. Let's see what @eroell says.

Thank you! I'm happy to see you around and contributing.

ehrapy/preprocessing/_missing_data.py
```diff
@@ -0,0 +1,115 @@
...
+        copy: If ``True``, return a modified copy; otherwise modify in place.
+
+    Returns:
+        ``None`` if ``copy=False``, otherwise the updated data object.
+
+    Examples:
+        >>> import ehrdata as ed
+        >>> import ehrapy as ep
+        >>> edata = ed.dt.mimic_2()
+        >>> ep.pp.missing_data_mask(edata)
```
eroell, line 55: Could you add here
```py
edata
```
And then the string representation? This shows nicely that the new layer has been created.

We are not having it everywhere yet unfortunately, but aim to provide a quick preview for what each function does

ehrapy/preprocessing/_missing_data.py
eroell: Indeed, we'd want to try to avoid `toarray()` except for imputation methods that densify an array by virtue of their function.

If you can come up with a sparse implementation, this would be great! If not, no worries, but then rather raise a not-implemented error than densifying "under the hood" where users cannot see it

ehrapy/preprocessing/_missing_data.py
```diff
@@ -0,0 +1,115 @@
...
+
+@_apply_sentinel_mask.register(np.ndarray)
+def _(mtx: np.ndarray, mask: np.ndarray, values: list) -> np.ndarray:
+    return mask | np.isin(mtx, values)
+
+
+@_apply_sentinel_mask.register(sp.csr_array)
+@_apply_sentinel_mask.register(sp.csc_array)
+def _(mtx: sp.csr_array | sp.csc_array, mask: np.ndarray, values: list) -> np.ndarray:
+    return mask | np.isin(mtx.toarray(), values)
```
eroell: same here about not using `toarray()`

ehrapy/preprocessing/_missing_data.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@_compute_nan_mask.register(DaskArray)
+def _(mtx: DaskArray) -> np.ndarray:
+    import dask.array as da
+
+    return da.isnan(mtx).compute()
+
+
+# singledispatch dispatches on the FIRST argument, so dispatch on mtx (the data matrix), not mask.
```
eroell:
```suggestion
```

this rather small function is self-explanatory

ehrapy/preprocessing/_missing_data.py
```diff
@@ -0,0 +1,115 @@
...
+# This ensures sparse and Dask paths are reached.
```
eroell:
```suggestion
```

tests/preprocessing/test_missing_data.py
eroell: Yes, there's multiple fixtures already that have some missing data generated

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+def edata_with_nan():
+    X = np.array([[1.0, np.nan, 3.0], [np.nan, 5.0, np.nan]], dtype=np.float64)
+    return ed.EHRData(
+        X=X,
+        obs=pd.DataFrame({"id": ["a", "b"]}),
+        var=pd.DataFrame(index=["v1", "v2", "v3"]),
+    )
+
+
+class TestMissingDataMaskDefault:
```
eroell: No test classes in ehrapy, simple functions are sufficient

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+            var=pd.DataFrame(index=["v1", "v2"]),
+        )
+
+        ep.pp.missing_data_mask(edata, mask_values=[999, -1])
+
+        expected = np.array([[True, False], [True, False]])
+        assert np.array_equal(edata.layers["missing_data_mask"], expected)
+
+
+class TestMissingDataMaskLayer:
```
eroell: also here no test class required, the functions are fine stand-alone

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+        X=X,
+        obs=pd.DataFrame({"id": ["a", "b"]}),
+        var=pd.DataFrame(index=["v1", "v2", "v3"]),
+    )
+
+
+class TestMissingDataMaskDefault:
+    """Test default NaN masking."""
+
+    @pytest.mark.parametrize("array_type", ARRAY_TYPES_NONNUMERIC)
```
eroell: This is already nice that you parametrize! Could you also verify that if the `edata.X` is a dask array, that it remains a dask array after ep.pp.missing_data_mask? and that the output is also a dask array, preferrably?

We have this testing pattern for checking that data from dask is not accidentally becoming a numpy array in e.g. the normalization functions

eroell (commented):
Thank you very much, this is an impressive first PR! 🙌

haoyu-haoyu:
Thanks @eroell for the thorough pass! All review comments addressed in 2d23b3c — summary below.

**`ehrapy/preprocessing/_missing_data.py`**
- Pure sparse NaN-mask path: reuse the input's `indices`/`indptr` with `isnan(data)` so CSR/CSC in → CSR/CSC bool out, no `.toarray()`.
- Dask path no longer eagerly `.compute()`s — `da.isnan` / `da.isin` keep the result a lazy `DaskArray`.
- For sparse + non-empty `mask_values`, raise `NotImplementedError` with a clear message rather than densify silently, per your guidance ("rather raise a not-implemented error than densifying under the hood"). Empty `mask_values` is treated as a no-op for every backend.
- Added `>>> edata` to the docstring example so the layer-creation effect is visible.
- Removed the singledispatch comment block.

**`tests/preprocessing/test_missing_data.py`**
- Switched to the existing `missing_values_edata` fixture from `conftest.py`.
- Removed all `Test*` classes — flat functions per ehrapy convention.
- Stripped the noise docstrings.
- Added `test_missing_data_mask_preserves_dask` that asserts the output layer is a `da.Array` when input is.
- Added CSR/CSC parametrized test for sparse preservation, bool dtype, and exact contents.
- Added empty-`mask_values` regression test on a sparse matrix.

**`docs/api/preprocessing_index.md`** — `preprocessing.missing_data_mask` now listed under Imputation.

15/15 local tests pass. Ready for re-review.

ehrapy/preprocessing/_missing_data.py
eroell: I tried the example, it currently fails:

```
TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

Also, it would be great to add the string representation I mentioned after the doc example. This is the string which is printed if you do e.g. print(edata)

tests/preprocessing/test_missing_data.py
```diff
@@ -0,0 +1,170 @@
...
+    edata.X = array_type(edata.X)
+
+    ep.pp.missing_data_mask(edata)
+
+    expected = np.array([[False, True, False], [True, True, False]])
+    assert "missing_data_mask" in edata.layers
+    assert np.array_equal(asarray(edata.layers["missing_data_mask"]), expected)
+
+
+def test_missing_data_mask_preserves_dask(missing_values_edata):
```
eroell, line 27: Thank you, please also add a check that the original array remains a dask array: currently, only the generated missing_data_mask is tested

eroell (commented):
Thank you! - some minor things left.

Very elegant solution for the sparse case.

haoyu-haoyu:
Thanks @eroell — both points addressed in e543922:

1. **Docstring example failed with TypeError on `mimic_2()`** — the dataset returns an object-dtype array (mixed numeric/categorical), and `np.isnan` only accepts numeric dtypes. The numpy branch of `_compute_nan_mask` now checks `dtype.kind == "O"` and falls back to `pd.isna`, which handles `None`/`np.nan`/`pd.NA` uniformly. The example now runs end-to-end, and the `>>> edata` line produces the post-mutation string repr that shows the new layer.

2. **Original array remains dask** — extended `test_missing_data_mask_preserves_dask` to also assert `isinstance(edata.X, da.Array)` after the call. Added `test_missing_data_mask_object_dtype` to cover the new `pd.isna` fallback.

16/16 tests pass locally. Ready for another look.

ehrapy/preprocessing/_missing_data.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Returns:
+        ``None`` if ``copy=False``, otherwise the updated data object.
+
+    Examples:
+        >>> import ehrdata as ed
+        >>> import ehrapy as ep
+        >>> edata = ed.dt.mimic_2()
+        >>> ep.pp.missing_data_mask(edata)
+        >>> edata
```
eroell, line 56: Please add the string representation of `edata` at the end of this example, we do so e.g. [here](https://github.com/theislab/ehrdata/blob/72d96325f7ebf4c0b8603c5e7954f1b1fed7b359/src/ehrdata/dt/datasets.py#L661)

As you're using a coding agent to speed implementation up, please give it a quick human look to ensure this is addressed, as the agent has seemed to struggle picking it up in the last two review rounds

eroell (commented):
This feature is now robust enough to merge it after the minor update in the docs and get this on!

haoyu-haoyu:
Thanks @eroell, and sorry for making you ask three times — that's on me. Pushed 2537359:

```python
    Examples:
        >>> import ehrdata as ed
        >>> import ehrapy as ep
        >>> edata = ed.dt.mimic_2()
        >>> ep.pp.missing_data_mask(edata)
        >>> edata
        EHRData object with n_obs × n_vars × n_t = 1776 × 46 × 1
            layers: 'missing_data_mask'
            shape of .X: (1776, 46)
            shape of .missing_data_mask: (1776, 46)
```

Captured by actually running `ed.dt.mimic_2()` → `ep.pp.missing_data_mask(edata)` → `print(edata)` locally and string-matching the output byte for byte, so the docstring is authoritative rather than guessed.

eroell:
No big deal, in any case: thank you for your contribution!

PR #1052, opened 2026-04-16
Add 3D support to `miss_forest_impute`

sueoglu:
fixes #948 
Extends `miss_forest_impute` to handle 3D `EHRData` inputs

- Removed `@function_2D_only()` decorator
-  If the input is 3D: flattens the layer to `(n_obs * n_t, n_vars)` (flattening along axis 0) before imputation and reshapes back to `(n_obs, n_vars, n_t)` afterwards using `_apply_over_time_axis`
- Added guard raising `ValueError` when input is 3D but no layer is specified
- Added tests for 3D imputation
- Raising NotImplemented error for unsupported array types (sparse and dask)

Comparison of `miss_forest_impute` and `simple_impute `with mean strategy using physionet2012 dataset with a missingness of 10%: 

```python3
import ehrapy as ep
import ehrdata as ed
import numpy as np

edata_full = ed.dt.physionet2012(layer="tem_data")
edata = edata_full[:200]

full_layer = edata.layers["tem_data"].copy()

rng = np.random.default_rng(42)
non_nan_mask = ~np.isnan(full_layer)
mask_indices = np.where(non_nan_mask)
n_to_mask = int(0.1 * len(mask_indices[0]))
chosen = rng.choice(len(mask_indices[0]), size=n_to_mask, replace=False)

masked_layer = full_layer.copy()
masked_layer[mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]] = np.nan
edata.layers["tem_data"] = masked_layer

edata_mf = edata.copy()
edata_simple = edata.copy()

ep.pp.simple_impute(edata_simple, layer="tem_data", strategy="mean")
ep.pp.miss_forest_impute(edata_mf, layer="tem_data")

true_vals = full_layer[mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]]
mf_vals = edata_mf.layers["tem_data"][mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]]
simple_vals = edata_simple.layers["tem_data"][mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]]

print("True mean:", np.nanmean(true_vals))
print("Miss forest mean:", np.nanmean(mf_vals), "diff:", abs(np.nanmean(mf_vals) - np.nanmean(true_vals)))
print("Simple mean:", np.nanmean(simple_vals), "diff:", abs(np.nanmean(simple_vals) - np.nanmean(true_vals)))

[...]

PR #1055, opened 2026-04-21
3D support for `mice_forest_impute`

### LINKED ISSUE #949: Longitudinal mice_forest_impute

**3D**
If 3D allowed:

If 3D not allowed:
- [ ] only_2D decorator

**array-type**

**Encoding**
If encoded data is required

sueoglu:
fixes #949 
Extends `mice_forest_impute` to handle 3D inputs, following the same flatten/unflatten pattern established for `knn_impute`

- Raising ValueError when input is 3D but no layer is specified
- Added 3D support: flattening the layer to `(n_obs * n_t, n_vars)` before building the DataFrame and reshapes back to `(n_obs, n_vars, n_t)` after imputation
- Added tests for 3D 

*Technical notes*

The 3D path is asymmetric with the 2D:
- 2D: keeps the full matrix, uses `column_indices = var_indices` to select columns to impute, writes back the full DataFrame
- 3D: pre slices to `var_indices` before flattening, uses `column_indices = list(range(len(var_indices)))`, writes back only to `var_indices` positions

This is necessary for two reasons I observed when trying to making it symmetric:
1. Flattening the full 3D matrix with `.astype()` crashes on categorical columns (object dtype arrays with string values were not able to be cast to float)
2. Without explicit float casting, object dtype arrays cause NaN predictions in miceforest's internal KD-tree, raising `ValueError: 'x' must be finite`

Comparison of `mice_forest_impute `and `simple_impute `with mean strategy using physionet2012 dataset with a missingness of 10%:

```python3
import ehrapy as ep
import ehrdata as ed
import numpy as np

edata_full = ed.dt.physionet2012(layer="tem_data")
ed.infer_feature_types(edata_full, layer="tem_data")
edata = edata_full[:200]

full_layer = edata.layers["tem_data"].copy()

non_nan_mask = ~np.isnan(full_layer)
mask_indices = np.where(non_nan_mask)
n_to_mask = int(0.1 * len(mask_indices[0]))
chosen = rng.choice(len(mask_indices[0]), size=n_to_mask, replace=False)

masked_layer = full_layer.copy()
masked_layer[mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]] = np.nan
edata.layers["tem_data"] = masked_layer

edata_mf = edata.copy()
edata_simple = edata.copy()

ep.pp.simple_impute(edata_simple, layer="tem_data", strategy="mean")
ep.pp.mice_forest_impute(edata_mf, layer="tem_data")

true_vals = full_layer[mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]]
mf_vals = edata_mf.layers["tem_data"][mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]]
simple_vals = edata_simple.layers["tem_data"][mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]]

print("True mean:", np.nanmean(true_vals))
print("Mice forest mean:", np.nanmean(mf_vals), "diff:", abs(np.nanmean(mf_vals) - np.nanmean(true_vals)))
print("Simple mean:", np.nanmean(simple_vals), "diff:", abs(np.nanmean(simple_vals) - np.nanmean(true_vals)))

print("True std:", np.nanstd(true_vals))
print("Mice forest std:", np.nanstd(mf_vals), "diff:", abs(np.nanstd(mf_vals) - np.nanstd(true_vals)))
print("Simple std:", np.nanstd(simple_vals), "diff:", abs(np.nanstd(simple_vals) - np.nanstd(true_vals)))
```

Output:
 
```
True mean: 70.79201158141458
Mice forest mean: 71.61526264993796 diff: 0.8232510685233763
Simple mean: 71.76743450695217 diff: 0.9754229255375861
True std: 58.31532031908203
Mice forest std: 61.67046958546801 diff: 3.355149266385979
Simple std: 41.852838829183426 diff: 16.462481489898607
```

Note: I used a subset of observations (200) for this evaluation since mice_forest_impute got computationally too expensive (11988 observations × 37 variables × 48 timepoints which after flattening to (n_obs * n_t, n_vars))


ehrapy/preprocessing/_imputation.py
eroell, line 608: Should we add another line as for the other methods about the vertical stacking coming with the 3D option?

ehrapy/preprocessing/_imputation.py
```diff
@@ -654,40 +671,54 @@ def _miceforest_impute(
...
+    if mtx.ndim == 3:
+        is_3d = True
+        n_obs, n_vars, n_t = mtx.shape
+        mtx = (
+            mtx[:, var_indices, :]
+            .astype(input_dtype, copy=True)
+            .transpose(0, 2, 1)
+            .reshape(n_obs * n_t, len(var_indices))
         )
+        col_names = edata.var_names[var_indices]
```
eroell: wouldn't here `col_names` just be `var_names`?

ehrapy/preprocessing/_imputation.py
```diff
@@ -654,40 +671,54 @@ def _miceforest_impute(
...
+        selected_columns,
+        num_datasets=1,
+        save_all_iterations_data=save_all_iterations_data,
+        random_state=random_state,
+    )
 
-        kernel.mice(iterations=iterations, variable_parameters=variable_parameters or {}, verbose=verbose)
-        data_df = kernel.complete_data(dataset=0, inplace=inplace)
+    kernel.mice(iterations=iterations, variable_parameters=variable_parameters or {}, verbose=verbose)
+    data_df.iloc[:, column_indices] = kernel.complete_data(dataset=0, inplace=inplace)
```
eroell: Since the data is copied, could `inplace` always be set to true? And then omitted as argument from the function?

The data is potentially copied into mtx, and then once more into data_df

eroell (commented):
Thank you!

And as for the other PR with missforest #1052 : could you create an analog snippet as PR comment with the physionet challenge data, comparing the imputed data with e.g. simpleimpute in terms of imputation quality?

ehrapy/preprocessing/_imputation.py
sueoglu: Yes, I forgot about it

ehrapy/preprocessing/_imputation.py
sueoglu: when inplace is set to True, it returns None. Also I noticed that mice_forest_impute has both inplace and copy as arguments and I agree to omit the inplace argument, set the internal miceforest operations `inplace` argument to False (as we only call the complete_data() only once).

ehrapy/preprocessing/_imputation.py
```diff
@@ -644,50 +660,63 @@ def load_dataframe(arr, columns, index):
...
+    if mtx.ndim == 3:
+        is_3d = True
+        n_obs, n_vars, n_t = mtx.shape
+        mtx = (
+            mtx[:, var_indices, :]
+            .astype(input_dtype, copy=True)
+            .transpose(0, 2, 1)
+            .reshape(n_obs * n_t, len(var_indices))
         )
+        var_names = edata.var_names[var_indices]
```
eroell: var_names already exist, right?

ehrapy/preprocessing/_imputation.py
```diff
@@ -644,50 +660,63 @@ def load_dataframe(arr, columns, index):
...
 
-@load_dataframe.register
+@load_dataframe.register(np.ndarray)
 def _(arr: np.ndarray, columns, index):
     return pd.DataFrame(arr, columns=columns, index=index)
 
 
 def _miceforest_impute(
-    edata, var_names, save_all_iterations_data, random_state, inplace, iterations, variable_parameters, verbose, layer
+    edata, var_names, save_all_iterations_data, random_state, iterations, variable_parameters, verbose, layer
```
eroell, line 747: singledispatch pattern

ehrapy/preprocessing/_imputation.py
```diff
@@ -510,19 +510,9 @@ def miss_forest_impute(
...
-        [[-12.12732884, -18.37304373],
-        [         nan,  -0.91339411],
-        [         nan,  -7.88514984]]
-        >>> edata_imputed.layers["tem_data"][0, :, :]
-        [[-12.12732884, -18.37304373],
-        [ -0.3278448 ,  -0.91339411],
-        [ -4.39722201,  -7.88514984]]
+        >>> edata = ed.dt.mimic_2()
+        >>> edata = ep.pp.encode(edata, autodetect=True)
+        >>> ep.pp.miss_forest_impute(edata)
```
eroell, line 515: Can we keep the example output? It looked quite instructive

ehrapy/preprocessing/_imputation.py
```diff
@@ -634,13 +623,29 @@ def mice_forest_impute(
...
+        [ -2.29990557,  -3.13348925],
+        [ -6.72812888, -10.87061402]]
+
     """
     if copy:
         edata = edata.copy()
 
+    if edata.X is None and layer is None:  # if edata is 3D
+        raise ValueError(
+            "3D imputation requires a layer to be specified. Pass the layer containing the full temporal data."
```
eroell: This is not specific for 3D though right? The edata object is just empty in that case, and if an error needs to be raised, it can be general and does not need to mention 3D

sueoglu: Yes true

ehrapy/preprocessing/_imputation.py
sueoglu: its miss_forest example output but I can fur sure add that back here

PR #1056, opened 2026-04-24
Efficient  non dependency version of MCARTest

### LINKED ISSUE #980: Implement efficient version of pyampute's MCARTest
### Description of feature

We're currently only using MCARTest in one place. The original implementation seems to be slow and inefficient. We should implement our own MCARTest to have one less dependency and have a more efficient implementation

agerardy:
**PR Checklist**

**Description of changes**

- Fixes #980 
- Implemented efficient internal version of MCARTest to remove pyampute dependency
- Benchmark results: ~20x speed up compared to pyampute implementation for Little's test across tested dataset sizes, even larger improvements for pairwise welch t-test (see table below)
- resulting p-values are very close to pyampute results but not identical due to floating point accumulation
- updated tests
- `_little_mcar_test` is a lot of new code, might require additional inline comments for explanation

** Technical Details**
- New `mcar_test` implementation:
  - accepts `EHRData` 2D input (`@function_2D_only()`)
  - rejects sparse matrices
  - materializes dask arrays with a warning (global covariance/pattern grouping requires in-memory data)
- `little` method:
  - computes global means/covariance
  - groups rows by missingness pattern
  - accumulates Little’s chi-square statistic
  - uses `chi2.sf` for p-value
- Numerical robustness:
  - uses `np.linalg.solve`
  - falls back to `np.linalg.pinv` when covariance blocks are singular/ill-conditioned

**Additional context**
 Benchmark results:
#### Method: `little`

| n | p | missing | new (s) | new (MB) | new p | old (s) | old (MB) | old p | speedup |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 100 | 10 | 10% | 0.0196 | 0.05 | 0.2394 | 0.3753 | 0.10 | 0.2392 | 19.1x |
| 500 | 20 | 20% | 0.2019 | 0.26 | 0.3134 | 4.0057 | 0.30 | 0.3161 | 19.8x |
| 1000 | 50 | 20% | 0.3283 | 1.28 | 0.4479 | 7.3379 | 1.30 | 0.4487 | 22.4x |
| 2000 | 50 | 30% | 0.6513 | 2.52 | 0.4684 | 14.6964 | 2.53 | 0.4671 | 22.6x |
| 5000 | 50 | 20% | 1.6888 | 6.24 | 0.4981 | 38.0112 | 6.22 | 0.4982 | 22.5x |

#### Method: `ttest`

| n | p | missing | new (s) | new (MB) | new min-p | old (s) | old (MB) | old min-p | speedup |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 100 | 10 | 10% | 0.0147 | 0.03 | 0.0054 | 0.6314 | 0.05 | 0.0054 | 43.0x |
| 500 | 20 | 20% | 0.0342 | 0.23 | 0.0062 | 2.5633 | 0.15 | 0.0062 | 75.0x |
| 1000 | 50 | 20% | 0.1252 | 0.86 | 0.0001 | 16.2233 | 0.55 | 0.0001 | 129.6x |
| 2000 | 50 | 30% | 0.2113 | 1.54 | 0.0000 | 15.5949 | 0.95 | 0.0000 | 73.8x |
| 5000 | 50 | 20% | 0.3757 | 3.93 | 0.0002 | 15.6982 | 2.22 | 0.0002 | 41.8x |


eroell (commented):
Thank you!
Now that we have a custom implementation, can you generate a few representative samples of missing data, compute pyampute's results, save them as test reference files, and check in the tests that this implementation matches the results please?

ehrapy/preprocessing/_quality_control.py
```diff
@@ -637,33 +635,120 @@ def _outlier_flags_and_scores(
...
+    mtx = edata.X if layer is None else edata.layers[layer]
+
+    if sparse.issparse(mtx):
+        raise TypeError("mcar_test does not support sparse matrices. Convert to dense first.")
+
+    if isinstance(mtx, DaskArray):
+        logger.warning("Dask array detected – materialising into memory for MCAR test.")
+        mtx = mtx.compute()
+
+    X = np.asarray(mtx, dtype=np.float64)
```
Zethson: Such lines can easily create fat copies of the matrix which can be a lot of overhead. Moreover, you're not using the data type that the user used initially which can lead to unexpected results.
Can this be avoided?

ehrapy/preprocessing/_quality_control.py
```diff
@@ -637,33 +635,120 @@ def _outlier_flags_and_scores(
...
-        method: Whether to perform a chi-square test on the entire dataset (“little”) or separate t-tests for every combination of variables (“ttest”).
-        layer: Layer to apply the test to. Uses X matrix if set to `None`.
+        method: ``"little"`` for a global chi-square test or ``"ttest"`` for
+            pairwise Welch t-tests across all variable combinations.
+        layer: Layer to apply the test to. Uses ``X`` if ``None``.
 
     Returns:
-        A single p-value if the Little's test was applied or a Pandas DataFrame of the p-value of t-tests for each pair of features.
+        A single p-value (``float``) for ``"little"`` or a :class:`~pandas.DataFrame`
+        of p-values for ``"ttest"``.
```
Zethson: Formatting...

Zethson (changes requested):
Thank you very much! Mostly nitpicks.

1. The imports don't need to be scoped. I guess this is caused by the LLMs.
2. It's inconsistent that sparse matrices raise an error and Dask arrays are materialized. Is it possible for you to add support for these matrix types? Sparse 2D should not be too difficult I hope... but @eroell can surely say whether that's worth your time or not.
3. How much work would using the array API be here?
4. I see a few loops here. Are they necessary or can this be vectorized?
5. Could you please link to the implementation where you got the math from in the code?
6. @eroell note about ensuring that the new implementation matches the old is essential

ehrapy/preprocessing/_quality_control.py
agerardy: ah yes, everything on one line, right?

eroell: yes

agerardy: the old line reads better anyway to be honest

ehrapy/preprocessing/_quality_control.py
agerardy: the float64 is a requirement of covariance estimation and `linalg.solve` but i can add a warning if datatypes are converted. also added `copy=False`to ensure no fat copies are made.

Zethson: Ah, I didn't know. Could you please add a comment there explaining this?

agerardy:
1. moved the scoped imports, you were right those were made by the LLM
2. added support for sparse matrices by densifying them, kept the same warning as with dask arrays and added tests for it
3. array API would require a lot of workarounds, `np.unique`, `np.nanmean`, `np.nanstd` and the `chi2` are all not in the standard currently
4. vectorized the inner loop in `mcar_t_tests` need to keep the other loops, the outer one is unavoidable because each `i` defines a different row-grouping with a different number of rows, so the sub-matrices cant be stacked. The loop in `_little_mcar_test` is also unavoidable since each unique missingness pattern selects a different subset of columns (different submatrix size and different linalg.solve call), so it cant be batched without padding and masking tricks that would add more complexity 
5. added links to the original source for little's test and an open access paper that describes the implementation in R, which is what I actually used. I also read over the original source, but didn't have "official" access to it, and it is even more mathy and exhausting to read. I couldnt find a python implementation to copy, so I freestyled it based on the paper and R scripts I found and let claude correct and clean up the chaos I produced :)
6. On matching the old implementation: I added generated reference files and tests as well as the script to generate them for reference. The results of `ttest` are essentially identical to pyamputes (to 6 decimal places). For `little`: the tolerance is 1% which I had to do because of numerical differences in the covariance matrices. pyampute uses `pandas.Dataframe.cov()` with pairwise complete observations, so the mean is only computed for the rows where both columns are observed. The new implementation subtracts the global `np.nanmean` (computed per column over all non-NaN rows) before building the cross-products. When missingness is high and not uniform, the global column mean can differ from the pair-specific mean, introducing an offset. this leads to slightly different p-values but the same conclusion in all tested scenarios. I could also implement the pandas behaviour if we want to match the old behaviour more closely.
7. I ran new benchmarks with the updated code and updated the table in the original comment

Zethson:
Thank you very much for the elaborate and useful answer (and improvements!).

> added support for sparse matrices by densifying them, kept the same warning as with dask arrays and added tests for it

I've also done this plenty of times, but generally, we want to avoid a silent densification because it's not transparent, causes memory spikes, and may lead to unintended consequences. If you at least map back to sparse after densification, we only have to live with two of these problems but generally, it is to be avoided. The right solution is to implement sparse code.

The benchmarks seem to give good results so let's ship it then

agerardy:
> I've also done this plenty of times, but generally, we want to avoid a silent densification because it's not transparent, causes memory spikes, and may lead to unintended consequences. If you at least map back to sparse after densification, we only have to live with two of these problems but generally, it is to be avoided. The right solution is to implement sparse code.

It did seem like a too easy solution. I'll try implementing a sparse version, if it is too complicated I'll revert to the original version

agerardy:
I added sparse logic for both `little` and `ttest`.  
It is a lot more code than the dense path, but it avoids silent densification.
I also added a dedicated parity test (`test_mcar_test_sparse_matches_dense`) to ensure sparse and dense paths produce the same outputs (`little` scalar and `ttest` matrix). For Dask, I could not implement a true lazy path, so Dask arrays are still materialized with a warning
In principle, the covariance part could be computed lazily using the same algebraic trick as the sparse implementation, but the missingness-pattern step (`np.unique(mask, axis=0)`) still requires a global row-wise mask and is the main blocker for a practical non-materializing dask implementation
For sparse arrays the trick is to avoid  explicit centering/densification and instead compute the covariance from sparse matrix products with NaN aware counts handled vis sparse indicator matrices

Zethson:
Great! However, could you please consider using the singledispatch pattern that we're using all over the code base? Moreover, please ensure that docstrings are clean of useless information. We don't document that functions support sparse arrays in docstrings. I guess that's also LLM slop.

Referring to

>   Sparse matrices are handled natively without full densification.
    Dask arrays are materialised into memory with a warning.
    Non-float64 data is cast to ``float64`` with a warning, because this may temporarily increase memory usage.

agerardy:
I considered removing those, thought they were helpful. Also removed the docstrings from helper functions, I think we talked about this before. I'm sorry I keep making these same mistakes, adding this stuff to a checklist that I will go through before every future push, so you don't have to keep repeating yourselves

ehrapy/preprocessing/_quality_control.py
```diff
@@ -655,15 +658,247 @@ def mcar_test(
...
+    return float(chi2.sf(d2, df))
+
+
+def _sparse_prepare(X_sparse):
+    csc = X_sparse.tocsc()
+    n, p = csc.shape
+
+    is_nan = np.isnan(csc.data)
+    col_idx = np.repeat(np.arange(p), np.diff(csc.indptr))
+    nan_mask = np.zeros((n, p), dtype=bool)
```
Zethson: This materializes a full dense array. Therefore, this leads to a lot of memory consumption. The implementation needs an overhaul to ensure that everything stays sparse. Like ideally, you shouldn't be playing with numpy operations but just use scipy sparse array operations

ehrapy/preprocessing/_quality_control.py
```diff
@@ -655,15 +658,247 @@ def mcar_test(
...
+    X_clean.eliminate_zeros()
+
+    nan_ind = csc.copy()
+    nan_ind.data[:] = is_nan.astype(np.float64)
+    nan_ind.eliminate_zeros()
+
+    return X_clean, nan_ind, nan_mask
+
+
+@_little_mcar_test.register(sparse.spmatrix)
```
Zethson: Can you demonstrate that this never materializes a full dense array? There's a lot of things here where I'm not sure

ehrapy/preprocessing/_quality_control.py
```diff
@@ -655,15 +658,247 @@ def mcar_test(
...
+            equal_var=False,
+        ).pvalue
+
+    return pd.DataFrame(result, index=var_names, columns=var_names)
+
+
+@_mcar_t_tests.register(sparse.spmatrix)
+def _(X_sparse, var_names: np.ndarray) -> pd.DataFrame:
+    n, m = X_sparse.shape
+    result = np.full((m, m), np.nan)
```
Zethson: This also raises concern. A full matrix?

Zethson (changes requested):

ehrapy/preprocessing/_quality_control.py
```diff
@@ -655,15 +658,247 @@ def mcar_test(
...
+        layer: Layer to apply the test to. Uses ``X`` if ``None``.
 
     Returns:
         A single p-value if the Little's test was applied or a Pandas DataFrame of the p-value of t-tests for each pair of features.
     """
-    df = ed.io.to_pandas(edata, layer=layer)
-    from pyampute.exploration.mcar_statistical_tests import MCARTest
+    mtx = edata.X if layer is None else edata.layers[layer]
+
+    if isinstance(mtx, DaskArray):
```
eroell: We generally use the single-dispatch pattern to raise NotImplementedErrors in this case, rather than materialising.
Users could materialize if they really want it and we don't have it yet.

ehrapy/preprocessing/_quality_control.py
eroell: Demonstration would mean that memory consumption of a very large sparse array with many zero values would be significantly lower than for the same very large array in dense format (can be checked with a profiler like for instance scalene, whose report you could show here)

ehrapy/preprocessing/_quality_control.py
eroell: the matrix is of size #features x #features, which is a densely computed result of this function so it should be fine

tests/_scripts/generate_mcar_references.py
```diff
@@ -0,0 +1,58 @@
...
+
+import json
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+from pyampute.exploration.mcar_statistical_tests import MCARTest
+
+
+def _make_mcar(n_obs: int, n_vars: int, missing_rate: float, seed: int) -> pd.DataFrame:
```
eroell: can you use ehrdata blobs for this?

tests/_scripts/generate_mcar_references.py
```diff
@@ -0,0 +1,58 @@
...
+    rng = np.random.default_rng(seed)
+    data = rng.standard_normal((n_obs, n_vars))
+    miss = rng.random((n_obs, n_vars)) < missing_rate
+    data[miss] = np.nan
+    cols = [f"v{i}" for i in range(n_vars)]
+    return pd.DataFrame(data, columns=cols)
+
+
+def _make_mar(n_obs: int, n_vars: int, missing_pct: float, seed: int) -> pd.DataFrame:
```
eroell: also ehrdata blobs, with an additional small MAR mechanism after it has been generated since ehrdata blobs does not provide MAR missing data.

This, and having the same function as a fixture in the tests, would allow to not require us to store the rather large raw data csvs

tests/preprocessing/test_quality_control.py
```diff
@@ -438,26 +440,77 @@ def test_qc_lab_measurements_defaults_to_all_vars():
...
+
+def test_mcar_test_ttest_detects_mar(mar_edata):
+    result = mcar_test(mar_edata, method="ttest")
+    assert result.shape == (mar_edata.n_vars, mar_edata.n_vars)
+    p_col0_given_miss9 = result.iloc[-1, 0]
+    assert not np.isnan(p_col0_given_miss9)
+    assert p_col0_given_miss9 < 0.05
+
+
+def test_mcar_test_little_matches_pyampute_references():
```
eroell: Could you add a pytest parametrization for dense and sparse here?
This would avoid having a separate extra test between dense and sparse, and is also what we do a couple of other recent PRs

tests/preprocessing/test_quality_control.py
```diff
@@ -438,26 +440,77 @@ def test_qc_lab_measurements_defaults_to_all_vars():
...
+    for scenario_name, expected_p in expected.items():
+        df = pd.read_csv(ref_dir / f"{scenario_name}.csv")
+        edata = ed.EHRData(X=df.to_numpy(dtype=float), var=pd.DataFrame(index=df.columns))
+        observed = mcar_test(edata, method="little")
+        assert np.isclose(observed, expected_p, rtol=1e-2, atol=2e-3), (
+            f"Mismatch for {scenario_name}: observed={observed}, expected={expected_p}"
+        )
+
+
+def test_mcar_test_ttest_matches_pyampute_references():
```
eroell: Could you add a pytest parametrization for dense and sparse here?
This would avoid having a separate extra test between dense and sparse, and is also what we do a couple of other recent PRs

eroell (commented):
I'm not entirely sure the sparse implementation is actually sparse; could you show this by benchmarking implementations of large arrays with scalene?

If it turns out to be not sparse and hard to do, we'd ultimately fall back to support dense only.

ehrapy/preprocessing/_quality_control.py
Zethson: I guess I was thinking genes here - then it wouldn't be that small but yeah probably fine

agerardy:
I decided not to implement sparse logic after all, you were right, the last implementation wasn't actually sparse as a benchmark also showed it actually used more memory than the dense path. I didn't manage to make it entirely sparse. Full list of changes:

- Dropped sparse matrix implementation
- dropped dask materialization, both dask and sparse now raise NotImplementedError via singeldispatch
- removed large CSV files
- test data now generated from ehrdata_blobs
- added MAR mechanism for ehrdata_blobs
- changed the .py reference script to .ipynb to be consistent with the rest of the scripts
- merged the not implemented tests into one parametrised test
- the failed notebook run is due to a failed download as far as I can tell, not any of my changes

docs/tutorials/notebooks
eroell, line 1: The submodule is wrong again, it should point to the latest commit on `ehrapy-tutorials`.

ehrapy/preprocessing/_quality_control.py
```diff
@@ -655,15 +657,137 @@ def mcar_test(
...
+
+    Examples:
+        >>> import ehrdata as ed
+        >>> import ehrapy as ep
+        >>> edata = ed.dt.ehrdata_blobs(
+        ...     n_observations=100, n_variables=5, missing_values=0.1, random_state=0, n_centers=1, base_timepoints=1
+        ... )
+        >>> ep.pp.mcar_test(edata)
+        0.327...
+        >>> ep.pp.mcar_test(edata, method="ttest")
```
eroell:
```suggestion
```

submodules/ehrdata
eroell, line 1: Point this to either the latest ehrdata commit, or what the ehrapy main branch points to currently

tests/_scripts/generate_mcar_references.ipynb
```diff
@@ -0,0 +1,296 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Generate MCAR reference files\n",
+    "\n",
+    "Runs `pyampute` on deterministic datasets generated with `ehrdata.dt.ehrdata_blobs`\n",
```
eroell, line 9: Please print pyampute, ehrdata, and ehrapy versions used

tests/preprocessing/test_quality_control.py
```diff
@@ -438,26 +468,63 @@ def test_qc_lab_measurements_defaults_to_all_vars():
...
+    exp_vals = expected.to_numpy(dtype=float)
+    finite = np.isfinite(obs_vals) & np.isfinite(exp_vals)
+    nan_match = np.isnan(obs_vals) & np.isnan(exp_vals)
+
+    assert np.all(finite | nan_match), "NaN pattern mismatch vs pyampute reference"
+    assert np.allclose(obs_vals[finite], exp_vals[finite], rtol=1e-6, atol=1e-10)
+
+
+@pytest.mark.parametrize("method", ["little", "ttest"])
+@pytest.mark.parametrize("array_type", [csr_matrix, as_dense_dask_array])
```
eroell: please look at other PR's where the array type is used for test parametrization, we never (should) use `csr_matrix`

eroell (changes requested):
See comments. The tests seem to now capture a ground truth comparison, that is that the test seems to be interchangeably useable with pyampute, cool

tests/preprocessing/test_quality_control.py
agerardy: you're saying I should use `sp.csr_array` or something else? I found some uses of csr_matrix but most seem to use csr-array

eroell: Should have been more clear, these types https://github.com/theislab/ehrapy/blob/a51236f3cdc33fcfc49f48b3d3f1049d900dc620/ehrapy/_types.py#L24-L29 are what we should aim to cover in our type checks :)

So if possible use the parametrization of `ARRAY_TYPES_NUMERIC`, and if the tests are set up in a way that this would yield too many if-else statements, getting e.g. the sparse types `sp.csr_array` and `sp.csc_array` for sparse test cases would be good

agerardy: ok, thats what I did, uses `as_dense_dask_array` and `sp.csr_array` now, `sp.csc_array` would be redundant if i understand it correctly

pyproject.toml
```diff
@@ -46,6 +46,7 @@ classifiers = [
 ]
 
 dependencies = [
+    "anndata<0.12.12", # temporarily pinning because ehrdata is incompatible with newer anndata for now
```
Zethson:
```suggestion
```

PR #1060, opened 2026-04-28
sparse array support for  `_warn_imputation_threshold`

### LINKED ISSUE #1058: Refactor _warn_imputation_threshold to handle sparse arrays without qc_metrics
Currently _warn_imputation_threshold internally calls `qc_metrics` to compute missing value percentages, which does not support sparse arrays. This causes imputation functions that could support sparse arrays to crash before even reaching the imputer

sueoglu:
fixes #1058 
`_warn_imputation_threshold` is calling `qc_metrics` to compute missing value percentages, which does not support sparse arrays

- Refactored _warn_imputation_threshold: a sparse aware path that computes missing value percentages directly using `np.isnan` on the stored values assuming missing values are explicitly stored as NaN in the sparse structure as well
- Extended test_warning() to cover sparse arrays

### DIFF
--- a/ehrapy/_types.py
+++ b/ehrapy/_types.py
@@ -10,7 +10,7 @@ from fast_array_utils.conv import to_dense
 from ehrapy._compat import as_dense_dask_array
 
 KnownTransformer = Literal["pynndescent", "sklearn"]
-CSBase = sp.csr_matrix | sp.csc_matrix
+CSBase = sp.csr_matrix | sp.csc_matrix | sp.csr_array | sp.csc_array
 RNGLike = np.random.Generator | np.random.BitGenerator
 SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
 AnyRandom = int | np.random.RandomState | None
--- a/ehrapy/_types.py
+++ b/ehrapy/_types.py
@@ -10,7 +10,7 @@ from fast_array_utils.conv import to_dense
 from ehrapy._compat import as_dense_dask_array
 
 KnownTransformer = Literal["pynndescent", "sklearn"]
-CSBase = sp.csr_matrix | sp.csc_matrix | sp.csr_array | sp.csc_array
+CSBase = sp.csr_array | sp.csc_array
 RNGLike = np.random.Generator | np.random.BitGenerator
 SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
 AnyRandom = int | np.random.RandomState | None
--- a/ehrapy/preprocessing/_imputation.py
+++ b/ehrapy/preprocessing/_imputation.py
@@ -24,6 +24,7 @@ from ehrapy._compat import (
 )
 from ehrapy._progress import spinner
 from ehrapy._types import CSBase
+from ehrapy.preprocessing._quality_control import _compute_missing_values
 
 if TYPE_CHECKING:
     from anndata import AnnData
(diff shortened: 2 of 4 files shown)


ehrapy/preprocessing/_imputation.py
```diff
@@ -695,6 +695,10 @@ def _warn_imputation_threshold(
 ) -> dict[str, int]:
     """Warns the user if the more than $threshold percent had to be imputed.
 
+    For sparse arrays, missing values are assumed to be represented as ``np.nan``.
+    Use :func:`ehrdata.harmonize_missing_values` to convert other missing value
+    symbols to ``np.nan`` before imputing.
```
Zethson: Formatting

ehrapy/preprocessing/_imputation.py
```diff
@@ -704,9 +708,22 @@ def _warn_imputation_threshold(
     try:
         edata.var["missing_values_pct"]
     except KeyError:
-        from ehrapy.preprocessing import qc_metrics
+        mtx = edata.X if layer is None else edata.layers[layer]
+        if isinstance(mtx, (sp.csr_array, sp.csc_array)):
```
Zethson: https://github.com/theislab/ehrapy/blob/main/ehrapy/_types.py#L13

Zethson (commented):
Thank you!

ehrapy/preprocessing/_imputation.py
sueoglu: should I use CSBase in the isinstance check and change the test fixtures to use csr_matrix/csc_matrix to match? Because currently in the tests we use sp.csr_array and sp.csc_array

Zethson: We should actually only use array nowadays. The matrix API is dead.

sueoglu: so then should I update CSBase to use sp.csr_array and sp.csc_array, and use it in the isinstance check?

Zethson: Yup! Or all 4 even for now. But I honestly don't think that anyone should use the deprecated matrix API. They might have the same super class?

ehrapy/_types.py
```diff
@@ -10,7 +10,7 @@
 from ehrapy._compat import as_dense_dask_array
 
 KnownTransformer = Literal["pynndescent", "sklearn"]
-CSBase = sp.csr_matrix | sp.csc_matrix
+CSBase = sp.csr_matrix | sp.csc_matrix | sp.csr_array | sp.csc_array
```
eroell:
```suggestion
CSBase = sp.csr_array | sp.csc_array
```

ehrapy/preprocessing/_imputation.py
```diff
@@ -704,9 +709,22 @@ def _warn_imputation_threshold(
...
+            n_nan_per_col = np.array(
+                [np.sum(np.isnan(mtx_csc.data[mtx_csc.indptr[i] : mtx_csc.indptr[i + 1]])) for i in range(mtx.shape[1])]
+            )
+            missing_pct = (n_nan_per_col / n_obs) * 100
+            edata.var["missing_values_pct"] = missing_pct
+        else:
+            from ehrapy.preprocessing import qc_metrics
 
-        qc_metrics(edata, layer=layer)
+            qc_metrics(edata, layer=layer)
```
eroell: `qc_metrics` computes a bit more than `edata.var["missing_values_pct"]`.

Hence the computed results that are stored in `edata` would vary depending on whether sparse or dense data is passed.

I'd prefer to
1. move this missingness computation into the qc file, where the qc_metrics function dispatches the missingness computation to either dense or sparse implementation
2. for imputation methods, only compute missingness with this new function, and not the entire qc metrics

what do you think?

sueoglu: Yes i think that make sense to have consistent columns, im on it

sueoglu: this means then we're only calculating missingnes via this new single dispatched function in qc file and using that here right?

eroell: yes exactly

ehrapy/preprocessing/_imputation.py
```diff
@@ -730,6 +732,10 @@ def _warn_imputation_threshold(
 ) -> dict[str, int]:
     """Warns the user if the more than $threshold percent had to be imputed.
 
+    For sparse arrays, missing values are assumed to be represented as `np.nan`.
```
eroell:
```suggestion
```

Not just for sparse arrays, right?

eroell (approved):
I did not check whether the performance of converting csc to csr is a performance degradation, but think it would not be dramatic.

One last comment, to make the internal docstring a bit more clear.

Generally, all tests still passing looks good!

PR #1063, opened 2026-05-08
address pandas futurewarning

### LINKED ISSUE #1059: Address FutureWarning Pandas
See the current warning in our tests:

`/home/runner/work/ehrapy/ehrapy/ehrapy/plot/_variable_correlation_plot.py:197: FutureWarning: The previous implementation of stack is deprecated and will be removed in a future version of pandas. See the What's New notes for pandas 2.1.0 for details. Specify future_stack=True to adopt the new implementation and silence this warning.`

TODO:
Check what is new or changed here, and adjust the implementation to work with the future behaviour

agerardy:
**PR Checklist**

- [ ] If you've fixed a bug or added code that should be tested, add tests!

**Description of changes**

Fixes #1059 pandas `FutureWarnings` in the tests. Affected files

- `ehrapy/plot/_variable_correlation_plot.py`: Updated  `.stack()` calls to use `future_stack=True` (the new default in pandas 3.0) and removed `dropna=False`, which is not accepted when `future_stack=True`.
- `ehrapy/plot/_sankey.py`: Added `observed=True` to a `groupby` call on obs columns
- `ehrapy/preprocessing/_bias.py`: `observed=True` again

**Technical details**

`stack`: pandas 2.1 introduced a new implementation of `DataFrame.stack()` that will become the only version in 3.0. The key behavioural difference is that the new implementation never introduces spurious `NaN` values, so `dropna=False` is no longer needed. The output for our use cases is identical.

`groupby(observed=True)`: The pandas default for `observed` on categorical columns changes from `False` to `True` in pandas 3.0. With `observed=False`, all possible category combinations appear in the result even with zero counts. For our use cases here this is incorrect behaviour: zero count category pairs in the bias analysis would result in `0/0 = NaN` ratios. In a sankey diagram a zero count flow makes no sense as well, so in both cases I set `observed=True`.

### DIFF
--- a/ehrapy/plot/_sankey.py
+++ b/ehrapy/plot/_sankey.py
@@ -69,7 +69,7 @@ def sankey_diagram(
     source_levels, target_levels = [], []
     for i in range(len(columns) - 1):
         col_from, col_to = columns[i], columns[i + 1]
-        flows = df.groupby([col_from, col_to]).size().reset_index(name="count")
+        flows = df.groupby([col_from, col_to], observed=True).size().reset_index(name="count")
         sources.extend(col_from + ": " + flows[col_from].astype("string"))
         targets.extend(col_to + ": " + flows[col_to].astype("string"))
         values.extend(flows["count"].to_numpy())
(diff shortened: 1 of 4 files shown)


eroell (approved):
Thanks!

PR #1016, opened 2026-02-04
Remove Legacy Code

eroell:
**PR Checklist**

- [x] This comment contains a description of changes (with reason)
- [x] Referenced issue is linked
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] Documentation in `docs` is updated

**Description of changes**
- Remove legacy code (all of which had a `FutureWarning` in the last release):
  - `ehrapy.io`: `read_csv`, `read_h5ad`, `read_fhir`, `write`
  - `ehrapy.ad`: `move_to_obs`, `move_to_x`, `infer_feature_types`, `feature_type_overview`, `replace_feature_type`, `df_to_anndata`, `anndata_to_df`
  - `ehrapy.dt`: All datasets and the `download` function used internally
  As those functions exist in enhanced versions in the `ehrdata` library.
- Fixes #964.

**664 warnings -> <100 warnings**

**Technical details**
Further minor changes:
- Adapt new scanpy arguments:
  - `ep.pl.violin`: `scale` -> `density_norm`
  - `ep.pl.stacked_violin`: `scale` -> `density_norm`, `order` -> `categories_order`
  - `ep.pl.rank_features_groups_violin`: `scale` -> `density_norm`
- Adjust behavior
  - `edata.obs_keys()` -> `edata.obs`
  - `edata.uns_keys()` -> `edata.uns`
- Register `gpu` pytest mark
- Fix a few tests to not raise Warnings anymore

**Additional context**

PR #1080, opened 2026-06-11
Cache CI datasets and harden against flaky downloads

sueoglu:
**Problem**
CI jobs download example datasets (mimic_2, diabetes_130_fairlearn, …) from upstream hosts such as physionet.org on every run. When those hosts are slow, rate-limiting, or down, the test and notebook jobs fail or hang — a recurring source of red CI that has nothing to do with the change under test. This mirrors the issue fixed in ehrdata ([theislab/ehrdata#250](https://github.com/theislab/ehrdata/pull/250)); this PR ports the same approach to ehrapy.

**Changes**
Dataset downloads in ehrapy are fully delegated to the ehrdata submodule (the legacy in-repo `ehrapy/data/_dataloader.py` was removed in #1016), so no Python changes are needed here — ehrdata's downloader already short-circuits when a file exists and already has the resilient retry loop. This PR is CI-only.

`.github/workflows/test_cpu.yml` (test job):

Restore `ehrapy_data/` before the run via `actions/cache/restore@v4`, key ehrapy-data-tests-v1 (static, shared across the Python matrix).
Save the cache only on `cache-hit != 'true' && success()`, so a partial/failed download is never persisted under the immutable key.
Set `EHRDATA_DOWNLOAD_MAX_RETRIES=5 / EHRDATA_DOWNLOAD_RETRY_DELAY=10` on the test step so a genuinely unreachable host fails fast instead of hanging.
.github/workflows/run_notebooks.yml (notebook job):

Same restore/save pattern, but keyed per notebook (`ehrapy-data-notebook-${{ matrix.notebook }}-v1`) since each tutorial pulls a different subset.
Cache path is `docs/tutorials/notebooks/ehrapy_data` — nbconvert executes with the notebook's directory as the working directory, and ehrdata's DEFAULT_DATA_PATH is the relative ehrapy_data, so downloads land there.
Same EHRDATA_DOWNLOAD_* env vars on the execution step.

**How caching works here**
ehrdata's `_download()` returns early when the target file already exists, issuing no network request. So once a cache is populated, a cache hit means the datasets are present on disk and the loaders never touch the network. The first run after a key bump is a cache miss that downloads and then saves; every subsequent run restores.

**Why the immutable-key guard matters**
actions/cache keys are write-once. Saving only after a green run that actually downloaded prevents a truncated download (e.g. a request that died mid-stream) from being sealed into ehrapy-data-*-v1 and breaking every future run. Bump the -v1 suffix when a dataset URL changes.

**Verification**
Ran the two networked loaders locally through the same ehrdata submodule code and default path CI uses:

ed.dt.mimic_2() → downloaded, shape (1776, 46, 1)
ed.dt.diabetes_130_fairlearn() → downloaded, shape (101766, 22, 1)
Confirmed: (1) cold download fetches and parses both; (2) a warm re-run does zero network calls and returns instantly — the exact behavior a cache hit relies on; (3) the active ehrdata submodule has the fixed retry guard (retry_count < max_retries).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,6 @@ classifiers = [
 
 dependencies = [
     "ehrdata",
-    "pooch",  # required by the bundled ehrdata submodule, which is installed with --no-deps
     "scanpy",
     "scverse-misc[settings]>=0.0.7",
     "requests",
@@ -124,13 +123,13 @@ allow-direct-references = true
 installer = "uv"
 features = [ "dev" ]
 post-install-commands = [
-  "uv pip install --reinstall --no-deps -e submodules/ehrdata",
+  "uv pip install --reinstall -e submodules/ehrdata",
 ]
 
 [tool.hatch.envs.docs]
 features = [ "doc", "dask", "leiden" ]
 post-install-commands = [
-  "uv pip install --reinstall --no-deps -e submodules/ehrdata",
+  "uv pip install --reinstall -e submodules/ehrdata",
 ]
 scripts.build = "sphinx-build -M html docs docs/_build -W {args}"
 scripts.open = "python -m webbrowser -t docs/_build/html/index.html"
@@ -140,7 +139,7 @@ scripts.clean = "git clean -fdX -- {args:docs}"
 [tool.hatch.envs.hatch-test]
 features = [ "dask", "leiden" ]
 post-install-commands = [
-  "uv pip install --reinstall --no-deps -e submodules/ehrdata",
+  "uv pip install --reinstall -e submodules/ehrdata",
 ]
 
 [[tool.hatch.envs.hatch-test.matrix]]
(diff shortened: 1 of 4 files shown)


Zethson (commented):
Please ensure that we don't have useless AI comments in the workflows. They clutter.

Thank you! Great idea.

sueoglu:
> Please ensure that we don't have useless AI comments in the workflows. They clutter.
> 
> Thank you! Great idea.

Yess sure! I am trying out Claude Code and will clean up unnecessary comments etc

eroell:
by merging main, the notebook will execute again

pyproject.toml
```diff
@@ -47,6 +47,7 @@ classifiers = [
 
 dependencies = [
     "ehrdata",
+    "pooch",  # required by the bundled ehrdata submodule, which is installed with --no-deps
```
eroell: Better to drop the --no-deps (ehrdata dependencies are quite minimal anyways) than to duplicate pooch as dependency of ehrapy

eroell (approved):
Thank you!

Issue #534, opened 2023-06-04
Add a notebook for the synthea FHIR dataset 


namsaraeva:
### Description of feature

Add a jupyter notebook with simple preprocessing of the synthea dataset

Issue #532, opened 2023-06-01
Add more statistical tests for comparing features


Zethson:
### Description of feature

Discussed with Vladimir and it's not wrong to just apply two different statistical tests depending on the type of data. The p-values are still calculated per feature and stored in the array.

PR #1087, opened 2026-07-08
Handle falsy explicit impute replacements

driavysinus:
## Problem

`explicit_impute()` accepts mapping replacement values typed as `str | int | float`, but falsy values such as `0` are skipped because the current implementation checks replacement values by truthiness.

For example, `replacement={'intcol': 0}` should be a valid per-column replacement, but `0` is treated as if no replacement was configured.

## Fix

Use mapping key existence checks in `_extract_impute_value()` and `is not None` in the caller, so valid falsy replacements are applied. Added a regression test for `replacement={'intcol': 0}`.

## Validation

```text
.venv\Scripts\pytest.exe tests\preprocessing\test_imputation.py::test_explicit_impute_subset_accepts_falsy_replacement_value -q
# 2 passed

.venv\Scripts\pytest.exe tests\preprocessing\test_imputation.py::test_explicit_impute_subset -q
# 2 passed

python -m py_compile ehrapy\preprocessing\_imputation.py tests\preprocessing\test_imputation.py
```

## Risk

Low. This changes behavior only for mapping replacement values that are valid by type but were previously skipped because they are falsy.


driavysinus:
The `GPU tests pass` check did not execute the GPU suite; it failed at the fork-permission gate in `.github/workflows/test_gpu.yml`. The maintainer-applied `skip-gpu-ci` label reflects that.

tests/preprocessing/test_imputation.py
```diff
@@ -580,6 +580,19 @@ def test_explicit_impute_subset(impute_edata, array_type):
     assert np.sum([edata_imputed.X == "REPLACED"]) == 1
 
 
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_NONNUMERIC)
+def test_explicit_impute_subset_accepts_falsy_replacement_value(impute_edata, array_type):
```
sueoglu: It would be nice to test other falsy values than only 0, what do you think?

sueoglu (commented):
Thanks!

eroell (approved):
I see the point. Thank you!

Additional regression test is great, with all previous tests still passing: as this is of course correcting the function towards intended behavior, although for quite a special case.

Please add a release note - this is also to give your work visible credit!

eroell:
Seems here on the CI some tests fail, which are most likely unrelated and are due to new releases.

For `miceforest`, the problem is with the upstream package itself:

https://github.com/AnotherSamWilson/miceforest/issues/104

driavysinus:
Thanks again for the review and approval. I rechecked this against the latest `main`. #1097 removed `ehrapy/preprocessing/_imputation.py` and `explicit_impute()` entirely, so this fix no longer has a valid target; rebasing it would reintroduce an API that upstream intentionally removed. I’m closing this PR as superseded by that upstream removal.

Zethson:
Wait, are you sure? It should have only removed mice_forest_impute

driavysinus:
You're right — I misread #1097. It removed only `mice_forest_impute`; `explicit_impute()` remains. I've merged the current `main` and reopened this PR.

The current CPU failure is unrelated to this change: PR run #534 and the scheduled `main` run #532 fail on the same two matplotlib image-size comparisons (`test_catplot_vanilla` and `test_stratified_table_one_plot`) on Python 3.12 and 3.14.

eroell (approved):

eroell:
Thank you very much!

PR #1097, opened 2026-07-25
Remove `mice_forest_impute` and drop the `miceforest` dependency

Zethson:
## Why

`miceforest` is effectively unmaintained and is now broken against current `lightgbm`.
Removing it is easiest.


eroell:
Thank you!

Issue #1093, opened 2026-07-14
Documentation update: replace .layers["tem_data"] with `.X`


eroell:
- [ ] docstrings
- [ ] tutorials

PR #1098, opened 2026-08-04
Docstring examples: use 3D .X instead of .layers["tem_data"]

sueoglu:
Addresses the docstrings checkbox of #1093.

Docstring examples no longer pass layer="tem_data": the time series is kept in the default 3D .X

**Changes**
- ehrapy/preprocessing/_normalization.py: 36 occurrences across all 8 documented functions
- ehrapy/preprocessing/_imputation.py: locf_impute


ehrapy/plot/_ncp.py
```diff
@@ -218,7 +220,9 @@ def ncp_cluster_trajectories(
 
     Examples:
         >>> import ehrdata as ed, ehrapy as ep
-        >>> edata = ed.dt.ehrdata_blobs(n_variables=8, n_centers=3, n_observations=30, base_timepoints=12)
+        >>> edata = ed.dt.ehrdata_blobs(
+        ...     n_variables=8, n_centers=3, n_observations=30, base_timepoints=12, layer="tem_data"
```
eroell, line 224: avoid "tem_data" entirely? should work now?

ehrapy/plot/_sankey.py
```diff
@@ -164,7 +164,9 @@ def sankey_diagram_time(
     Examples:
         >>> import ehrapy as ep
         >>> import ehrdata as ed
-        >>> edata = ed.dt.ehrdata_blobs(base_timepoints=5, n_variables=1, n_observations=5, random_state=59)
+        >>> edata = ed.dt.ehrdata_blobs(
```
eroell, line 167: here also, could `"tem_data"` be entirely removed? This should be using native `.X` now

PR #1105, opened 2026-08-17
Fix mcar_test crash on 3D data with a single timepoint

### LINKED ISSUE #1104: mcar_test crashes on 3D data with a single timepoint
`ep.pp.mcar_test` is decorated with `@function_2D_only()`, which explicitly permits 3D arrays as long as the trailing time axis has size 1.. But the function body never squeezes that axis before use, so it crashes on exactly the input the decorator says is allowed.

This is easy to hit now that ehrdata.dt.ehrdata_blobs() stores its tensor in .X by default ([ehrdata#285](https://github.com/theislab/ehrdata/pull/285)): a single-timepoint dataset (base_timepoints=1) yields a .X of shape (n_obs, n_vars, 1) instead of the old 2D snapshot.

**Code:**
```python
import ehrdata as ed
import ehrapy as ep

edata = ed.dt.ehrdata_blobs(
    n_observations=100, n_variables=5, missing_values=0.1, random_state=0, n_centers=1, base_timepoints=1
)
ep.pp.mcar_test(edata)
```

**Traceback:**
```
Traceback (most recent call last):
  File "test_mcar.py", line 7, in <module>
    print(ep.pp.mcar_test(edata))
          ~~~~~~~~~~~~~~~^^^^^^^
  File "ehrapy/_compat.py", line 117, in wrapper
    return func(*args, **kwargs)
  File "ehrapy/preprocessing/_quality_control.py", line 699, in mcar_test
    return _little_mcar_test(mtx)
  File "functools.py", line 934, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "ehrapy/preprocessing/_quality_control.py", line 717, in _
    n, p = X.shape
    ^^^^
ValueError: too many values to unpack (expected 2)
```

sueoglu:
Fixes #1104
- squeeze the trailing single time axis after resolving `mtx`, mirroring what `function_2D_only()` already tolerates
- updated the docstring's expected output, which changes as a result

### DIFF
--- a/ehrapy/preprocessing/_quality_control.py
+++ b/ehrapy/preprocessing/_quality_control.py
@@ -683,9 +683,11 @@ def mcar_test(
         ...     n_observations=100, n_variables=5, missing_values=0.1, random_state=0, n_centers=1, base_timepoints=1
         ... )
         >>> ep.pp.mcar_test(edata)
-        0.327...
+        0.1416...
     """
     mtx = edata.X if layer is None else edata.layers[layer]
+    if mtx.ndim == 3:
+        mtx = mtx[:, :, 0]
 
     # float64 required: covariance estimation and linear solves need stable floating-point math
     if mtx.dtype != np.float64:
--- a/ehrapy/preprocessing/_quality_control.py
+++ b/ehrapy/preprocessing/_quality_control.py
@@ -686,6 +686,8 @@ def mcar_test(
         0.1416...
     """
     mtx = edata.X if layer is None else edata.layers[layer]
+
+    # sequeeze the array if input is inherently 2D (only 1 timepoint of shape (x,y,1)), 3D already checked by @function_2D_only()
     if mtx.ndim == 3:
         mtx = mtx[:, :, 0]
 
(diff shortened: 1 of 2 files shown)


ehrapy/preprocessing/_quality_control.py
```diff
@@ -683,9 +683,11 @@ def mcar_test(
         ...     n_observations=100, n_variables=5, missing_values=0.1, random_state=0, n_centers=1, base_timepoints=1
         ... )
         >>> ep.pp.mcar_test(edata)
-        0.327...
+        0.1416...
     """
     mtx = edata.X if layer is None else edata.layers[layer]
+    if mtx.ndim == 3:
+        mtx = mtx[:, :, 0]
```
eroell, line 692: very clean, since 2D wrapper - can you make a comment here, that means only if the 3d dim is 1, an array can enter here?

eroell (approved):

PR #1107, opened 2026-08-18
Fix test suite for ehrdata_blobs()'s old 2D-.X default

### LINKED ISSUE #1106: Test suite assumes ehrdata_blobs() old default .X being always 2D
**Report**

`ehrdata.dt.ehrdata_blobs()` changed its default behavior in theislab/ehrdata#285: previously `.X` was always populated with a 2D snapshot regardless of the `layer` argument, and `layer` only controlled where the full 3D tensor was additionally written. Now `layer` controls the tensor's only destination: omit it and the tensor lands in `.X` (making it genuinely 3D); pass a name and `.X` is left `None`.

Bumping `submodules/ehrdata` to a commit including #285 causes widespread test failures because several fixtures/tests rely on the old dual-write guarantee:

**1. `edata_blob_small` (tests/conftest.py)** : used by dozens of unrelated tests (PCA, neighbors, encoding, outliers, survival analysis, etc.) as a "give me 2D data" fixture:

```python
edata = ed.dt.ehrdata_blobs(n_variables=10, n_centers=2, n_observations=50, base_timepoints=10)
edata.layers["layer_2"] = edata.X.copy()
ep.pp.neighbors(edata)
```
No `layer=` passed, so` .X` is now genuinely (50, 10, 10). Every 2D-only consumer fails:

`ValueError: Found array with dim 3, while dim <= 2 is required by KNeighborsTransformer.`

Affects: `tests/preprocessing/test_scanpy_pp_api.py`, `tests/tools/test_scanpy_tl_api.py`, `tests/preprocessing/test_neighbors.py`, `tests/preprocessing/test_encoding.py`, `tests/preprocessing/test_outliers.py`, `tests/tools/test_sa.py`, `tests/preprocessing/test_highly_variable_features.py`, `tests/plot/test_scanpy_pl_api.py`, `tests/plot/test_missingno_pl_api.py`, `tests/preprocessing/test_summarize_measurements.py`, `tests/tools/feature_ranking/*`, and more.

**2. `edata_blobs_timeseries_small` (tests/conftest.py)** : passes `layer=DEFAULT_TEM_LAYER_NAME` explicitly:

```python
edata = ed.dt.ehrdata_blobs(..., layer=DEFAULT_TEM_LAYER_NAME)
edata.layers["layer_2"] = edata.X.copy()
```

`.X` is now `None` under this call pattern, so `.copy()` fails:

```
AttributeError: 'NoneType' object has no attribute 'copy'
```

Affects: `tests/preprocessing/test_normalization.py`, `tests/preprocessing/test_correlation.py`, `tests/tools/test_embeddings.py::test_famd_ehrdata_integration`, `tests/plot/test_correlation_plot.py`.

**3. `tests/plot/test_sankey.py`** : calls `ehrdata_blobs()` without `layer=`, then reads `edata.layers["tem_data"]` directly:

```
KeyError: 'tem_data'
```

Affects: `test_error_cases`, `test_sankey_time_bokeh_plot`, `test_sankey_time_plot`.

**4. `test_mcar_test_*_matches_pyampute_reference` (tests/preprocessing/test_quality_control.py)**: hardcoded expected p-values were computed against the old (buggy) 2D-snapshot logic, which didn't always propagate per-timepoint missingness correctly. With the real 3D tensor now used consistently, the actual values differ:

```
AssertionError: Mismatch for mcar_small: observed=0.7707353468664506, expected=0.8127459507579607
AssertionError: Mismatch for mcar_medium_high_missing: observed=0.1963445209509899, expected=0.001045380246672778
IndexError: boolean index did not match indexed array along axis 1; size of axis is 8 but size of corresponding boolean axis is 1
```

**Versions**

ehrdata >= commit 16c86d4 / theislab/ehrdata#285; anndata >= 0.13 (required for 3D .X)

sueoglu:
fixes #1106 
**Changes**

- `submodules/ehrdata`: bumped to include theislab/ehrdata#285.
- `tests/conftest.py`:
    - edata_blob_small: now explicitly requests the tensor via `layer=DEFAULT_TEM_LAYER_NAME` and derives a 2D .X snapshot from it, so the 15 tests that usethis fixture as 2D data (PCA, neighbors, encoding, outliers, survival analysis, etc.) keep working, while tests/plot/test_timeseries.py's use of the same fixture's 3D `layers[DEFAULT_TEM_LAYER_NAME] `still works too.
     - edata_blobs_timeseries_small: layer_2 now copies from `layers[DEFAULT_TEM_LAYER_NAME]` instead of the now-None .X.
- `tests/plot/test_sankey.py`: pass` layer=DEFAULT_TEM_LAYER_NAME` explicitly to `ehrdata_blobs()` in the three tests that need the named layer to exist.
- `tests/tools/test_embeddings.py`: `test_famd_ehrdata_integration` derived its single-timepoint layer from` edata.X` (now None); reworked to slice it from the existing 3D layer instead.
     
- `tests/preprocessing/test_quality_control.py`:
    - _make_mar_edata: squeezes the 3D-time .X before doing 2D-style indexing
   - Regenerated the hardcoded pyampute reference values (`tests/data/preprocessing/mcar_refs`) that this helper feeds, fixing the squeeze changes the actual data generated (the old 2D-snapshot logic had a bug where missingness wasn't always propagated correctly), so the old reference numbers no longer apply.


ehrapy/preprocessing/_quality_control.py
```diff
@@ -683,9 +683,11 @@ def mcar_test(
         ...     n_observations=100, n_variables=5, missing_values=0.1, random_state=0, n_centers=1, base_timepoints=1
         ... )
         >>> ep.pp.mcar_test(edata)
-        0.327...
+        0.1416...
```
eroell, line 686: For the future readers: Please reply here quickly on the reason for the number change - you already had a lot of details on this I think, so its worth it to mention this here, too :) If I recall correctly, its essentially different data being used for the test now due to updated ehrdata blobs

tests/plot/test_sankey.py
```diff
@@ -26,7 +26,9 @@ def test_sankey_plot(diabetes_130_fairlearn_sample_100, check_same_image, hv_bac
 
 def test_sankey_time_plot(check_same_image, hv_backend):
     hv_backend("matplotlib")
-    edata = ed.dt.ehrdata_blobs(base_timepoints=5, n_variables=1, n_observations=5, random_state=59)
+    edata = ed.dt.ehrdata_blobs(
+        base_timepoints=5, n_variables=1, n_observations=5, random_state=59, layer=DEFAULT_TEM_LAYER_NAME
```
eroell, line 30: Fine for now - ehrdata will soon require anndata >= 0.13, then the layers can be dropped.

tests/preprocessing/test_quality_control.py
```diff
@@ -496,7 +497,21 @@ def test_mcar_test_ttest_detects_mar(mar_edata):
     assert p_col0_given_miss9 < 0.05
 
 
-@pytest.fixture(params=list(_SCENARIOS_LITTLE))
+# _little_mcar_test estimates pairwise-deletion covariance using each column's global mean rather than
```
eroell: linebreaks after sentences, not arbitrary in the middle of the sentence for our readability pls :)

tests/preprocessing/test_quality_control.py
```diff
@@ -496,7 +497,21 @@ def test_mcar_test_ttest_detects_mar(mar_edata):
...
+# the mean of the jointly-observed subset, which diverges from pyampute's pandas.cov()-based reference at
+# high missingness / many variables. Pre-existing algorithm issue, unrelated to the ehrdata_blobs 2D/3D .X
```
eroell: Hmm, so does that mean our little test is different than pyampute? That would not be good. Could you check this with claude, and ask it to check both and give examples of where they differ?
If they do, please open a new issue (and ultimately a PR) :)

Fixing it is beyond the scope of this PR. but we don't need this comment here warning about an implementation error in our tests, it can be deleted

ehrapy/preprocessing/_quality_control.py
sueoglu: Fixing the crash also changed .X actually contains, before the fix it was containing a 2D snapshot of the data which sued to silently subsitute non-missing placeholder values (no NaNs at all) for every row where feature_0 happened to be missing. (This way "is feature_0 missing" was treated as as a proxy for "should this row get its real data at all," when feature_0 missing was actually not telling anything about whether feature_1 feautre_2 etx are missing ) Since .X now holds the real 3D tensor directly rather than the old buggy 2D snapshot, the test runs against a 'more correct' dataset than before.

tests/preprocessing/test_quality_control.py
```diff
@@ -501,7 +502,17 @@ def test_mcar_test_ttest_detects_mar(mar_edata):
     assert p_col0_given_miss9 < 0.05
 
 
-@pytest.fixture(params=list(_SCENARIOS_LITTLE))
+_LITTLE_SCENARIO_XFAILS = {"mcar_medium_high_missing": "diverges from pyampute reference at high missingness"}
```
eroell, line 505: btw was noticed here, and will be fixed separately https://github.com/theislab/ehrapy/issues/1110

PR #1109, opened 2026-08-31
ep.pp.knn_impute clean behaviour with different array types

### LINKED ISSUE #849: Compatibility of ep.pp.knn_impute with different array types

**3D**
If 3D allowed:
- [x] function handles 3D data

If 3D not allowed:
- [ ] only_2D decorator

**array-type**

sueoglu:
closes #849 

**Description of changes**

Adapts `knn_impute `to follow the same array-type dispatch pattern already used by `simple_impute` and `miss_forest_impute`. The core matrix processing logic (2D/3D reshaping, anchor-column selection, and the call to the imputer's `fit_transform`) is now extracted into a singledispatch function`, _knn_impute_function`, registered for `np.ndarray`. Any other array type (dask arrays, scipy sparse arrays) now raises a clear NotImplementedError. 

**Technical details**
Neither of the two backends knn_impute supports (sklearn.impute.KNNImputer, fknni.FastKNNImputer) can operate on sparse or dask input. Thats why, it is now restricted to `np.ndarray`

Added `test_knn_impute_array_types `in `tests/preprocessing/test_imputation.py`, parametrized over `ARRAY_TYPES_NUMERIC `and both knn_impute backends (scikit-learn, faiss), asserting NotImplementedError for dask/sparse input and correct imputation for np.ndarray.


ehrapy/preprocessing/_imputation.py
```diff
@@ -373,6 +373,48 @@ def knn_impute(
...
+    raise _raise_array_type_not_implemented(_knn_impute_function, type(arr))
+
+
+@_knn_impute_function.register(np.ndarray)
+def _(arr: np.ndarray, var_indices: list[int], numerical_indices: list[int], imputer) -> tuple[list[int], np.ndarray]:
+    var_indices_original = var_indices
+    is_3d = arr.ndim == 3
+    input_dtype = arr.dtype if np.issubdtype(arr.dtype, np.floating) else np.float64
+
+    # if input data is 3D, flatten along axis 0 before passing it to the imputer: each timepoint becomes a row
```
eroell: This should be usable, no?

https://github.com/theislab/ehrapy/blob/f045b1b5f8dc0dbfb789fe6f0ca337d9acd12876/ehrapy/_compat.py#L55

tests/preprocessing/test_imputation.py
```diff
@@ -352,6 +352,20 @@ def test_knn_impute_numerical_data(impute_num_edata):
     _base_check_imputation(impute_num_edata, edata_imputed)
 
 
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_NUMERIC)
+@pytest.mark.parametrize("backend", ["scikit-learn", "faiss"])
+def test_knn_impute_array_types(impute_num_edata, array_type, backend):
```
eroell, line 357: I think this does not test the 3D case? Please correct me if I'm wrong, otherwise please add a 3D case - as per usual, ideally with a fixture we already have or ehrdata blobs.

Next to the testing, please also give the imputation also a quick human trial check. Does it work on e.g. a subset of physionet2012, are the values imputed reasonable, are indeed all missing values imputed that should be so. Could be a very quick few lines in a new PR comment :)

eroell (commented):
cool, we're getting there

ehrapy/preprocessing/_imputation.py
sueoglu: We had a different case here where we narrowed down the imputed slice to only requested variables and computed the anchors from that column set (we did that because the full column set was not guaranteed to contain only numeric variables, it was as far as I can remember the simple way around this problem). so the reshaping to 2D and then 3D on the ehrdata with this smaller subset of variables didnt work directly with apply_over_time_axis due to the shape mismatch. but I think we can slice under _apply_over_time_axis to get the numeric columns and then compute the anchors inside the single dispatched function and it should be fine

sueoglu:
**Working example on subset of Physionet2012 with a missingness of 10%:**

```python
import ehrapy as ep
import ehrdata as ed
import numpy as np

edata_full = ed.dt.physionet2012(layer="tem_data")
ed.infer_feature_types(edata_full, layer="tem_data")
edata = edata_full[:200]

full_layer = edata.layers["tem_data"].copy()

non_nan_mask = ~np.isnan(full_layer)
mask_indices = np.where(non_nan_mask)
n_to_mask = int(0.1 * len(mask_indices[0]))
chosen = rng.choice(len(mask_indices[0]), size=n_to_mask, replace=False)

masked_layer = full_layer.copy()
masked_layer[mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]] = np.nan
edata.layers["tem_data"] = masked_layer

edata_knn = edata.copy()

ep.pp.knn_impute(edata_knn, layer="tem_data")

true_vals = full_layer[mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]]
knn_vals = edata_knn.layers["tem_data"][mask_indices[0][chosen], mask_indices[1][chosen], mask_indices[2][chosen]]

print("True mean:", np.nanmean(true_vals))
print("KNN mean:", np.nanmean(knn_vals), "diff:", abs(np.nanmean(knn_vals) - np.nanmean(true_vals)))

print("True std:", np.nanstd(true_vals))
print("KNN std:", np.nanstd(knn_vals), "diff:", abs(np.nanstd(knn_vals) - np.nanstd(true_vals)))
```

Output: 
```
True mean: 70.79201158141458
KNN mean: 71.27235533784102 diff: 0.4803437564264357
True std: 58.31532031908203
KNN std: 41.676400044254244 diff: 16.63892027482779
```

eroell:
CellRank notebook failure cause is upstream https://github.com/scverse/cellrank/issues/1377

eroell (approved):

Issue #1110, opened 2026-08-31
`mcar_test(method="little")` diverges from pyampute's reference


sueoglu:
`_little_mcar_test` estimates the pairwise-deletion covariance matrix incorrectly: it centers each covariance entry using each column's **global** mean (computed from *all* of that column's non-missing values across the whole dataset), instead of the **pairwise** mean (recomputed per variable pair, using only the rows where both variables in that pair are observed). Standard pairwise-deletion covariance (and what `pandas.DataFrame.cov()` (used internally by `pyampute`) computes) uses the pairwise mean.

The two formulas agree closely when missingness is low, because the "all valid" subset and the "pairwise valid" subset mostly overlap. They diverge as missingness rises and each pairwise subset excludes a much larger, more different chunk of rows than the global-mean calculation accounts for.

### Repro

```python
import ehrdata as ed
import ehrapy as ep

edata = ed.dt.ehrdata_blobs(
    n_observations=900, n_variables=50, missing_values=0.5, random_state=7,
    n_centers=1, cluster_std=1.0, base_timepoints=1,
)

ep.pp.mcar_test(edata, method="little")
# 0.19634452095098048

# Reference, computed via pyampute on the same data:
import pandas as pd
from pyampute.exploration.mcar_statistical_tests import MCARTest
df = pd.DataFrame(edata.X[:, :, 0], columns=list(edata.var_names))
MCARTest(method="little")(df)
# 0.20943902578030515
```

### Concrete covariance differences

Each variable pair here has only ~200 jointly-observed rows out of 900 (roughly 50% × 50% overlap). Comparing ehrapy's current (global-mean) covariance against pandas' `.cov()` (pairwise-mean) on identical data:
```

pair                          pandas .cov()  ehrapy (current)    abs diff  rel diff
(feature_13, feature_34):           -0.0264           -0.0042      0.0222    84.0%
(feature_23, feature_40):           -0.1262           -0.1406      0.0144    11.4%
(feature_0, feature_44):            -0.0015           -0.0135      0.0120   813.3%
(feature_27, feature_39):            0.0760            0.0769      0.0009     1.2%
```

[...]

PR #1115, opened 2026-09-01
fix(qc): centre Little's covariance on pairwise means

asaraog:
Addresses the covariance half of #1110. **It does not close the issue** — see the caveat at the bottom.

`_little_mcar_test` divided pairwise cross-products by pairwise counts but centred every entry on the column's *global* mean. Those disagree once missingness is high enough that a pair's jointly observed rows differ from the column's own observed rows.

Max absolute deviation from `pandas.DataFrame.cov()` (what pyampute uses), 400x8 normal data:

| missing | before | after |
|---|---|---|
| 0% | 1.1e-16 | 2.2e-16 |
| 10% | 3.6e-04 | 2.2e-15 |
| 30% | 2.6e-03 | 1.2e-15 |
| 50% | 8.3e-03 | 1.6e-15 |
| 70% | 3.3e-02 | 1.3e-15 |

Extracted the estimator as `_pairwise_deletion_cov` so it can be tested directly rather than only through a p-value, and added a parametrized test against `pandas.DataFrame.cov()`.

### What this does not fix

`test_mcar_test_little_matches_pyampute_reference` still fails for `mcar_small` and `mcar_medium_high_missing`, essentially unchanged:

| scenario | before | after | reference |
|---|---|---|---|
| `mcar_small` | 0.77074 | 0.77065 | 0.81275 |
| `mcar_medium_high_missing` | 0.19634 | 0.20944 | 0.00105 |

So the covariance was not the dominant source of the divergence. On `mcar_medium_high_missing` (900x50 at 50% missing) the statistic comes out at d²=22765 on df=22585 — at that missingness almost every row is its own missingness pattern, so `df = Σ k_j - p` is dominated by single-row patterns. Two things look worth deciding before that test can pass:

1. **Estimator.** Little's test is defined on the ML (EM) estimates of μ and Σ; this implementation uses available-case `np.nanmean` and a pairwise covariance.
2. **Degrees of freedom.** Whether single-observation patterns should contribute to `df` at all.

Both are calls for a maintainer, so I've kept this PR to the piece that is unambiguously wrong on its own terms. Happy to follow up on either if you have a preference.

No regressions: the failing set in `tests/preprocessing/test_quality_control.py` is identical before and after (the pre-existing dask and 3D-fixture failures in my environment), plus the 5 new passing cases.

### DIFF
--- a/ehrapy/preprocessing/_quality_control.py
+++ b/ehrapy/preprocessing/_quality_control.py
@@ -706,6 +706,35 @@ def mcar_test(
     raise ValueError(f"Unknown method {method!r}. Choose from 'little' or 'ttest'.")
 
 
+def _pairwise_deletion_cov(X: np.ndarray) -> np.ndarray:
+    """Covariance matrix of `X` by pairwise deletion.
+
+    Each entry is centred on the means of its own variable pair, computed over
+    the rows where both variables are observed. Centring on the global column
+    means instead biases every entry once missingness is high enough that a
+    pair's observed rows differ appreciably from the column's own observed
+    rows. This matches :meth:`pandas.DataFrame.cov`, the reference Little's
+    test is defined against.
+
+    Args:
+        X: 2D array of observations, missing values encoded as NaN.
+
+    Returns:
+        The pairwise-deletion covariance matrix.
+    """
+    valid = ~np.isnan(X)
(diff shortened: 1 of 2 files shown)


Zethson:
AI generated spam

PR #1121, opened 2026-09-16
Fix ep.pl.timeseries and ep.pl.variable_correlations/variable_dependencies don't support the default .X storage for 3D data

### LINKED ISSUE #1119: `ep.pl.timeseries` and `ep.pl.variable_correlations` don't support the default `.X` storage for 3D data
Since the 3D-data default changed to `.X`, most `ep.pp`/`ep.tl` functions already accept `layer: str | None = None` and fall back to `.X`. Two plotting functions were missed and currently require an explicitly named layer, which blocks migrating tutorials/docstrings off `layers["tem_data"]`.Found out about this while working on removing the layers["tem_data"] references in `longitundal_with_ehrapy.ipynb` notebook. 

**1. `ep.pl.variable_correlations`: signature is stricter than necessary**

`layer` is required. But the function's body does nothing with `layer` except forward it to `ep.pp.variable_correlations()` and `ep.pp.variable_correlations` already accepts `layer: str | None = None` . The fix would be to change the signature of `ep.pl.variable_correlations` to `layer: str | None = None`

**2. `ep.pl.timeseries`: no `.X` fallback exists at all**

There is no code path that reads `edata.X`. Fix would be `layer: str | None = None`; when `None`, read from `edata.X` instead of raising an error

**Checklist**

- [ ] `ep.pl.variable_correlations`: `layer: str` → `layer: str | None = None`
- [ ] `ep.pl.timeseries`: add `.X` fallback when `layer is None`; update default from `"tem_data"` to `None`
- [ ] Update both functions' docstrings/examples accordingly
- [ ] Add/adjust tests covering the `layer=None` path for each
- [ ] Unblocks: removing the remaining `layers["tem_data"]` references in `docs/tutorials/notebooks/

sueoglu:
fixes #1119 
3D-data default storage recently moved to .X, and most preprocessing/analysis functions already accept `layer: str | None = None` with an .X fallback but` ep.pl.variable_correlations`/`variable_dependencies`/ didn't.

**Changes**
- `ep.pl.timeseries`: layer default changed from `tem_data` to `None`; reads from `edata.X` when layer is `None `instead of raising `KeyError` 
- `ep.pl.variable_correlations` / `ep.pl.variable_dependencies`: layer is now` str | None = None`, matching `ep.pp.variable_correlations`.
- Updated docstring examples to drop the now unnecessary explicit `layer="tem_data"`.
- Added tests covering the` layer=None` -> `.X` path for all three functions, and a new error-case test for timeseries when`.X` isn't 3D.


tests/plot/test_timeseries.py
```diff
@@ -20,6 +20,14 @@ def test_timeseries(edata_blob_small):
     assert isinstance(plot, hv.Layout)
 
 
+def test_timeseries_x():
```
eroell, line 23: can be parametrized into test_timeseries with pytest parametrization we also use in other places I believe

tests/plot/test_correlation_plot.py
```diff
@@ -17,3 +18,19 @@ def test_correlation_chord(edata_blobs_timeseries_small):
...
+
+def test_correlation_heatmap_x():
+    edata = ed.dt.ehrdata_blobs(n_variables=5, n_centers=2, n_observations=20, base_timepoints=3)
+
+    heatmap = ep.pl.variable_correlations(edata)
+    assert heatmap is not None
+    assert isinstance(heatmap, hv.Overlay)
+
+
+def test_correlation_chord_x():
```
eroell, line 31: can be parametrized into test_timeseries with pytest parametrization we also use in other places I believe

ehrapy/plot/_timeseries.py
```diff
@@ -68,11 +68,15 @@ def timeseries(
     opts_dict["shared_axes"] = True
     opts_dict["legend_position"] = "right"
 
-    if layer not in edata.layers:
-        raise KeyError(f"Layer {layer!r} not found in edata.layers. Available layers: {list(edata.layers)}")
-    mtx = np.asarray(edata.layers[layer])
+    if layer is None:
+        mtx = np.asarray(edata.X)
```
eroell, line 72: now that we're here: single-dispatch this function also with clear notimplementederrors when needed - no secret densifications of e.g. a sparse array

ehrapy/plot/_timeseries.py
```diff
@@ -68,11 +73,17 @@ def timeseries(
...
+        raw_mtx = edata.X
+    else:
+        if layer not in edata.layers:
+            raise KeyError(f"Layer {layer!r} not found in edata.layers. Available layers: {list(edata.layers)}")
+        raw_mtx = edata.layers[layer]
+    source = ".X" if layer is None else f"Layer {layer!r}"
+    if raw_mtx is None or np.ndim(raw_mtx) != 3:
+        shape = () if raw_mtx is None else np.shape(raw_mtx)
+        raise ValueError(f"{source} must be 3D (n_obs, n_vars, n_time), got shape {shape}.")
+    mtx = _get_dense_mtx(raw_mtx)
```
eroell, line 86: Lets please really try to avoid densifying sparse arrays or instantiating dask arrays under the hood.
Also, this function here is called `_get_dense_mtx`, but is actually more the single-dispatch gate for this function. Let's use a naming more similar as we do in other functions, the name is not at all describing what this function is doing.

If the function does not work for Dask, that's currently ok, but we should raise a NotImplementedError.

eroell (changes requested):

